PR #1, opened 2020-10-02
Initial package infrastructure and import from jwst

eslavich:
This PR includes some basic package infrastructure and imports datamodels core modules from jwst at commit [c353186397091d1821ca8723da49b182346d5626](https://github.com/spacetelescope/jwst/tree/c353186397091d1821ca8723da49b182346d5626).  I made the following changes to the original code:

- Remove unused functions and test fixtures
- Update package and test imports to use stdatamodels instead of jwst.datamodels
- Comment out code where jwst datamodel implementations are used and raise NotImplementedError instead
- Make schema_url optional and provide a default schema object (not a URL, but an actual schema tree) that includes the attributes that DataModel needs to access
- Mark skip any tests that need jwst-specific code or schemas


src/stdatamodels/model_base.py
```diff
@@ -117,11 +137,14 @@ def __init__(self, init=None, schema=None, memmap=False,
...
-                                             resolver=asdf_file.resolver,
-                                             resolve_references=True)
+            if self.schema_url is None:
+                schema = _DEFAULT_SCHEMA
+            else:
+                # Create an AsdfFile so we can use its resolver for loading schemas
+                asdf_file = AsdfFile()
+                schema = asdf_schema.load_schema(self.schema_url,
+                                                resolver=asdf_file.resolver,
+                                                resolve_references=True)
```
jdavies-st, line 147: This can probably be switched to the new way
```suggestion
                schema = asdf_schema.load_schema(self.schema_url, resolve_references=True)
```
as now the default resolver pulls in all the installed extensions, right?

src/stdatamodels/model_base.py
```diff
@@ -263,6 +286,17 @@ def __init__(self, init=None, schema=None, memmap=False,
...
+    @property
+    def schema_url(self):
+        """
+        The schema URI to validate the model against.  If
+        None, no validation will occur.
+        Returns
+        -------
+        str or None
+        """
+        return None
```
jdavies-st, line 298: Will this property have to be overwritten by any subclasses that have a `schema_url`?

jdavies-st (commented):
Looks good!

I notice that there's  some association JSON files in the test/data dir.  Also a FITS file.  Are these needed for the tests?

Couple questions below.

src/stdatamodels/model_base.py
eslavich: Yes, though this is a valid way to overwrite it:

```
class FooModel(DataModel):
    schema_url = "http://stsci.edu/schemas/foo"
```

so this should be compatible with the existing jwst model classes.  The only reason why I went with a property over a class attribute was so I could add a docstring.

src/stdatamodels/model_base.py
eslavich: Yes, that's the better way for sure

eslavich:
> I notice that there's some association JSON files in the test/data dir. Also a FITS file. Are these needed for the tests?

The tests do use those, though I think most of the relevant tests are currently marked skipped.

tests/data/collimator_fake.asdf
```diff
@@ -0,0 +1,40 @@
+#ASDF 1.0.0
```
nden, line 1: This file should stay in jwst/datamodels/tests. The test that uses it is commented out so it's safe to remove it.

nden (commented):
Minor comments. The tests can be further cleaned by leaving some in jwst. It can be done in subsequent PRs.

setup.cfg
```diff
@@ -0,0 +1,56 @@
+[metadata]
+name = stdatamodels
+description = TODO
+long_description = TODO
+author = TODO
+license = BSD-3-Clause
+url = https://github.com/stdatamodels/jwst
+project_urls =
+    Bug Tracker = https://github.com/spacetelescope/stdatamodels/issues
+    Source Code = https://github.com/stdatamodels/jwst
```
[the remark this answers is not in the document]
eslavich: Oops!  I sure got those backwards.

src/stdatamodels/model_base.py
eslavich: I'll switch this back to an attribute

perrygreenfield (approved):
Didn't see any glaring issues. But maybe that is due to visual impairments...

nden (approved):

PR #2, opened 2020-10-05
Remove stdatamodels.open

eslavich:
The `open` function relies heavily on jwst-specific DataModel subclasses, so for now at least we're just going to continue using `jwst.datamodels.open`.

This is branched off of #1 so it will be easier to review after that one is merged.


jdavies-st (approved):
Looks good!

nden (approved):

PR #4, opened 2020-10-15
Add test coverage of validation code

eslavich:
This adds new tests for the validation features of DataModel, and makes the following changes:

- Make attribute updates respond to pass_invalid_values
- Remove unreachable code in validate.py relating to type validation
- Remove unnecessary call to custom_tree_to_tagged_tree

The testing exposed some bugs/missing features that I'll open issues for:

- Validation on read does not occur for ASDF files
- Validation on write does not occur for ASDF files
- The JSON Schema `required` property is ignored

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,6 +21,9 @@ setup_requires =
     setuptools_scm
 install_requires =
     jsonschema>=3.0.2
+    # The tests require code that is in asdf master
+    # but not yet released.  The package itself
+    # is compatible with asdf 2.7.
     asdf>=2.7.1
     psutil>=5.7.2
     numpy>=1.16
--- a/src/stdatamodels/validate.py
+++ b/src/stdatamodels/validate.py
@@ -4,9 +4,7 @@ Functions that support validation of model changes
 
 import warnings
 import jsonschema
-from asdf import AsdfFile
 from asdf import schema as asdf_schema
-from asdf.util import HashableDict
 from asdf import yamlutil
 
 
@@ -14,45 +12,28 @@ class ValidationWarning(Warning):
     pass
 
 
-def value_change(path, value, schema, pass_invalid_values,
-                 strict_validation):
+def value_change(path, value, schema, ctx):
     """
     Validate a change in value against a schema.
     Trap error and return a flag.
     """
     try:
-        _check_value(value, schema)
+        _check_value(value, schema, ctx)
         update = True
 
     except jsonschema.ValidationError as error:
         update = False
         errmsg = _error_message(path, error)
-        if pass_invalid_values:
+        if ctx._pass_invalid_values:
             update = True
-        if strict_validation:
+        if ctx._strict_validation:
             raise jsonschema.ValidationError(errmsg)
         else:
             warnings.warn(errmsg, ValidationWarning)
     return update
 
 
-def _check_type(validator, types, instance, schema):
-    """
-    Callback to check data type. Skips over null values.
-    """
-    if instance is None:
-        errors = []
-    else:
-        errors = asdf_schema.validate_type(validator, types,
-                                           instance, schema)
-    return errors
-
-
-validator_callbacks = HashableDict(asdf_schema.YAML_VALIDATORS)
-validator_callbacks.update({'type': _check_type})
-
-
-def _check_value(value, schema):
+def _check_value(value, schema, ctx):
     """
     Perform the actual validation.
     """
@@ -62,21 +43,8 @@ def _check_value(value, schema):
             raise jsonschema.ValidationError("%s is a required value"
                                               % name)
     else:
-        validator_context = AsdfFile()
-        validator_resolver = validator_context.resolver
-
-        temp_schema = {
-            '$schema':
-            'http://stsci.edu/schemas/asdf-schema/0.1.0/asdf-schema'}
-        temp_schema.update(schema)
-        validator = asdf_schema.get_validator(temp_schema,
-                                              validator_context,
-                                              validator_callbacks,
-                                              validator_resolver)
-
-        value = yamlutil.custom_tree_to_tagged_tree(value, validator_context)
-        validator.validate(value, _schema=temp_schema)
-        validator_context.close()
+        value = yamlutil.custom_tree_to_tagged_tree(value, ctx._asdf)
+        asdf_schema.validate(value, schema=schema)
 
 
 def _error_message(path, error):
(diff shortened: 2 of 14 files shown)


tests/test_validation.py
```diff
@@ -0,0 +1,237 @@
...
+        model.meta.required_keyword = "foo"
+    assert len(warnings) == 0
+
+    with pytest.warns(ValidationWarning):
+        del model.meta.required_keyword
+    assert model.meta.required_keyword is None
+
+
+def test_pass_invalid_values_attribute_assignment(monkeypatch):
+    def assert_passed(model, passed):
```
jdavies-st, line 132: This is essentially an assertion helper, so it might be good to set `__tracebackhide__ = True` in the function.

tests/test_validation.py
```diff
@@ -0,0 +1,237 @@
...
+
+    assert_strict(ValidationModel(), False)
+    assert_strict(ValidationModel(strict_validation=True), True)
+    assert_strict(ValidationModel(strict_validation=False), False)
+
+    monkeypatch.setenv("STRICT_VALIDATION", "true")
+    assert_strict(ValidationModel(), True)
+
+    monkeypatch.setenv("STRICT_VALIDATION", "false")
+    assert_strict(ValidationModel(), False)
```
jdavies-st, line 183: It might be clearer to structure this as a parametrized test like:

```
@pytest.mark.parametrize("strict_validation, patch_env, result",
    [
        (None, None, ValidationWarning),
        (True, None, ValidationError),
        (False, None, ValidationWarning),
        (None, "true", ValidationError),
        (None, "false", ValidationWarning),

    ]
```

And then the test becomes very simple with no assertion helper and a single logic check `if patch_env is not None:` to do the patching.

tests/test_validation.py
```diff
@@ -0,0 +1,237 @@
...
+
+    with pytest.warns(ValidationWarning):
+        model.validate()
+
+
+def test_validate_required_fields():
+    model = FitsRequiredModel()
+
+    with pytest.warns(ValidationWarning):
+        model.validate_required_fields()
```
jdavies-st, line 205: Is this an existing function in `jwst.datamodels`?  Seems odd to have a separate validator for `fits_required`.

tests/test_validation.py
```diff
@@ -0,0 +1,237 @@
...
+
+    with pytest.warns(ValidationWarning):
+        model.meta.string_attribute = 42
+    assert model.meta.string_attribute == 42
+
+    with pytest.warns(ValidationWarning):
+        model.validate()
+
+
+def test_validate_required_fields():
```
jdavies-st, line 201: Test name should indicate it is for `fits_required` and not the normal JSON `required`.

tests/test_validation.py
```diff
@@ -0,0 +1,237 @@
...
+    with asdf.AsdfFile() as af:
+        af["meta"] = {"string_attribute": "foo"}
+
+        with pytest.warns(None) as warnings:
+            ValidationModel(af)
+        assert len(warnings) == 0
+
+        af["meta"]["string_attribute"] = 42
+        with pytest.warns(ValidationWarning):
+            ValidationModel(af)
```
jdavies-st, line 226: This is validating on `__init__` not necessarily on read, right?  Or is this a typical way ASDF files will be opened, by passing the open `AsdfFile` instance onto `__init__`?

jdavies-st (commented):
This looks really good!  Some comments and questions below.

setup.cfg
```diff
@@ -21,6 +21,9 @@ setup_requires =
     setuptools_scm
 install_requires =
     jsonschema>=3.0.2
+    # The tests require code that is in asdf master
+    # but not yet released.  The package itself
+    # is compatible with asdf 2.7.
```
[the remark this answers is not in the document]
eslavich: I tried adding an additional `asdf @ git ...` dependency in the `[test]` extras and it had no effect.  I guess pip figures it's doing us a favor by not installing it twice.

src/stdatamodels/validate.py
```diff
@@ -4,55 +4,36 @@
...
-    if instance is None:
-        errors = []
-    else:
-        errors = asdf_schema.validate_type(validator, types,
-                                           instance, schema)
-    return errors
-
-
-validator_callbacks = HashableDict(asdf_schema.YAML_VALIDATORS)
-validator_callbacks.update({'type': _check_type})
```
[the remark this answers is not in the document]
eslavich: The code that checked `None` turned out to be unreachable.  The `None` case is caught early by [special handling in _check_value](https://github.com/spacetelescope/stdatamodels/blob/6b9a3e170b99ef625063d9e0540741e37c98b6bf/src/stdatamodels/validate.py#L40).

tests/test_validation.py
eslavich: Yes, 'tis!  I have no idea why this is necessary.  I'll add an issue to investigate removing it:

https://github.com/spacetelescope/stdatamodels/issues/12

tests/test_validation.py
eslavich: Hmm... yes the validation will be handled by `__init__` even when users call jwst/romancal.datamodels.open.  On the other hand, there's no `open` in this package.  I'll go ahead and change this test name to "test_validation_on_init".

tests/test_validation.py
eslavich: That's a good idea, I'll give it a go

tests/test_validation.py
eslavich: I reworked this one to be parametrized as well, so the assertion helper disappeared.

tests/test_validation.py
eslavich: This has been reworked as suggested.

tests/test_validation.py
eslavich: Done

tests/test_validation.py
eslavich: Done

src/stdatamodels/validate.py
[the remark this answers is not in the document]
jdavies-st: Ah, I see.  Inadvertent, as it looks like it is there to check the `fits_required` field.  Might be worth putting something like

```python
if value is None:
    if schema.get('fits_required'):
        blah blah
    else:
        # Allow `None` for any value, even where schema doesn't allow it.  These will
        # eventually be stripped when written out to FITS or ASDF
        pass
```

setup.cfg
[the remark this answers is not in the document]
jdavies-st: Right.  It must be because it can't resolve versions because when installing from Github, there's no version specifier, and it needs to be able to resolve it before it installs it.

src/stdatamodels/validate.py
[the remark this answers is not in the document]
eslavich: Oh yes definitely worth a note.  Thanks, I'll add that now.

jdavies-st (approved):
🚀

PR #8, opened 2020-10-16
Add test coverage of util module

eslavich:
This PR adds test coverage of util.py.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -139,29 +139,6 @@ def test_gentle_asarray_invalid_conversion():
         util.gentle_asarray(object(), dtype=np.float32)
 
 
-def test_get_short_doc():
-    assert util.get_short_doc({}) == ""
-    assert util.get_short_doc({"title": "Some schema title."}) == "Some schema title."
-    assert util.get_short_doc({
-        "title": "Some schema title.\nWhoops, another line."
-    }) == "Some schema title."
-    assert util.get_short_doc({
-        "title": "Some schema title.",
-        "description": "Some schema description.",
-    }) == "Some schema title."
-    assert util.get_short_doc({
-        "description": "Some schema description.",
-    }) == "Some schema description."
-    assert util.get_short_doc({
-        "description": "Some schema description.\nWhoops, another line.",
-    }) == "Some schema description."
-
-
-def test_ensure_ascii():
-    for inp in [b"ABCDEFG", "ABCDEFG"]:
-        util.ensure_ascii(inp) == "ABCDEFG"
-
-
 def test_create_history_entry():
     entry = util.create_history_entry("Once upon a time...")
     assert isinstance(entry, HistoryEntry)
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -11,15 +11,24 @@ from stdatamodels import util
 
 
 def test_gentle_asarray_array_input():
+    """
+    Test gentle_asarray with various forms of np.ndarray input,
+    which may need to be cast to a different data type.
+    """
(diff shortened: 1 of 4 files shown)


tests/test_util.py
```diff
@@ -0,0 +1,180 @@
...
+    assert result.dtype == np.float32
+    assert_array_equal(result, np.array([3.14159], dtype=np.float32))
+
+
+def test_gentle_asarray_invalid_conversion():
+    with pytest.raises(ValueError):
+        util.gentle_asarray(object(), dtype=np.float32)
+
+
+def test_get_short_doc():
```
jdavies-st, line 94: Is there a reason to have `get_short_doc` is not over in the `fits_support.py` module?  It's only used for populating the FITS keyword comment when writing out files to FITS format I believe.

tests/test_util.py
```diff
@@ -0,0 +1,180 @@
...
+        "description": "Some schema description.",
+    }) == "Some schema description."
+    assert util.get_short_doc({
+        "description": "Some schema description.\nWhoops, another line.",
+    }) == "Some schema description."
+
+
+def test_ensure_ascii():
+    for inp in [b"ABCDEFG", "ABCDEFG"]:
+        util.ensure_ascii(inp) == "ABCDEFG"
```
jdavies-st, line 114: Is there a reason to have the `ensure_ascii` function anymore now that we are python3-only?  I.e., we probably shouldn't be passing around bytes as text, and if we are, it's probably because of python2.  But if it really requires `ascii`, then we should be making sure it actually is ascii, and it seems like this function only decodes bytes to ascii.  About unicode?

jdavies-st (commented):
Looks good.  Some 1-line comments in the tests would be nice to guide the reader what is being tested.  The names of the tests merely hint.  =)

Some other questions about perhaps outdated code in `ensure_ascii`.

tests/test_util.py
eslavich: It does appear to only be used in fits_support.py, so I can quarantine it over there.

tests/test_util.py
eslavich: That is odd, isn't it.  The code in fits_support.py is using this on values extracted from the schema, which will always be strings anyway.  The FITS standard restricts the characters that may be used in a header:

image

So, maybe this method should actually be ensuring characters that fall within the allowed range?

tests/test_util.py
```diff
@@ -0,0 +1,180 @@
...
+    assert result.dtype == np.int8
+    assert_array_equal(result, inp)
+
+    dtype = np.dtype([("col1", np.int16), ("col2", np.float32)])
+    result = util.gentle_asarray(inp, dtype=dtype)
+    assert result.dtype == dtype
+    assert result["col1"].dtype == np.int16
+    assert_array_equal(result["col1"], np.array([1, 2, 3, 4], dtype=np.int16))
+    assert result["col2"].dtype == np.float32
+    assert_array_equal(result["col1"], np.array([1, 2, 3, 4], dtype=np.float32))
```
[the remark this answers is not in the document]
eslavich: Yes, broadcasting out to a recarray with multiple columns from a regular array.  This hits a corner case in the `gentle_asarray` code that wasn't getting tested otherwise.  I'll add notes to this and the rest of the tests like you suggested.

tests/test_util.py
eslavich: I'm going to punt on this by also moving it to fits_support.py, which I'm planning to add tests for in a future PR.

jdavies-st (approved):
🚀

PR #10, opened 2020-10-19
Update tests in test_models.py

eslavich:
This PR updates test_models.py to use non-jwst models defined in the tests directory.  It also moves FITS-related tests to test_fits.py (I'm planning to isolate the FITS stuff as much as possible).

~~Includes commits from #4 so best to wait to review this until that one is merged (or review only the latest commit).~~

### DIFF
--- a/tests/models.py
+++ b/tests/models.py
@@ -11,9 +11,6 @@ class BasicModel(DataModel):
     """
     schema_url = "http://example.com/schemas/basic_model"
 
-    def get_primary_array_name(self):
-        return "data"
-
 
 class ValidationModel(DataModel):
     """
--- a/tests/models.py
+++ b/tests/models.py
@@ -11,6 +11,9 @@ class BasicModel(DataModel):
     """
     schema_url = "http://example.com/schemas/basic_model"
 
+    def get_primary_array_name(self):
+        return "data"
+
 
 class ValidationModel(DataModel):
     """
@@ -33,3 +36,10 @@ class RequiredModel(DataModel):
     """
     schema_url = "http://example.com/schemas/required_model"
 
+
+class AnyOfModel(DataModel):
+    """
+    Model for which the attribute 'meta.foo' has conflicting
+    default values due to use of an anyOf combiner.
+    """
+    schema_url = "http://example.com/schemas/anyof_model"
--- a/tests/schemas/basic_model.yaml
+++ b/tests/schemas/basic_model.yaml
@@ -1,11 +1,27 @@
 $schema: http://stsci.edu/schemas/yaml-schema/draft-01
 id: http://example.com/schemas/basic_model
-title: Basic model with a data array and minimal metadata.
+title: Basic model with some data arrays and metadata.
 
 allOf:
   - $ref: core_metadata
   - type: object
     properties:
+      meta:
+        type: object
+        properties:
+          origin:
+            type: string
+          telescope:
+            type: string
       data:
         ndim: 2
         datatype: float32
+        default: 0.0
+      dq:
+        ndim: 1
+        datatype: uint32
+        default: 0
+      area:
+        ndim: 2
+        datatype: float32
+        default: 0.0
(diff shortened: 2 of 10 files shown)


tests/models.py
```diff
@@ -11,6 +11,9 @@ class BasicModel(DataModel):
     """
     schema_url = "http://example.com/schemas/basic_model"
 
+    def get_primary_array_name(self):
+        return "data"
+
```
jdavies-st, line 16: Isn't the default primary array name `data`?  Is this needed?

tests/schemas/basic_model.yaml
```diff
@@ -1,11 +1,27 @@
...
         datatype: float32
+        default: 0.0
+      dq:
+        ndim: 1
+        datatype: uint32
+        default: 0
+      area:
+        ndim: 2
+        datatype: float32
+        default: 0.0
```
jdavies-st, line 27: Might want a return here plus a `...`.  As on the other schemas too.

tests/test_fits.py
```diff
@@ -512,3 +512,100 @@ def test_ensure_ascii():
...
+        assert dm.meta.subarray.xstart == 42
+
+    def test():
+        with _create_source() as dm:
+            with func(dm) as dm2:
+                _check_output(dm2)
+
+    test.__name__ = func.__name__
+
+    return test
```
jdavies-st, line 595: This is a really nutty anti-pattern.  We should fix this by using parametrization.

tests/test_models.py
```diff
@@ -1,341 +1,122 @@
...
-        del dm.data
-
-
-@pytest.mark.skip("base model no longer includes attribute meta.subarray")
-def test_subarray():
-    with DataModel(FITS_FILE) as dm:
-        dm.meta.subarray.xstart
-
+    with BasicModel() as m:
+        del m.data
```
jdavies-st, line 49: What does this do?  It's not a test and doesn't seem to be called by anything.

tests/test_models.py
```diff
@@ -1,341 +1,122 @@
...
-        warnings.simplefilter("always")
-        with datamodels.open(FITS_FILE) as model:
-            pass
-
-        class_name = model.__class__.__name__
-        j = None
-        for i, w in enumerate(warners):
-            if class_name in str(w.message):
-                j = i
-        assert j is not None
```
jdavies-st, line 261: These 2 tests are deleted since `open` is no longer implemented.  We should probably keep these over in `jwst.datamodels`, and clean up that second one to use `with pytest.warns()`.

jdavies-st (approved):
This looks good.  Some comments below.

Also, for all the test that are deleted here, we need to make sure those stick around in `jwst.datamodels`, and to a vice-verse delete of tests over there that are covered here.

tests/models.py
eslavich: Lemme try again without it, I thought it wasn't working without that but I could well have been confused.

eslavich: You're right, works fine without it, will remove.

tests/test_fits.py
eslavich: Is this decorator only being used once?  Maybe we don't even need parametrization.

tests/test_models.py
eslavich: I imagine at some point in history there was an exception when you deleted an ndarray field?

tests/test_models.py
eslavich: Yes there will have to be a wave of cleanup over in jwst once the tests have been sorted between the two packages.

tests/test_fits.py
jdavies-st: Yup.  Only once.

tests/test_models.py
jdavies-st: 🥽

tests/test_models.py
jdavies-st: Peharps.  It can be removed now.

PR #11, opened 2020-10-19
Run docstring tests in tox jobs

eslavich:


jdavies-st (approved):
Looks good.

One could put the test paths `tests` and `src` into the pytest section of `setup.cfg`, or into a section of `tox.ini`.  Then these would be (I think) always set without having to specify them on the command line.  Not sure how this works on testing installed vs source (develop) versions of the codebase.

eslavich:
> One could put the test paths tests and src into the pytest section of setup.cfg

Good call, I'll give that a try...

PR #21, opened 2020-10-29
Compare model type as string when verifying FITS hash

eslavich:
We commented this code out originally because jwst model classes aren't available in stdatamodels.  In this PR it's revived, but compares model classes by their string names so that jwst models aren't required.

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -670,16 +670,9 @@ def _verify_skip_fits_update(skip_fits_update, hdulist, asdf_struct, context):
         return False
 
     # Ensure model types match
-    # TODO: This doesn't work as written in stdatamodels because
-    # we don't have access to the jwst DataModel subclasses.
-    # hdulist_class = util._class_from_model_type(hdulist)
-    hdulist_class = None
-    if hdulist_class is None:
-        log.debug('Cannot determine model of the FITS file.'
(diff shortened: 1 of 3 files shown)


src/stdatamodels/fits_support.py
```diff
@@ -670,16 +670,9 @@ def _verify_skip_fits_update(skip_fits_update, hdulist, asdf_struct, context):
...
-    hdulist_class = None
-    if hdulist_class is None:
-        log.debug('Cannot determine model of the FITS file.'
-                  ' Cannot skip updating from FITS headers.')
-        return False
-    if not isinstance(context, hdulist_class):
-        log.debug(f'Input model {hdulist_class} does not match the'
+    hdulist_model_type = util.get_model_type(hdulist)
+    if hdulist_model_type != context.__class__.__name__:
+        log.debug(f'Input model type {hdulist_model_type} does not match the'
```
jdavies-st, line 675: The tests that check if `_verify_skip_fits_update()` is working correctly are currently disabled and probably need to be rewritten a little to be more general?  Is that worth doing here or doing later in the general test coverage update?

eslavich: I'm angling to enable them in the more general update of the tests for FITS support.  That should be coming in the not too distant future as part of AL-407.

jdavies-st (approved):

PR #22, opened 2020-11-02
Don't use ASDF ndarray validators for now

eslavich:
This PR reverts the change in this repo that enabled the ASDF ndarray validators.  It turns out those aren't compatible with some of the jwst table schemas, which include columns with nested arrays that aren't yet supported by ASDF.

We're just going to keep using the ndarray validation in `properties._cast` for now, and switch to asdf built-in validators once they support all the jwst use cases.


jdavies-st (approved):
Looks good!  Thanks.

Such a simple change.  =)

PR #24, opened 2020-11-03
Initialize documentation

eslavich:
This creates a bare-bones docs directory, currently with nothing but the auto-generated API documentation for `DataModel`.  I looked into migrating documentation over from jwst but what's there is very jwst-specific and I don't have the time to rewrite it right now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,3 +3,4 @@ requires = ["setuptools>=42", "setuptools_scm[toml]>=3.4", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
+write_to = "src/stdatamodels/_version.py"
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
 #!/usr/bin/env python3
 from setuptools import setup
 
-setup(use_scm_version=True)
+setup()
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,9 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# setuptools-scm generated module
+src/stdatamodels/_version.py
+
+# auto-generated API docs
+docs/source/api
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,16 @@
+version: 2
+
+sphinx:
+  configuration: docs/source/conf.py
+
+formats:
+  - htmlzip
+  - pdf
+
+python:
+  version: 3.8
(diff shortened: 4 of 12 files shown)


pyproject.toml
```diff
@@ -3,3 +3,4 @@ requires = ["setuptools>=42", "setuptools_scm[toml]>=3.4", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
+write_to = "src/stdatamodels/_version.py"
```
jdavies-st, line 6: Is there a reason to write this file out?  In the past, we've just had code in `__init__.py` that sets it using `pkg_resources`.  I.e. for `jwst`:

https://github.com/spacetelescope/jwst/blob/c04bc7e1bcd7cb96c34308771a89a5e39abccf49/jwst/__init__.py#L8-L11

Thought guess letting `setuptools_scm` do all the work does make `__init__.py` for the package a bit simpler.  Is there any downside in writing out a `_version.py` file?

setup.py
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env python3
 from setuptools import setup
 
-setup(use_scm_version=True)
+setup()
```
jdavies-st, line 4: Keeping `use_scm_version=True` here can't hurt for backwards compatibility, right?

jdavies-st (commented):
Looks good!  Just a question about the version changes and backwards compatibility.

pyproject.toml
eslavich: I'm not sure if there are any downsides, this is the setuptools_scm documentation on that option:

```
A path to a file that gets replaced with a file containing the current version. It is ideal for creating a version.py file within the package, typically used to avoid using pkg_resources.get_distribution (which adds some overhead).
```

and astropy uses this feature: https://github.com/astropy/astropy/blob/master/setup.py#L69

setup.py
eslavich: You mean compatibility with older setuptools versions?  We have a requirement in the pyproject.toml that requires a sufficiently recent setuptools:

```
[build-system]
requires = ["setuptools>=42", "setuptools_scm[toml]>=3.4", "wheel"]
```

The `jwst` package also requires `setuptools>=42`.

jdavies-st: Hmm, yeah.  `setuptools 42` is only a year ago, but I guess this is fine since so many other package requirements are bleeding edge too.  👍

jdavies-st (approved):

PR #27, opened 2020-11-05
Catch up to latest jwst master

### LINKED ISSUE #3: Handle null serialization for asdf 2.8
`stdatamodels` was forked from `jwst.datamodels` before null serialization was fixed to work with `asdf 2.8+` in PR https://github.com/spacetelescope/jwst/pull/5371

Those changes should be ported over here.

eslavich:
The only relevant change turned out to be support for asdf 2.8's change of opinion on null values.  There was also a logging function added to model_base.py, but its use was commented out and is seemingly only to be enabled when debugging, which doesn't work well now that this code is in one of jwst's dependencies.

Branched off of #10.  Resolves #3

### DIFF
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -17,8 +17,15 @@ from astropy.wcs import WCS
 
 import asdf
 from asdf import AsdfFile
-from asdf import yamlutil
+from asdf import yamlutil, treeutil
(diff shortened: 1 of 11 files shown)


jdavies-st (approved):
🚀 

Though pending review/changes to #10, maybe hold off merging this until that PR is resolved?

PR #30, opened 2020-11-06
Update tests in test_schema.py

eslavich:
These tests were:

- Updated to use test models defined in this package
- Removed if they were only relevant to jwst, or if they duplicated coverage in test_validation.py and test_util.py
- Moved to test_fits.py (the final frontier) if they didn't really belong in test_schema.py

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1,278 +1,63 @@
-# Licensed under a 3-clause BSD style license - see LICENSE.rst
-
 from datetime import datetime
-import os
-import shutil
-import tempfile
-import warnings
 
 import pytest
 import numpy as np
-from numpy.testing import assert_array_equal, assert_array_almost_equal
 
 import jsonschema
-from astropy.io import fits
 from astropy.modeling import models
 from astropy import time
 
-from stdatamodels import util, validate
-# from .. import _defined_models as defined_models
-# from .. import (DataModel, ImageModel, RampModel, MaskModel, MultiSlitModel,
-#     AsnModel, CollimatorModel, SourceModelContainer, MultiExposureModel,
-#     DrizProductModel, MultiProductModel, MIRIRampModel)
-from stdatamodels import DataModel
 from stdatamodels.schema import merge_property_trees, build_docstring
-
-# from ..extension import URL_PREFIX
+from stdatamodels import DataModel
 
 import asdf
-from asdf import schema as mschema
 
-FITS_FILE = None
-MASK_FILE = None
-TMP_FITS = None
-TMP_FITS2 = None
-TMP_YAML = None
-TMP_ASDF = None
-TMP_DIR = None
-
-def setup():
-    global FITS_FILE, MASK_FILE, TMP_DIR, TMP_FITS, TMP_YAML, TMP_ASDF, TMP_FITS2
-    ROOT_DIR = os.path.join(os.path.dirname(__file__), 'data')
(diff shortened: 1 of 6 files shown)


tests/test_schema.py
```diff
@@ -1,278 +1,63 @@
...
         assert dm2.meta.foo == {'a': 42, 'b': ['a', 'b', 'c']}
 
 
-@pytest.mark.skip("requires the jwst core schema")
 def test_find_fits_keyword():
-    with DataModel() as x:
-        assert x.find_fits_keyword('DATE-OBS') == \
-          ['meta.observation.date']
+    with FitsModel() as x:
+        assert x.find_fits_keyword('TELESCOP') == ['meta.telescope']
```
jdavies-st, line 44: I know the code for `find_fits_keyword` is in `schema.py`, but it might worth moving the code and the test for it into `fits_support.py` and `test_fits_support.py` at some point.

tests/test_schema.py
```diff
@@ -578,87 +137,27 @@ def test_add_schema_entry():
...
-                        strict_validation=True)
-    m.meta.description = "Test validate a WCS reference file."
-    m.meta.author = "ND"
-    m.meta.pedigree = "GROUND"
-    m.meta.useafter = "2018/06/18"
-    m.meta.reftype = "collimator"
+    m = TransformModel(
+        transform=models.Shift(1) & models.Shift(2),
+        strict_validation=True
+    )
```
jdavies-st, line 147: Maybe test some of the metadata here too?  I think that was the purpose of the other test, as it also tested some of the required metadata for the reference file?

tests/test_schema.py
```diff
@@ -578,87 +137,27 @@ def test_add_schema_entry():
...
     m.validate()
 
 
-@pytest.mark.skip("requires CollimatorModel")
-def test_validate_transform_from_file():
+def test_validate_transform_from_file(tmp_path):
```
jdavies-st, line 151: This test looks like it is round-tripping the previous test.  Perhaps they can be combined in the same test?

tests/test_schema.py
```diff
@@ -1,278 +1,63 @@
...
-
-    TMP_DIR = tempfile.mkdtemp()
-    TMP_FITS = os.path.join(TMP_DIR, 'tmp.fits')
-    TMP_FITS2 = os.path.join(TMP_DIR, 'tmp2.fits')
-    TMP_YAML = os.path.join(TMP_DIR, 'tmp.yaml')
-    TMP_ASDF = os.path.join(TMP_DIR, 'tmp.asdf')
-
-
-def teardown():
-    shutil.rmtree(TMP_DIR)
```
jdavies-st, line 53: So nice to see this gone.  👋

jdavies-st (approved):
Looks good.  Just some minor comments.

tests/test_schema.py
eslavich: https://github.com/spacetelescope/stdatamodels/issues/33

tests/test_schema.py
eslavich: That's a good point, I'm not sure why it's necessary to test validation separately when the DataModel is read from a file.  I'll mash these two together.

tests/test_schema.py
eslavich: I removed most of the metadata validation tests from this file because there's good coverage of the same code in test_validation.py.  Probably that module and this should be merged at some point?

PR #31, opened 2020-11-06
Update tests in test_fits.py

eslavich:
This went pretty smoothly in the end.  I didn't make any attempt to improve the tests, just coaxed the relevant specimens into passing.  We can take another pass at cleaning them up once we know how we're going forward with tags or not-tags.

### DIFF
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -74,10 +74,6 @@ def test_from_scratch(tmp_path):
             assert np.all(dm2.data == data)
 
 
-def _header_to_dict(x):
-    return dict((a, b) for (a, b, c) in x)
-
-
 def test_extra_fits(tmp_path):
     file_path = tmp_path/"test.fits"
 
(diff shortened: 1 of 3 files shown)


tests/test_fits.py
```diff
@@ -84,182 +51,93 @@ def test_setting_arrays_on_fits():
...
-    fits = fits.HDUList(
-        [fits.PrimaryHDU(),
-         fits.ImageHDU(name='COEFFS', data=np.array([0.0], np.float32))])
-
-    with DataModel(fits, schema=schema) as dm:
-        assert_array_equal(dm.coeffs, [0.0])
-
-
 def _header_to_dict(x):
     return dict((a, b) for (a, b, c) in x)
```
jdavies-st, line 78: Is this needed anymore now that it has been removed from `test_extra_fits`?

jdavies-st (approved):
🛋️ 
☝️

tests/test_fits.py
eslavich: No sir!  I shall remove it.

PR #35, opened 2020-11-09
Add xfail'ed test for nested null values

eslavich:
I was daydreaming about DataModel validation (doesn't everybody?) and thought of this case.  If we wanted to fix it I think we'd need to override jsonschema's `properties`, `patternProperties`, `required`, and `additionalProperties` validators to all treat `None` as absent.  We can do that, but it's a bit of a hassle so I think we ought to wait and see if we give up on validating on assignment entirely.

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -72,6 +72,15 @@ def test_list_attribute_ssignment():
     assert len(model.meta.list_attribute) == 0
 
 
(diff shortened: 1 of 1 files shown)


jdavies-st (approved):
Interesting.  I didn't realize this didn't get validated correctly, or should I say, I didn't realize null values weren't handled the same in dict assignments.

PR #36, opened 2020-11-11
Attr validation

### LINKED ISSUE #6: No validation occurs when a DataModel is written to ASDF
When a DataModel is written to a FITS file, the model is validated along the way due to walking the schema to match the tree nodes to FITS header keywords.  When writing to ASDF, there is no validation.

kmacdonald-stsci:
Added a `validate_on_assignment` option when creating a `DataModel` instance.  It defaults to `True`, which would cause any change to be validated against the data model schema upon assignment.  When set to `False` schema validation only occurs when saving to a file.  A warning is raised if a schema error is found.

Resolves spacetelescope/jwst#5430 / [JP-1760](https://jira.stsci.edu/browse/JP-1760)
Resolves #6


src/stdatamodels/model_base.py
```diff
@@ -138,8 +146,12 @@ def __init__(self, init=None, schema=None, memmap=False,
         if strict_validation is None:
             strict_validation = get_envar_as_boolean("STRICT_VALIDATION",
                                                           False)
+        if validate_on_assignment is None:
+            validate_on_assignment = get_envar_as_boolean("VALIDATE_ON_ASSIGNMENT",
+                                                          True)
```
jdavies-st, line 151: Since the default is `validate_on_assignment=True`, the env var will never have an effect unless explicitly set to `None`.  I.e.:

```python
DataModel(validate_no_assignment=None)
```

So it might be better to have the default be `None`.

kmacdonald-stsci: I changed the default to `None` in order to use environment variables and everything passed pytest.

src/stdatamodels/model_base.py
```diff
@@ -113,6 +114,13 @@ def __init__(self, init=None, schema=None, memmap=False,
             If `None`, value will be taken from the environmental STRICT_VALIDATION.
             Otherwise, the default value is `False`.
 
+        validate_on_assignment : bool
+            Defaults to 'True'.
+            If 'True', attribute assignments are validated at the time of assignment.
+            Validation errors generate warnings.
+            If 'False', schema validation occurs only once at the time of write.
+            Validation errors generate warnings.
```
jdavies-st, line 122: You'll want to update this to `bool or None` and explain what the default of None means.  Basically, it will follow the same pattern as the `strict_validation` kwarg.

src/stdatamodels/properties.py
```diff
@@ -396,24 +403,33 @@ def __setslice__(self, i, j, other):
...
             self._instance.append(item)
 
+
     def insert(self, i, item):
         schema = _get_schema_for_index(self._schema, i)
         item = _cast(item, schema)
         node = ObjectNode(self._name, item, schema, self._ctx)
-        if node._validate():
+        if self._ctx._validate_on_assignment:
+            if node._validat():
```
jdavies-st, line 430: `note._validate():` typo

tests/test_validation.py
```diff
@@ -244,3 +243,24 @@ def test_validation_on_write(tmp_path):
...
+    model.meta.string_attribute = 42  # Bad assignment should cause no warning
+
+    with pytest.warns(ValidationWarning):
+        model.save(file_path)
+
+
+def test_validate_on_assignment_with_environ(tmp_path):
+    file_path = tmp_path/"test.asdf"
+
+    os.environ['VALIDATE_ON_ASSIGNMENT'] = '0'
```
jdavies-st, line 269: Instead of using `os.environ`, try using Pytest's `monkeypatch` fixture, which can set an env var only for this test, while leaving the environment of the user (and other tests which might be running in parallel) unchanged.  So something like:

```python
def test_validate_on_assignment_with_environ(tmp_path, monkeypatch):
    file_path = tmp_path/"test.asdf"

    monkeypatch.setenv("VALIDATE_ON_ASSIGNMENT", False)
```

jdavies-st (commented):
Looking good.  Would be nice to test all the permutations:

```
VALIDATE_ON_ASSIGNMENT = True, False, unset
validate_on_assignment = True, False, None
```
And make sure they are producing the correct result.

src/stdatamodels/model_base.py
kmacdonald-stsci: I expanded the docstring to explain the usage of and default to `None`.

src/stdatamodels/properties.py
kmacdonald-stsci: Corrected typo.

tests/test_validation.py
kmacdonald-stsci: Used `monkeypatch` instead of `os.environ` and expanded the tests to capture unset conditions, as well as for `True` and `False` for the environment variable and parameter for class instantiation.

src/stdatamodels/fits_support.py
```diff
@@ -476,21 +476,27 @@ def callback(schema, path, combiner, ctx, recurse):
...
+                else:
                     properties.put_value(path, result, tree)
 
         elif 'fits_hdu' in schema and (
                 'max_ndim' in schema or 'ndim' in schema or 'datatype' in schema):
             result = _fits_array_loader(
                 hdulist, schema, ctx.get('hdu_index'), known_datas)
 
-            if result is None:
+            if result is None and context._validate_on_assignment:
```
eslavich, line 493: This if/else block is identical to the one in the previous case, is it work defining a little inner function to do this?

src/stdatamodels/model_base.py
```diff
@@ -120,6 +121,15 @@ def __init__(self, init=None, schema=None, memmap=False,
             If `None`, value will be taken from the environmental STRICT_VALIDATION.
             Otherwise, the default value is `False`.
 
+        validate_on_assignment : bool or None
+            Defaults to 'None'.
+            If `None`, value will be taken from the environmental VALIDATE_ON_ASSIGNMENT,
+            defaulting to 'True' if  no environment variable is set.
+            If 'True', attribute assignments are validated at the time of assignment.
+            Validation errors generate warnings and values will be set to `None`.
```
eslavich, line 129: Doesn't the warning vs exception still depend on the value of `strict_validation`?  Or do we always get warnings when `validate_on_assignment` is set?

tests/test_validation.py
```diff
@@ -130,7 +130,6 @@ def test_pass_invalid_values_attribute_assignment(monkeypatch, init_value, env_v
         assert model.meta.string_attribute is None
 
 
-@pytest.mark.xfail(reason="validation on write not yet implemented for ASDF files", strict=True)
```
eslavich, line 133: Hooray!  Thanks for fixing that too.

tests/test_validation.py
```diff
@@ -208,3 +206,120 @@ def test_validation_on_write(tmp_path):
...
+    assert model.meta.string_attribute is None
+
+
+def test_validate_on_assignment_false(tmp_path):
+    file_path = tmp_path/"test.asdf"
+
+    model = ValidationModel(validate_on_assignment=False)
+    assert model._validate_on_assignment==False
+
+    model.meta.string_attribute = 42  # Bad assignment should cause no warning
```
eslavich, line 225: In order to assert no warnings, we need to do something like this:
```python
with pytest.warns(None) as warnings:
  model.meta.string_attribute = 42
assert len(warnings) == 0
```
otherwise the test succeeds, since warnings don't interrupt the program like exceptions do.  Another option is to enable strict validation.

kmacdonald-stsci: I added this structure to ensure no warnings in the various tests where that is expected.

tests/test_validation.py
```diff
@@ -208,3 +206,128 @@ def test_validation_on_write(tmp_path):
...
+    model = ValidationModel(validate_on_assignment=False)
+    assert model._validate_on_assignment==False
+
+    with pytest.warns(None) as warnings:
+        model.meta.string_attribute = 42  # Bad assignment should cause no warning
+    assert len(warnings) == 0
+    assert model.meta.string_attribute == 42
+
+    with pytest.warns(ValidationWarning):
+        model.save(file_path)
```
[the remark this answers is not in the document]
kmacdonald-stsci: I have added tests for these situations.  An exception is raised if `strict_validation=True`.

tests/test_validation.py
```diff
@@ -208,3 +206,296 @@ def test_validation_on_write(tmp_path):
...
+def test_validate_on_assignment_default():
+    model = ValidationModel()
+    assert model._validate_on_assignment==True
+
+    with pytest.warns(ValidationWarning):
+        model.meta.string_attribute = 42  # Bad assignment
+    assert model.meta.string_attribute is None
+
+
+def test_validate_on_assignment_false(tmp_path):
```
jdavies-st, line 219: It would be good to combine this with the previous test and parametrize.  This true for a number of tests below too.

tests/test_validation.py
```diff
@@ -208,3 +206,296 @@ def test_validation_on_write(tmp_path):
...
+    model = ValidationModel(validate_on_assignment=False)
+    assert model._validate_on_assignment==False
+
+    with pytest.warns(None) as warnings:
+        model.meta.string_attribute = 42  # Bad assignment should cause no warning
+    assert len(warnings) == 0
+    assert model.meta.string_attribute == 42
+
+    with pytest.warns(ValidationWarning):
+        model.save(file_path)
```
jdavies-st, line 231: Looking through the tests more carefully, I see there's nothing unique about writing out an invalid tree when `validate_on_assignment=True`.  An invalid tree is an invalid tree, so `test_validation_on_write` test above should be sufficient to test this.  Was good to verify this while writing tests, but in terms of the final test code, I think this test case is covered.

jdavies-st (commented):
The runtime code is good.  Tests could use a bit of clean up, mostly to be able to see what the expected behavior is at a glance, without having to read through 4 verbose tests.  I think parametrization can work well here and actually had to try it out to understand what was going on in these tests.  So feel free to look at my PR to this branch

https://github.com/kmacdonald-stsci/stdatamodels/pull/1

and if it is OK, go ahead and merge.  If you have questions, let's discuss them there, resolve them, and then once we are happy, merge. Then this can be merged.  Happy to have input from @eslavich as well on the above parametrization cleanup.

eslavich:
> Happy to have input from @eslavich as well on the above parametrization cleanup.

I left a comment on the PR on Ken's fork, the test updates look great to me.

tests/test_validation.py
kmacdonald-stsci: I merged your PR to my branch.  I liked the way your refactored my tests.

jdavies-st (approved):
This is merge-ready.  🚀

eslavich:
I'll add a changelog entry in a separate commit

PR #38, opened 2020-11-11
Add jwst and romancal jobs to CI

### LINKED ISSUE #34: Add jwst and romancal tests as allowed failure jobs
Just so we know when a change here necessitates an update to either of our two main customers...

jdavies-st:
This adds 2 new `allow_failures` jobs to the Travis matrix that run the `jwst.datamodels` and `romancal.datamodels` tests.  This will give us a heads-up if anything in an `stdatamodels` PRs breaks something in the dev versions of those 2 packages.

Only running the `.datamodels` tests now for each package for efficiency.  And running on the installed versions of the packages.

Also added a Python 3.9 job.

Resolves #34

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,14 @@
 os: linux
 dist: xenial
 language: python
-python: 3.8
+python: 3.8.5
 
 jobs:
   include:
+    - name: Python 3.9
+      python: 3.9.0
+      env: TOXENV=py39
+
     - name: Python 3.8
       env: TOXENV=py38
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,10 +1,14 @@
 os: linux
 dist: xenial
 language: python
-python: 3.8
+python: 3.8.5
```
eslavich, line 4: It's annoying that Travis doesn't update the `3.8` alias when a new patch comes out.  Any reason not to specify `3.8.6` here?

jdavies-st: Doh!  Good point.  Done.

PR #43, opened 2020-11-23
Proof of concept for roman tags

perrygreenfield:
The locations and names of the new files are not intended to be final in any way. In fact much of this may move to a different repository. The following will highlight the files of note.

- stnode.py: contains the code for the object nodes as well as the converters for those objects. This enables the access to ASDF attributes as python attributes. There is much repetitive structure in the converter code, and it is likely that this could be greatly reduced.

- stuserdict.py: uses the code from UserDict, except that it puts the dict in ._data instead of .data, which I have a feeling we may want to use as an attribute ;-)

- exposure.py, integration.py are updated to use the new objects.

- setup.cfg (in the home directory) defines the entry points needed to enable the converters and such.

- There is a schemas/roman subdirectory for the needed schemas that define the tags.

- The manifest/datamodels-1.0.0.yaml file has been updated to include the schemas.

- stnode_test.py: has code to define a python tree to be used to create an  ASFD file along the lines of the schema that was used as a basis.

The code to create the file basically consists of (after installing stdatamodels as updated)

import stdatamodesl.stnode_test as snt
rd = snt.mk_test_object()
import asdf
af  =  asdf.AsdfFile()
af.tree = {'roman': rd} # Attribute to put the roman data in?
af.write_to('zzz.asdf') 

af2 = asdf.open('zzz.asdf')
af2.tree['roman'].meta.exposure.nframes
type(af2.tree['roman'].meta.exposure) # stdatamodesl.stnode.Exposure object.


eslavich:
The concept has been realized!

PR #44, opened 2020-11-23
Remove automatic management of meta.date and introduce on_init hook

eslavich:
This removes automatic mangement of the meta.date attribute from this package and introduces an on_init hook that will let us re-implement the feature in jwst and romancal.

Here's what the updates to jwst will look like:

https://github.com/spacetelescope/jwst/compare/master...eslavich:AL-420-implement-meta-date-management

And for romancal:

https://github.com/spacetelescope/romancal/compare/main...eslavich:AL-420-meta-date-management

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -192,19 +192,18 @@ def test_update_from_dict(tmp_path):
 
 
 def test_object_node_iterator():
-    m = BasicModel()
+    m = BasicModel({"meta": {"foo": "bar"}})
     items = []
     for i in m.meta.items():
         items.append(i[0])
 
-    assert 'date' in items
-    assert 'model_type' in items
+    assert 'foo' in items
 
 
 def test_hasattr():
-    model = DataModel()
-    assert model.meta.hasattr('date')
-    assert not model.meta.hasattr('filename')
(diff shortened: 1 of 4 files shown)


tests/test_models.py
```diff
@@ -269,3 +268,27 @@ def __init__(self, *args, **kwargs):
...
+
+def test_on_save_hook(tmp_path):
+    class OnSaveModel(DataModel):
+        def on_save(self, init):
+            super().on_save(init)
+
+            self.meta.foo = "bar"
+
+    model = OnSaveModel()
+    assert model.meta.get("foo") is None
```
jdavies-st, line 292: Why use `get()` here instead of 

```python
assert model.meta.foo is None
```
?

jdavies-st (approved):
Looks good.  Since this will cause a unit test failure in `jwst`:

https://travis-ci.com/github/spacetelescope/stdatamodels/jobs/446101262#L254

It might be good to open that PR immediately after this is merged.

tests/test_models.py
eslavich: `foo` isn't in the schema, so retrieving its value before it has been set will cause an error

PR #45, opened 2020-12-01
Remove nested None values before validating

eslavich:
This fixes validation behavior when an object with nested None values is assigned to a DataModel.

Resolves spacetelescope/jwst#5491

### DIFF
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -17,15 +17,7 @@ from astropy.wcs import WCS
 
 import asdf
 from asdf import AsdfFile
-from asdf import treeutil
 from asdf import schema as asdf_schema
-
-try:
-    from asdf.treeutil import RemoveNode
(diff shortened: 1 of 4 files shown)


eslavich:
No failures now in the jwst regression tests with dev dependencies:

https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/51/console

jdavies-st (approved):
Looks good.  Thanks for the elegant fix!

I notice that we are not testing against `asdf~=2.7.1`.  I think it would be good to do so, even if we have to skip some tests here to do so.

I.e. `stdatamodels` has a testing dependency on `asdf>=2.8` but runtime dependency of `asdf~=2.7.1` is fine.  It would be good to maintain that runtime dependency by testing it so that we don't break it in the future.  Currently this runtime dependency is only being tested by the `jwst` unit and nightly regression tests.

PR #46, opened 2020-12-01
example JWST data file using tags

perrygreenfield:
This requires a fair amount of explanation and context, and what was tested (manually).

First I took the example tutorial nircam image file, read it in, deleted all large arrays, wrote the rest out with remaining arrays as inline (I wished to be able to edit directly; asdftool explode was not useful since gwcs will end up with many external arrays, and there is no need to deal with that complication). All my testing was done on this pure yaml file (and I left it with a yaml extension to take advantage of syntax highlighting in my editor).

I proceeded to define new tags one at a time using existing JWST schemas content copied into the new schemas, and create the associated converter and class code (I'm sure that could be made much faster with metaclasses or some such, but I didn't want to spend a lot of time fiddling with that if this is not decided to be useful). I plowed all that into the same stnode.py file I used for the Roman example (ultimately they should not be together). I created a schemas/jwst directory under resources, and all the new stuff is in the same manifest file and extensions.py file.

I used my own naming scheme for the tags and schema files. Usually similar to the attribute names, but with one major difference; in particular, many now are nircam specific. That was not necessary for testing how well this works within the existing pipeline, but I think if it were done for JWST, this the way it would be done for reasons I'll go into later. I did not remove irrelevant items from the pasted schema content (for reasons given later, again). 

As previously mentioned, in order to use a tag for the full content, everything has been moved down a level (under the attribute "jwst" (I don't know what the ultimate attribute name should be yet). This has consequences :-). This asdf file is opened directly in asdf, not datamodels. The asdf file I have been testing against is `resources/data/jwst_tag.yaml` to verify that it opens fine and creates the node objects I expect. I haven't renamed the basic node parent class to match that in JWST data models yet. The attribute access should be identical to the JWST datamodel case except that for the time being the tree attribute 'jwst' must be referenced and all attributes attached to that. I believe we can figure out a change to asdf to permit the asdfobject.attribute access that datamodels provides now, but I'm pretty sure that requires a change to asdf.

There are a handful of attributes in the file that are not covered by any schema in my example (e.g., bunit stuff) but nearly all of it is covered by tags and schemas. There is no use of required yet, but ultimately, that should be present for all required attributes.

### Where is all this going?

Good question. In all the excitement I kind of lost track myself. The question you have to ask yourself is do you feel luck...Sorry, got sidetracked. Can this be used in the current pipeline with minimal changes to current pipeline code? I suspect yes. Is that needed? I'm not sure. We really need to understand what we think will be shared between Roman and JWST. Do we really think the schemas will be identical, for example? Or even very similar? If they aren't what is really being shared? If algorithmic code only, then we will need to rework JWST code to make that more possible. In other words, I'm not convinced that both Roman and JWST need to use the same datamodel code yet.

On the other hand, having the same system for ASDF files for both would be good. Enabling that for JWST will entail mostly work on the schema/data model side, and I'm guessing it would be at least a moderate, and perhaps a significant, amount of work. To do it right I think means coming up a schema arrangement that is generated from SDP schemas so that there is automatic consistency, and more rigorous schemas targeted specifically to instrument/observing mode combinations. That would result in more schema files, but in this case the schema files would be much, much closer descriptions of what is expected than is currently specified. Right now, I can't tell in any particular case whether an attribute (or FITS keyword) is really needed or not for a given case, or whether I can count on it being there. 

I see no reason to do this narrowing of the example schemas since doing it right is nontrivial, and ultimately they should be generated from the SDP schemas, which do correspond much more closely to what is expected (I believe they have a small amount of code to deal with cases that schemas can't handle).

I've gone on long enough and will stop here.


eslavich:
I believe this is no longer needed but do reopen if I guessed wrong...

PR #51, opened 2020-12-21
Validateassign

perrygreenfield:
Proof of concept of validate on assignment for tagged datamodels


src/stdatamodels/stuserdict.py
```diff
@@ -0,0 +1,94 @@
...
+#     pass
+# else:
+#     _collections_abc.MutableSequence.register(deque)
+
+# try:
+#     from _collections import defaultdict
+# except ImportError:
+#     pass
+
+class STUserDict(_collections_abc.MutableMapping):
```
eslavich, line 25: This seems to be identical to the 3.8.0b1 version of [collections.UserDict](https://github.com/python/cpython/blob/v3.8.0b1/Lib/collections/__init__.py#L971), do we need that version for some reason?  Maybe we can just use the built-in UserDict?

src/stdatamodels/stnode.py
```diff
@@ -0,0 +1,565 @@
...
+        """
+        if key[0] != '_':
+            if key in self._data:
+                print('***** (setattr)', key)
+                if validate:
+                    print('._data', self._data)
+                    self._schema()
+                    schema = self._x_schema.get('properties').get(key, None)
+                    print('schema from __setattr__:', key, schema)
+                    if _validate(key, value, schema, self.ctx):
```
eslavich, line 143: Is the intention to eventually allow users to e.g. assign a dict to the jwst.meta.exposure key and automatically convert it to an `Exposure` object?  Or will the dict always fail validation here (due to the schema's tag validator)?

src/stdatamodels/stnode.py
```diff
@@ -0,0 +1,565 @@
...
+            print('---------- (getattr) key = ', key)
+            value = self._data[key]
+            if isinstance(value, dict):
+                return DNode(value, parent=self, name=key)
+            elif isinstance(value, list):
+                return LNode(value)
+            else:
+                return value
+        else:
+            raise KeyError(f"No such key ({key}) found in node")
```
eslavich, line 129: So if a node key like jwst.meta.exposure is missing entirely, and a user attempts to access jwst.meta.exposure.type, they'll receive a `KeyError` and not just a `None` value right?

eslavich (commented):
I had some questions about details, but overall I think this proves that the concept is workable

src/stdatamodels/stuserdict.py
perrygreenfield: Which version to use is a good question. It took me a few minutes to recall why I couldn't use the built in, but then it hit me like a 2x4. The problem is the built-in has a data attribute, and that would conflict with our need for a data attribute (if we didn't use it, someone would), so the only difference in this version is that `.data` was changed to `._data`. Is there a better way to do it?

src/stdatamodels/stnode.py
perrygreenfield: Not at the moment. Currently it should require an Exposure object, though eventually I suppose that could be permitted. I would have to think about it.

src/stdatamodels/stnode.py
perrygreenfield: Yes, I believe so. That is the behavior we want now isn't it? Though we should work out how we deal with permitting null as values we handle. In other words, if we want `None`, we should be explicit about it.

src/stdatamodels/stuserdict.py
eslavich: Ohhhhhh ok!  That makes sense.  We may be able to get away with subclassing dict directly, since (at least in this code so far) we're not overriding any of the MutableMapping methods.

eslavich:
This is pre-roman-datamodels work, I believe, so I'll close the PR for now.

PR #52, opened 2020-12-23
Add actions workflow for tox jobs

eslavich:
This replaces our Travis build with an actions workflow.  The jwst.datamodels failure is a known issue that will be resolved by https://github.com/spacetelescope/jwst/pull/5512


jdavies-st (approved):
Looks good.

PR #54, opened 2021-01-29
AL-456 Manage resources among shallow clones

stscieisenhamer:
Resolves [AL-456](https://jira.stsci.edu/browse/AL-456)


stscieisenhamer:
PR mainly for review/discussion purposes. There is probably some better containerizing on the open/close concepts.

src/stdatamodels/model_base.py
```diff
@@ -346,31 +348,35 @@ def close(self):
...
+                if hasattr(self, "_instance"):
+                    self._drop_arrays()
+
+                if hasattr(self, "_files_to_close"):
+                    for fd in self._files_to_close:
+                        if fd is not None:
+                            fd.close()
 
     @staticmethod
     def clone(target, source, deepcopy=False, memo=None):
```
eslavich, line 368: Would it be terrible to re-open the file, rather than share file pointers among DataModel instances?  `HDUList.fileinfo` seems to make the original file path available, as does `AsdfFile.uri`.

In any case, I'm in favor of forcing some kind of deeper copy here, so that DataModel instances never have to share file pointers.  The tricky case is when the arrays are memory mapped and potentially too large to be copied...

stscieisenhamer: New file descriptors would be good, but not sure how that would work? Was not sure how the whole buffering would work. I'm under the impression that it would be more than just creating a new file descriptor. But have not delved deep enough to know. One could go through the opening process, but the situation may be that some things may have changed in the source model and one would not want to loose that.

About forcing deepcopy: This could quickly introduce memory issues and, at least for jwst pipeline, some easy, but significant rewrite. With the current fix, which is simple, or something with re-opening, which may also be very simple, loosing the feature of buffering seems unnecessary. Unless there is some major drawback to the resource management.

eslavich: There are some perhaps unexpected consequences of doing a shallow copy, for example:
```python
In [12]: from stdatamodels import DataModel
    ...:
    ...: model = DataModel()
    ...: model.meta.filename = "original.asdf"
    ...: clone = DataModel(model)
    ...: clone.filename = "clone.asdf"

In [13]: model.filename
Out[13]: 'clone.asdf'
```
Maybe what we want is a deep copy of the tree structure *except* arrays, which can remain references to the same objects.

As for arrays, what do you think about making them read-only somehow if they were read from a file?  This would force the user to reassign an array if they wanted to transform it somehow.  Would protect them from accidentally modifying the original model.

stscieisenhamer: Well, depends on what is meant by "unexpected": Python-wise, for "shallow copy", this behavior is very expected. Question is, is this expected DataMode-wise?

The issue may be in how DataModel instantiation has been used, as alluded to earlier: There are two use cases that JWST has been using. First is with `jwst.datamodels.open`:

```
def process(input):
    with jwst.datamodels.open(input) as local_model:
```

and

```
def process(input):
    with SomeModel(input) as local_model:
```

This is done mainly as convenience to handle when input is either just a string or DataModel. A shallow copy is returned so that when the context manager exits, input is not closed. But otherwise, one does not want any deeper copying to occur and the side-effects of the shallow copy are completely expected. Note that results from nearly all such situations are returned in separately created models.

This was actually managed already, though incompletely, by the `DataModel._iscopy` attribute. So, resource management has always been apart of DataModels.

That being said, the following is, most likely mis-informed, options:

First: Not sure that shallow copying of DataModels makes sense at all. I would be all for not having this feature.

Second: The utility function, `jwst.datamodes.open`, when given an already exiting DataModel, simply returns the model. When used as a context manager, not closing the already existing DataModel can be handled. This is possibly already the existing `_iscopy` attribute, but without the shallow copy.

Finally, the case of re-instantiation: Not sure about this. In reality, this is effectively casting. The simplest, and if "no shallow copy allowed" is OK, then this is easy: A deep-copy and initialization cast is done. Even in the case where the cast is to the same class.

As such, the only real change would be to not allow shallow copy and to ensure the context management/_iscopy semantic holds.

eslavich: I like the idea of casting.  Maybe at some point we should add a `DataModel.cast` method and arrange to have it make the previous instance of the model inoperable?

stscieisenhamer: and down the rabbit-hole we go...:scream:

eslavich:
Closing in favor of #59

PR #57, opened 2021-02-12
Add CRDS method implementations to be DataModel class

### LINKED ISSUE #56: Add default implementation/documentation of get_crds_parameters and crds_observatory
The stpipe package now relies upon the presence of the `get_crds_parameters` method and `crds_observatory` property, but neither of these is implemented or documented in stdatamodels.

eslavich:
These two methods are needed by stpipe.  I thought we'd better add them here as a guide for anyone who needs to implement a DataModel subclass in the future.

Resolves #56


jdavies-st (approved):
🚀

PR #58, opened 2021-02-12
Recognize asdf.tags.core.NDArrayType as an array

### LINKED ISSUE #55: DataModel.to_flat_dict does not properly handle NDArrayType instances
Since `NDArrayType` does not subclass `np.ndarray`, the check for arrays doesn't catch them.

eslavich:
The `asdf.tags.core.NDArrayType` class does not actually subclass `np.ndarray`, so DataModel wasn't removing them when it should have been.

Resolves #55

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@
 - Remove automatic management of meta.date attribute and create
   on_init hook. [#44]
 
+- Fix bug where asdf.tags.core.NDArrayType instances remain
+  in flat dict when include_arrays=False. [#58]
+
 0.1.0 (2020-12-04)
 ==================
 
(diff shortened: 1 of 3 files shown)


jdavies-st (approved):
Looks good!

Any thoughts on why `NDArrayType` doesn't (or shouldn't) subclass `np.ndarray`?

eslavich:
> Any thoughts on why NDArrayType doesn't (or shouldn't) subclass np.ndarray?

NDArrayType is one of those asdf library "I'm both a serializer and the thing that I serialize" classes.  For that reason it already needs to subclass `AsdfType`, I wonder if someone just didn't want to fall into the multiple inheritance quagmire.

PR #59, opened 2021-02-12
Improve management of open files among shallow DataModel copies

eslavich:
This is a riff on https://github.com/spacetelescope/stdatamodels/pull/54 that I think will give us a little more flexibility in the future.  I also added parameterized the tests to cover models opened from ASDF files.

### DIFF
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -174,8 +174,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
 
         # Determine what kind of input we have (init) and execute the
         # proper code to intiailize the model
-        self._files_to_close = []
-        self._iscopy = False
+        self._file_references = []
         is_array = False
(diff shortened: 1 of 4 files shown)


src/stdatamodels/model_base.py
```diff
@@ -349,50 +350,27 @@ def __enter__(self):
...
-        def _drop_array(d):
-            # Walk tree and delete numpy arrays
-            if isinstance(d, dict):
-                for val in d.values():
-                    _drop_array(val)
-            elif isinstance(d, list):
-                for val in d:
-                    _drop_array(val)
-            elif isinstance(d, np.ndarray):
-                del d
```
eslavich, line 362: This method had no effect because `del d` just deletes the local variable, it doesn't modify the dict or list that contained the array.

jdavies-st (approved):
This is pretty slick.  I like.

PR #60, opened 2021-02-12
Handle multiple calls to DataModel.close gracefully

eslavich:
A bug revealed by tests in jwst...


jdavies-st (approved):
Ah, yes.  We might also want to clean up where the same shallow copy is getting closed multiple times in pipelines.

PR #64, opened 2021-03-05
Don't set level on library loggers

eslavich:
I can't think of why we'd want logger levels set to DEBUG in a library package, but let me know if there's something I'm overlooking.

https://jira.stsci.edu/browse/AL-473


nden (approved):

nden:
Needs a rebase.

PR #68, opened 2021-05-03
Fix CRDS env vars in tox

jdavies-st:
This fixes the problem seen here

https://github.com/spacetelescope/stdatamodels/runs/2493534318?check_suite_focus=true#step:5:15

where the `jwst` tests use CRDS via the new CRDS_CONTEXT reporter plugin, and therefore need the CRDS env vars set.

This also expands the testing to the whole downstream `jwst` package unit tests, not just those in `jwst.datamodels`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
             python-version: 3.8
             toxenv: py38
 
-          - name: jwst.datamodels tests
+          - name: jwst tests
             runs-on: ubuntu-latest
             python-version: 3.8
             toxenv: jwst
--- a/tox.ini
+++ b/tox.ini
@@ -3,9 +3,20 @@ envlist= py38,style,bandit
 
 [testenv]
 usedevelop= true
+
 extras= test
-# astropy will complain if the home directory is missing
-passenv= HOME
+
+passenv =
+    TOXENV
+    CI
+    CODECOV_*
+    HOME
+    CRDS_*
+    STRICT_VALIDATION
+    PASS_INVALID_VALUES
+    VALIDATE_ON_ASSIGNMENT
+    TEST_BIGDATA
+
 deps=
     # The tests require code that is in asdf
     # master but not yet released.  This can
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -58,10 +68,15 @@ commands=
     sphinx-build -W docs/source build/docs
 
 [testenv:jwst]
+setenv=
+    CRDS_SERVER_URL = https://jwst-crds.stsci.edu
+    CRDS_PATH = ~/crds_cache
```
eslavich, line 73: I tried running this on my laptop and it created a directory named `~` in the repo root.  Maybe we ought to specify a path relative to one of the tox directory variables?

tox.ini
```diff
@@ -3,9 +3,20 @@ envlist= py38,style,bandit
...
 extras= test
-# astropy will complain if the home directory is missing
-passenv= HOME
+
+passenv =
+    TOXENV
+    CI
+    CODECOV_*
+    HOME
+    CRDS_*
```
eslavich, line 14: Why do we need `CRDS_*` and the others that follow?  We're not passing any of them in from the actions workflow...

.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
             python-version: 3.8
             toxenv: py38
 
-          - name: jwst.datamodels tests
+          - name: jwst tests
```
eslavich, line 53: Is the benefit of running all the jwst tests worth the cost?  Without this the test suite takes about a minute.

tox.ini
jdavies-st: That's a good idea.  The other option would be to define these at the job level in the CI workflow and have them pass through to the jwst job.  I wonder if that's the better alternative?

.github/workflows/ci.yml
jdavies-st: Good question.  The reason we only ran `jwst.datamodels` before is that that submodule was the only one that could run with `--pyargs` when we set up this job eons ago.  Now the whole suite can run with `--pyargs`, so, perhaps better coverage is  better?  That said, I don't know if it actually is better coverage of the code here.  Hmm.

tox.ini
jdavies-st: But they do get defined in the `jwst.datamodels` tests, though maybe we don't need these defined as this only is needed if they are passed in from env vars outside the tox/pytest run itself?

jdavies-st:
This PR is no longer needed.  Closing.

PR #70, opened 2021-06-08
Make arrays C-contiguous on save

eslavich:
This is a fix for https://github.com/spacetelescope/jwst/issues/6104.

I think I understand the full story now --  asdf is trying to write the groupdq array into a block because of this change (released in asdf 2.7.4):

https://github.com/asdf-format/asdf/pull/949

Since RefPixStep is assigning groupdq to a view of itself:

https://github.com/spacetelescope/jwst/blob/1.2.2/jwst/refpix/irs2_subtract_reference.py#L251

it ends up with a base array with a wacky strides value that asdf doesn't know how to compute a view against, so it's forced to make a copy.  Before 2.7.4 asdf did not make a copy but that left the door open to other problems.

In the end I think the sanest thing we can do here is make the arrays C-contiguous before saving to an ASDF-in-FITS file, since that is how astropy.io.fits will write them anyway.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.2.2 (unreleased)
+==================
+
+- Make arrays contiguous on save to prevent issue with duplicate
+  array data between ASDF and FITS. [#70]
+
 0.2.1 (2021-03-08)
 ==================
 
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -352,6 +352,25 @@ def _save_from_schema(hdulist, tree, schema):
     validator.validate(tree, _schema=schema)
 
 
+def _normalize_arrays(tree):
+    """
+    Convert arrays in the tree to C-contiguous, since that is
+    how they are written to disk by astropy.io.fits and we
+    don't want the asdf library to notice the change in memory
+    layout and duplicate the array in the embedded ASDF.
+    """
+    def normalize_array(node, json_id):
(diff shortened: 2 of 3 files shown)


jdavies-st:
We should probably do a bugfix release of `stdatamodels` after this is merged.

jdavies-st:
Also, looking at all the pre-write munging we are doing here in `stdatamodels`, I agree that it makes a lot of sense to pull `asdf.AsdfInFits` out of the `asdf` package and into `stdatamodels`, given that `stdatamodels` is really the only consumer of this feature.  Right now it's hard to determine where the responsibilities of one package end and those of the other start.  If someone else wanted to use `asdf.AsdfInFits`, would they have to worry about this bug?

eslavich:
> If someone else wanted to use asdf.AsdfInFits, would they have to worry about this bug?

Yes, though it's less severe than it used to be -- at least now, asdf won't write out a corrupted view, instead it just adds it to an internal block if it can't figure out how the base array would be written.  It would be better I think if we forced users to be explicit about the arrays that they intend to be stored in FITS and referenced from ASDF.  Something like:

```python
from asdf.fits_embed import FitsArray

# ...

af["data"] = FitsArray(hdul["SCI"])
```

Trying to automatically infer the user's intention gets us into trouble all over the place in asdf.

We could look into solving the general problem of recomputing a view when its base array's memory layout changes.  I'm not sure that's even possible in all cases?  At least, it seems like a hard problem.

PR #71, opened 2021-06-10
Fix _no_asdf_extension bug

jdavies-st:
This fixes a bug seen when reading in old CRDS reffiles and trying to write them back out.

Resolves [AL-556](https://jira.stsci.edu/browse/AL-556)
Resolves spacetelescope/jwst#6028 / [JP-2080](https://jira.stsci.edu/browse/JP-2080)

 - Don't allow ASDF hdus to get passed through `extra_fits`, ever
 - Never write out any ASDF extension if `self._no_asdf_extension=True`

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -539,23 +539,25 @@ def _load_extra_fits(hdulist, known_keywords, known_datas, tree):
 
     # Add header keywords and data not in schema to extra_fits
     for hdu in hdulist:
-        known = known_keywords.get(hdu, set())
+        # Don't add ASDF hdus to extra_fits for any reason
+        if hdu.name != "ASDF":
+            known = known_keywords.get(hdu, set())
 
-        cards = []
-        for key, val, comment in hdu.header.cards:
-            if not (is_builtin_fits_keyword(key) or
-                    key in known):
-                cards.append([key, val, comment])
+            cards = []
+            for key, val, comment in hdu.header.cards:
+                if not (is_builtin_fits_keyword(key) or
+                        key in known):
+                    cards.append([key, val, comment])
 
-        if len(cards):
-            properties.put_value(
-                ['extra_fits', hdu.name, 'header'], cards, tree)
+            if len(cards):
+                properties.put_value(
+                    ['extra_fits', hdu.name, 'header'], cards, tree)
 
-        if hdu not in known_datas:
-            if hdu.name.lower() != 'asdf':
-                if hdu.data is not None:
-                    properties.put_value(
-                        ['extra_fits', hdu.name, 'data'], hdu.data, tree)
+            if hdu not in known_datas:
+                if hdu.name.lower() != 'asdf':
+                    if hdu.data is not None:
+                        properties.put_value(
+                            ['extra_fits', hdu.name, 'data'], hdu.data, tree)
 
 
 def _load_history(hdulist, tree):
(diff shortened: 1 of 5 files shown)


src/stdatamodels/fits_support.py
```diff
@@ -539,23 +539,25 @@ def _load_extra_fits(hdulist, known_keywords, known_datas, tree):
...
+                if not (is_builtin_fits_keyword(key) or
+                        key in known):
+                    cards.append([key, val, comment])
+
+            if len(cards):
+                properties.put_value(
+                    ['extra_fits', hdu.name, 'header'], cards, tree)
+
+            if hdu not in known_datas:
+                if hdu.name.lower() != 'asdf':
```
eslavich, line 557: This check can be dropped now that we're skipping this entire block of code for HDUs named "ASDF"

src/stdatamodels/model_base.py
```diff
@@ -624,6 +624,13 @@ def to_fits(self, init, *args, **kwargs):
             with warnings.catch_warnings():
                 warnings.filterwarnings('ignore', message='Card is too long')
                 if self._no_asdf_extension:
+                    # For some old files that were written out before the
+                    # _no_asdf_extension existed, these will have an ASDF
+                    # extension, which may get passed along through extra_fits.
+                    # Avoid this.
+                    if "ASDF" in [hdu.name for hdu in ff._hdulist]:
```
eslavich, line 631: The HDUList class implements `__contains__` so we can just do `if "ASDF" in ff._hdulist` here?

src/stdatamodels/model_base.py
```diff
@@ -624,6 +624,13 @@ def to_fits(self, init, *args, **kwargs):
...
+                        del ff._hdulist["ASDF"]
+                    # Use HDUList.writeto instead of AsdfInFits.write_to
                     ff._hdulist.writeto(init, *args, **kwargs)
```
eslavich, line 634: I wonder why we're even using AsdfInFits in this case.  I can't think of anything valuable it does since we're discarding the tree anyway.

jdavies-st: Do we keep around a reference to the HDUList in the in-memory datamodel?  I couldn't find it.  I think we're mostly using `AsdfInFIts` as structure to map the tree to FITS keywords on `save`, and then discarding the tree by using the `HDUList.writeto()` method.  So it does serve some purpose, though I don't like they we're poking around in private attributes.

src/stdatamodels/model_base.py
jdavies-st: Nice catch.  I will change this here and in the test.

src/stdatamodels/model_base.py
eslavich: AsdfInFits doesn't know how to map to FITS keywords though, that's an stdatamodels feature.  Maybe we ought to change the stdatamodels.fits_support.to_fits method to return an HDUList, and construct the AsdfInFits in this method if we need it?

jdavies-st: That would make more sense, wouldn't it?  Let me have a look.

jdavies-st: It would have to return both the tree and the hdulist, as the tree has been modified in `fits_support.to_fits` function.

jdavies-st: ...unless it is modifying the tree in-place.

jdavies-st: It doesn't modify the tree in-place, so the return signature will have to be a tuple of the hdulist and the new tree.

eslavich (approved):
LGTM!  I like the way to_fits turned out.

PR #72, opened 2021-06-15
Remove DataModel.my_attribute from public interface

### LINKED ISSUE #14: Remove my_attribute as a public-facing method
It is not clear why this exists

https://github.com/spacetelescope/stdatamodels/blob/8f9c64b812de18072a7e5ab40285e396bbeab846/src/stdatamodels/model_base.py#L631-L633

or why it is a public-facing method.  If it is used internally it should be named appropriately and made private.

jdavies-st:
Resolves #14

### DIFF
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -656,12 +656,8 @@ class DataModel(properties.ObjectNode):
                 self._shape = primary_array.shape
         return self._shape
 
-    def my_attribute(self, attr):
-        properties = frozenset(("shape", "history", "_extra_fits", "schema"))
-        return attr in properties
-
     def __setattr__(self, attr, value):
-        if self.my_attribute(attr):
+        if attr in frozenset(("shape", "history", "_extra_fits", "schema")):
             object.__setattr__(self, attr, value)
         else:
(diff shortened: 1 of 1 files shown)


jdavies-st:
Not sure what we should milestone this as.  Bugfix?  Minor?  We could also postpone this until we tear out all of the `NDData` interface.

src/stdatamodels/model_base.py
```diff
@@ -646,14 +646,10 @@ def shape(self):
...
-    def my_attribute(self, attr):
-        properties = frozenset(("shape", "history", "_extra_fits", "schema"))
-        return attr in properties
-
     def __setattr__(self, attr, value):
-        if self.my_attribute(attr):
+        if attr in frozenset(("shape", "history", "_extra_fits", "schema")):
             object.__setattr__(self, attr, value)
-        elif ndmodel.NDModel.my_attribute(self, attr):
+        elif attr in frozenset(("data", "mask", "unit", "wcs", "unceratainty")):
```
eslavich, line 652: While NDModel still exists it seems like it should own this list.  Maybe we just rename the method to `_my_attribute` for now?

eslavich (approved):

PR #75, opened 2021-06-23
Raise error when gentle_asarray is asked to make an unsafe cast

eslavich:


eslavich:
This has been addressed by #82 and #79

PR #79, opened 2021-08-20
Add options for strict validation of arrays

eslavich:
This PR adds two keyword arguments to the DataModel initializer -- `cast_fits_arrays`, which controls the cast to schema dtype that occurs by default when reading an array from a FITS file, and `validate_arrays`, which enables full validation of the schema datatype, ndim, and/or max_ndim.

These options will be used by CRDS to impose strict validation on arrays for submitted reference files.

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -449,7 +449,7 @@ def _fits_keyword_loader(hdulist, fits_keyword, schema, hdu_index, known_keyword
     return val
 
 
-def _fits_array_loader(hdulist, schema, hdu_index, known_datas):
+def _fits_array_loader(hdulist, schema, hdu_index, known_datas, context):
     hdu_name = _get_hdu_name(schema)
     _assert_non_primary_hdu(hdu_name)
     try:
@@ -458,7 +458,7 @@ def _fits_array_loader(hdulist, schema, hdu_index, known_datas):
         return None
 
     known_datas.add(hdu)
-    return from_fits_hdu(hdu, schema)
+    return from_fits_hdu(hdu, schema, context._cast_fits_arrays)
 
 
 def _schema_has_fits_hdu(schema):
@@ -507,7 +507,7 @@ def _load_from_schema(hdulist, schema, tree, context, skip_fits_update=False):
         elif 'fits_hdu' in schema and (
                 'max_ndim' in schema or 'ndim' in schema or 'datatype' in schema):
             result = _fits_array_loader(
-                hdulist, schema, ctx.get('hdu_index'), known_datas)
+                hdulist, schema, ctx.get('hdu_index'), known_datas, context)
 
             if result is None and context._validate_on_assignment:
                 validate.value_change(path, result, schema, context)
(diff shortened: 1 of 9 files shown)


src/stdatamodels/fits_support.py
```diff
@@ -449,7 +449,7 @@ def _fits_keyword_loader(hdulist, fits_keyword, schema, hdu_index, known_keyword
     return val
 
 
-def _fits_array_loader(hdulist, schema, hdu_index, known_datas):
+def _fits_array_loader(hdulist, schema, hdu_index, known_datas, context):
```
jdavies-st, line 452: Is there a reason all these functions need to pass around a context, as opposed to just being methods of the parent class that already have access to `self`?

src/stdatamodels/validate.py
```diff
@@ -35,6 +39,100 @@ def value_change(path, value, schema, ctx):
...
 
+def _validate_datatype(validator, schema_datatype, instance, schema):
+    """
+    This extends the ASDF datatype validator to support ndim
+    and max_ndim within individual fields of structured arrays,
+    and handle the absence of the shape field correctly.
+
+    Additionally, dtypes are required to be equivalent, instead
+    of just "safe" to cast.
+    """
```
jdavies-st, line 50: Is there a reason to have this extend the validator just in `stdatamodels` and not have it available in `asdf`?  Or do you see this as something that might be moved over to `asdf` at some point?

src/stdatamodels/validate.py
```diff
@@ -35,6 +39,100 @@ def value_change(path, value, schema, ctx):
...
+        elif 'data' in instance:
+            array = ndarray.inline_data_asarray(instance['data'])
+            instance_datatype, _ = ndarray.numpy_dtype_to_asdf_datatype(array.dtype)
+        else:
+            raise ValidationError("Not an array")
+    elif isinstance(instance, (np.ndarray, ndarray.NDArrayType)):
+        instance_datatype, _ = ndarray.numpy_dtype_to_asdf_datatype(instance.dtype)
+    else:
+        raise ValidationError("Not an array")
+
```
jdavies-st, line 66: Here you are doing `raise ValidationError()` but elsewhere below you are `yield ValidationError()`.  Any reason for this?  Shouldn't they all be `raise`?

jdavies-st (approved):
Looks good!

Is the reason to have this as an option to turn because:

- lots of existing reffiles don't meet these requirements and therefore cause failures in `jwst`?
- (or) `jwst` step code doesn't respect strict casting and tries to stuff arrays of incorrect `dtype` into datamodels with conflicting schemas?

src/stdatamodels/validate.py
eslavich: Oops, good question, these should be yields.  I believe the yield allows the validator to descend into other paths through the combiners and still succeed, while the raise stops it in its tracks.

src/stdatamodels/validate.py
eslavich: I think we'll want to shore up the asdf validator eventually, but I'm not prepared to do that now -- it's going to cause havoc since previous valid files will become invalid.

src/stdatamodels/fits_support.py
eslavich: That's a good idea, we're overloading the term "context" in this module (sometimes it's the AsdfFile object, sometimes the DataModel) and it would be nice if one of them was just `self`.

eslavich:
> Is the reason to have strict validation as an option to turn because:
> 
> - lots of existing reffiles don't meet these requirements and therefore cause failures in `jwst`?
> - (or) `jwst` step code doesn't respect strict casting and tries to stuff arrays of incorrect `dtype` into datamodels with conflicting schemas?

Both reasons, really.  The first one really hurts us, because the default validation behavior in jwst is to emit a warning and then leave the array set at zeros, which leads to all sorts of confusing failures down the line.  The step code also has plenty of cases where arrays are assigned that fail strict validation even after the casting.

src/stdatamodels/fits_support.py
eslavich: I added an issue here: https://github.com/spacetelescope/stdatamodels/issues/80

PR #81, opened 2021-08-23
Remove romancal.datamodels tests from test suite

### LINKED ISSUE #78: Remove romancal from the downstream CI tests
The `romancal` package will no longer depend on `stdatamodels`, so remove it from downstream testing in our Github actions CI workflow.

eslavich:
Resolves #78


jdavies-st (approved):
Thank you!  Was on the to-do list.

PR #82, opened 2021-08-25
Raise error when asked to cast FITS binary table with pseudo-unsigned column

eslavich:
This PR changes the datamodels casting to raise an error when asked to cast a FITS binary table with a pseudo-unsigned integer column.  Due to a bug in astropy `np.asanyarray` causes the TZEROn offset to be applied twice, which makes a mess of the unsigned integer data in the column.

The miri flat references in the current context all have this problem and will need to be fixed and re-submitted to prevent this exception from occurring.


eslavich:
We've fixed the references, so I'm going to merge this and run the regression tests on master.

PR #85, opened 2021-09-20
Add support for deprecated properties

eslavich:
This adds a new schema feature, deprecated_properties, that describes a list of properties that when accessed are redirected to a new property with a warning.  This is intended to be used for Multi* model deprecations.


eslavich:
Closing in favor of https://github.com/spacetelescope/stdatamodels/pull/86 due to GitHub not showing any diff here

PR #86, opened 2021-09-21
Add support for deprecated properties

eslavich:
This adds a new schema feature, deprecated_properties, that describes a list of properties that when accessed are redirected to a new property with a warning. This is intended to be used for Multi* model deprecations.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.4.0 (unreleased)
+==================
+
+- Add schema feature to forward deprecated model attributes to
+  a new location. [#85]
+
 0.3.0 (2021-09-03)
 ==================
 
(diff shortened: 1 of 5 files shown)


jdavies-st (approved):
This is clever and looks like it works well.

In the deprecation warning, if I use

```
origin = model.old_origin
```

it's going to return the value at `model.meta.origin` and tell me that "Attribute 'old_origin' has been deprecated".  That useful, but how does a user find out what the new attribute (i.e. `model.meta.origin`) is?  Maybe a more verbose message?

eslavich:
>  That useful, but how does a user find out what the new attribute (i.e. model.meta.origin) is? Maybe a more verbose message?

Hmm... good point.  I'm going to fiddle with this a bit.

PR #87, opened 2021-11-12
Support casting of FITS_rec with unsigned integer column

eslavich:
Previously we raised an exception when asked to cast a FITS_rec with an unsigned integer column, since that hits a bug in astropy that garbles the data in the column by applying the bzero shift twice.  This PR adds special handling for such arrays so that we no longer need to raise the exception.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 - Add schema feature to forward deprecated model attributes to
   a new location. [#86]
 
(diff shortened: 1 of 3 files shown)


eslavich:
Passing regtest run: https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/125/

kmacdonald-stsci (approved):
Looks good!

PR #89, opened 2022-02-01
JP-2227: Making changes to properly create HDU from an ndarray.

kmacdonald-stsci:
Resolves JP-2227
Closes issue: https://github.com/spacetelescope/jwst/issues/6271

To support lazy loading in JWST data models a wrapper is used to wrap Numpy `ndarrays` called NDArrayType.  At the lower levels in the class hierarchy this causes problems during write.  When a data model is written an HDU needs to be created for the HDU list to be written out.  This can only be done properly if the `ndarray.dtype` is known.

For lazy loading this may not be known because the needed information is not in the NDArrayType wrapper, so an extra step must be taken to check if the type is NDArrayType, then ensure the `ndarray` gets loaded.  When this is done all the needed information for type checking on write is there and the round trip will succeed.

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -156,6 +156,9 @@ def get_hdu(hdulist, hdu_name, index=None):
 
 
 def _make_hdu(hdulist, hdu_name, index=None, hdu_type=None, value=None):
+    if isinstance(value, NDArrayType):
+        value = np.asarray(value)
+
     if hdu_type is None:
         hdu_type = _get_hdu_type(hdu_name, value=value)
         if hdu_type is None:
@@ -164,8 +167,6 @@ def _make_hdu(hdulist, hdu_name, index=None, hdu_type=None, value=None):
     if hdu_type == fits.PrimaryHDU:
         hdu = hdu_type(value)
     else:
-        if isinstance(value, NDArrayType):
-            value = np.asarray(value)
         hdu = hdu_type(value, name=hdu_name)
     if index is not None:
         hdu.ver = index + 1
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -13,6 +13,7 @@ import asdf
 from asdf import fits_embed
 from asdf import resolver
 from asdf import schema as asdf_schema
+from asdf.tags.core import NDArrayType
 from asdf.tags.core import ndarray, HistoryEntry
 from asdf import treeutil
 from asdf.util import HashableDict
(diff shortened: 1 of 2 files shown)


nden:
Needs a change log

src/stdatamodels/fits_support.py
```diff
@@ -163,6 +164,8 @@ def _make_hdu(hdulist, hdu_name, index=None, hdu_type=None, value=None):
     if hdu_type == fits.PrimaryHDU:
         hdu = hdu_type(value)
     else:
+        if isinstance(value, NDArrayType):
```
eslavich, line 167: This check needs to be higher up in the chain of methods.  The `hdu = hdu_type(value)` line just above will also fail if passed a NDArrayType, and there may be problems when updating an existing HDU that the check here doesn't address.

kmacdonald-stsci: Since `hdu_type` is only called in `_make_hdu` I kept the conditional in that function, but moved it to the beginning of the function.

hbushouse (approved):
Seems OK to me, but apparently my approval (or that of @stscieisenhamer) doesn't count for this repo.

src/stdatamodels/fits_support.py
```diff
@@ -155,6 +156,9 @@ def get_hdu(hdulist, hdu_name, index=None):
 
 
 def _make_hdu(hdulist, hdu_name, index=None, hdu_type=None, value=None):
+    if isinstance(value, NDArrayType):
```
eslavich, line 159: @kmacdonald-stsci did you check the update case?  I mean, when the HDU doesn't need to be created and the value is assigned to an existing HDU?

hbushouse:
There's still one unanswered question from @eslavich and once that's resolved it needs approval from someone who's approved reviewer for this package (Jonathan and I are not, and hence our approvals don't count). @nden ?

src/stdatamodels/fits_support.py
kmacdonald-stsci: I checked this and it works fine.

eslavich (approved):
LGTM.  Ken and I chatted on Slack and I'm convinced that the update case doesn't have this problem.

hbushouse:
Can someone with write access to `stdatamodels` merge this?  @nden @eslavich

eslavich:
@hbushouse I also cursed you with admin privileges

hbushouse:
> @hbushouse I also cursed you with admin privileges

A pox on both our houses?

PR #90, opened 2022-03-04
Portion of fix for asdf-transform-schemas develop.

WilliamJamieson:
Currently, jwst is failing due to issues with the additions to `transform-1.2.0` in asdf-transform-schemas development. This is due to issues with the `schemas` module in this package.

The PR spacetelescope/jwst#6752 fixes the majority of the this problem by altering how `transform-1.2.0` is used in some of the jwst datamodels; however, this does not completely fix the problem because the schema module still attempts to remove items from the tree which don't exist. This PR fixes that remaining issues.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@
 ==================
 
 - Changed the way NDArrayType wrappers are handled on write. [#89]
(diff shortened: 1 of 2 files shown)


eslavich (approved):
LGTM.  I double checked and it looks like the other references to `properties` are already careful to check that it exists.

PR #92, opened 2022-03-14
Fix issues with FITS writing validators with jsonschema 4.x

eslavich:
jsonschema 4.x does not support storing custom attributes on the validator object.


WilliamJamieson (approved):
LGTM, this is more clever than my solution.

PR #93, opened 2022-03-14
Resolve deprecation warnings due to use of pytest.warns(None)

eslavich:
Just fixing some warnings that cropped up due to a pytest deprecation...


WilliamJamieson (approved):
LGTM

PR #97, opened 2022-06-06
use GitHub Actions to cache environments and use dependent CI workflow structure

zacharyburnett:
image

Notable changes:

- workflow jobs are defined in GitHub Actions to simplify the environment caching and construction
- all environments are cached to speed up subsequent runs; cache key is set to hash the build files so that if a build requirement changes the environment is built from scratch
- matrices are dependent in a workflow such that a commit that fails style checks will not run tests, and so on
- matrices test a more comprehensive set of platforms and Python versions


zacharyburnett:
fixes https://github.com/spacetelescope/jwst/issues/6887

nden:
Could you redefine the "Required" checks so it checks as passed?

nden (approved):

PR #102, opened 2022-07-28
Add ABC for ModelContainer inheritance

tapastro:

Closes #

This PR adds an abstract base class that will exist only as a marker, to check against and alter pipeline behavior for datamodel container input versus datamodel input. Currently, stpipe checks instances against `Sequence`, which has some pitfalls.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ Changes to API
 Other
 -----
 
+- Add abstract base class for datamodel container inheritance [#102]
+
 0.4.3 (2022-06-03)
 ==================
 
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -8,6 +8,7 @@ import os
 from pathlib import PurePath
 import sys
 import warnings
+import abc
 
 import numpy as np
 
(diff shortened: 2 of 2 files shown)


braingram:
@tapastro is this still needed?

tapastro:
I think it would still be a useful modification - stpipe still checks for Containers using isinstance(input, Sequence): https://github.com/search?q=repo%3Aspacetelescope%2Fstpipe%20Sequence&type=code

It could lead to issues if Sequence-derived models are introduced that are not Containers.

braingram:
Thanks! I'll start taking a look at it. If this is the improve interoperability with stpipe and containers. Will this impact romancal? It appears the model container class there inherits from `Sequence`:
https://github.com/spacetelescope/romancal/blob/c5864bf4dac9bb8f2e66aba90da98e3e8fdf2414/romancal/datamodels/container.py#L30
Since romancal doesn't include stdatamodels as a dependency perhaps we instead want a protocol (or abstract base class that uses `__subclasshook__`) in stpipe?

tapastro:
You went over my head with `__subclasshook__`, but I would wager that the ModelContainer in both datamodels could be improved by inheriting from a unique ABC for purposes of selecting container-like datamodels during stpipe'ing. I'll go read about your suggestion. 🙂

braingram:
Excellent! I did just find an example in stpipe `AbstractDataModel`:
https://github.com/spacetelescope/stpipe/blob/5369e85b35c99280cd578b05c50f40f3ccb70d7f/src/stpipe/datamodel.py#L4

PR #103, opened 2022-11-08
remove use of deprecated override__dir__

braingram:
astropy 5.2 will deprecate override__dir__
The deprecation notes suggest an alternative (used here) which should work for all versions of python >=3.3

Deprecation: https://github.com/astropy/astropy/blob/0944607116f373e64e94ce3e5535c49509897c87/astropy/utils/compat/misc.py#L19
Root python issue: https://bugs.python.org/issue12166
Fixed in 3.3: https://docs.python.org/3.3/whatsnew/changelog.html

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Bug Fixes
 ---------
 
 - Increase asdf version to >=2.14.1 to fix hdu data duplication [#105]
+- Remove use of deprecated override__dir__ [#103]
 
 Changes to API
 --------------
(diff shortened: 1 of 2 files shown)


jdavies-st (approved):
Perfect.  This code is actually tested by the `jwst.datamodels` unit tests, not the `stdatamodels` tests, hence coverage not showing it.  Would be good to add the `jwst.datamodels` tests to coverage.

PR #105, opened 2022-11-17
Add test for asdf data duplication issue with fits

braingram:
See asdf issue https://github.com/asdf-format/asdf/issues/1232

To summarize the issue, resaving an asdf in fits file breaks the link between hdu
items and arrays within the asdf tree. This results in inclusion of array data as
internal blocks in the asdf tree and a (sometimes substantial) increase in the size
of the tree. The test added here checks that the asdf tree stored in the 'ASDF' hdu
does not change size on a resave.

This test is currently skipped as it will fail with the current asdf version.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 Bug Fixes
 ---------
 
--
+- Increase asdf version to >=2.14.1 to fix hdu data duplication [#105]
 
 Changes to API
 --------------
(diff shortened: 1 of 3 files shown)


nden:
Closed and opened to restart the tests

nden (approved):

PR #107, opened 2022-12-13
Fix CI

jdavies-st:
This PR fixes the currently broken CI.

The main issue was outdated versions of the github `action/setup-python`.  `action/checkout` was also updated, which allowed the use of the pip wheel caching as well, removing the need to use a separate `action/cache` call to do so.

Other updates:

- after style and security checks, run all the other jobs in parallel.
- report coverage from running `jwst.datamodels` tests.  Coverage increases here by a few percent.  All good!
- change cache restore key for CRDS_CACHE for the `jwst.datamodels` job so that the cache is renewed monthly, given the episodic dev cycle for this package.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,11 @@ jobs:
         with:
           path: ${{ env.CRDS_PATH }}
           key: crds-${{ steps.cache-date.outputs.today }}
-      - run: pytest -n auto --pyargs jwst.datamodels
+      - run: pytest -n auto --pyargs jwst.datamodels --cov-report=xml --cov=src/stdatamodels
+      - run: coverage report -m
+      - uses: codecov/codecov-action@v2
+        with:
+          file: ./coverage.xml
   test_with_coverage:
     name: Coverage
     needs: [ style, audit ]
@@ -124,4 +128,3 @@ jobs:
           cache-dependency-path: setup.cfg
       - run: pip install -e ".[docs]"
       - run: sphinx-build -W docs/source build/docs
-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,28 +18,20 @@ jobs:
     name: Code style checks
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


braingram:
Thanks for making these fixes to the CI. They look good to me. I'm pinging @zacharyburnett to get an expert opinion :)

One question, should python 3.11 be added to the CI or is that best handled in a separate PR?

.github/workflows/ci.yml
```diff
@@ -18,28 +18,20 @@ jobs:
...
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
-          python-version: 3.8
-      - uses: actions/cache@v3
-        with:
-          path: ${{ env.pythonLocation }}
-          key: style-${{ runner.os }}-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/setup.*') }}
+          python-version: '3.10'
```
zacharyburnett, line 24:
```suggestion
          python-version: '3.10'
          cache: 'pip'
          cache-dependency-path: setup.cfg
```

.github/workflows/ci.yml
```diff
@@ -18,28 +18,20 @@ jobs:
...
-      - uses: actions/setup-python@v2
-        with:
-          python-version: 3.8
-      - uses: actions/cache@v3
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
-          path: ${{ env.pythonLocation }}
-          key: audit-${{ runner.os }}-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/setup.*') }}
+          python-version: '3.10'
```
zacharyburnett, line 34:
```suggestion
          python-version: '3.10'
          cache: 'pip'
          cache-dependency-path: setup.cfg
```

zacharyburnett (commented):
this looks great, you could probably also cache the style and audit installations as well (unless they all use the same cache key; that was unclear from the documentation)

zacharyburnett (approved):

PR #98, opened 2022-06-16
move build configuration into `pyproject.toml`

zacharyburnett:
use PEP621 and `ini2toml` / `peppyproject` to move metadata into `pyproject.toml`

PR #110, opened 2023-01-08
Reimplement support for ASDF-in-FITS in this package

eslavich:
This PR reimplements support for ASDF-in-FITS in this package.  ~~It requires https://github.com/asdf-format/asdf/pull/1289, so we'll need to get that merged and released before merging this.~~

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -366,21 +366,16 @@ FITS_SCHEMA_URL_MAPPING = resolver.Resolver(
 
 
 def _save_from_schema(hdulist, tree, schema):
-    def callback(node, json_id):
+    def datetime_callback(node, json_id):
         if isinstance(node, datetime.datetime):
             node = time.Time(node)
 
         if isinstance(node, time.Time):
             node = str(time.Time(node, format='iso'))
-        elif (isinstance(node, (np.ndarray, NDArrayType))):
-            base = get_array_base(node)
-            for hdu_index, hdu in enumerate(hdulist):
-                if hdu.data is not None and base is get_array_base(hdu.data):
-                    return _create_tagged_dict_for_fits_array(node, base, hdu, hdu_index)
 
         return node
 
-    tree = treeutil.walk_and_modify(tree, callback)
+    tree = treeutil.walk_and_modify(tree, datetime_callback)
 
     if _ASDF_GE_2_6:
         kwargs = {"_visit_repeat_nodes": True}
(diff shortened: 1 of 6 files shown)


src/stdatamodels/fits_support.py
```diff
@@ -644,10 +713,52 @@ def from_fits_asdf(hdulist,
...
+            ignore_unrecognized_tag=ignore_unrecognized_tag,
+        )
+
+    generic_file = generic_io.get_file(io.BytesIO(asdf_extension.data), mode="rw")
+    return asdf.open(
+        generic_file,
+        ignore_version_mismatch=ignore_version_mismatch,
+        ignore_unrecognized_tag=ignore_unrecognized_tag,
+        ignore_missing_extensions=ignore_missing_extensions,
+        tree_mapper=partial(_convert_fits_arrays, hdulist)
```
braingram, line 732: Is this the only change required to the ASDF API to enable this PR? Is there a way to reimplement ASDF-in-FITS without this? One idea would be to implement the conversion in a Converter (which has precedence over old style types). This would require adding block/block_manager support to Converters and would likely benefit from a 'accept/reject' system like mentioned here: https://github.com/asdf-format/asdf/issues/1263

braingram:
While looking at increasing the coverage I ran into what looks like a reintroduction of [https://github.com/asdf-format/asdf/pull/1234](https://github.com/asdf-format/asdf/issues/1232)

Adding a breakpoint and inspecting file_path here:
https://github.com/spacetelescope/stdatamodels/blob/c434220010ef94675d6b4e5db4a08428e5eba442/tests/test_fits.py#L504-L508

I see a file with
```yaml
arr:
- data: !core/ndarray-1.0.0
    source: 0
    datatype: float64
    byteorder: little
    shape: [5, 5]
- data: !core/ndarray-1.0.0
    source: 1
    datatype: float64
    byteorder: little
    shape: [5, 5]
meta: {model_type: DataModel}
...
```
Which differs from what is seen without this PR:
```yaml
arr:
- data: !core/ndarray-1.0.0
    source: fits:FOO,1
    datatype: float64
    byteorder: big
    shape: [5, 5]
- data: !core/ndarray-1.0.0
    source: fits:FOO,2
    datatype: float64
    byteorder: big
    shape: [5, 5]
meta: {model_type: DataModel}
...
```

Specifically, the hdus are not being linked to arrays resulting in duplicate data which would reintroduce [https://github.com/asdf-format/asdf/pull/1234](https://github.com/asdf-format/asdf/issues/1232)

braingram:
Thanks for the details. I had the wrong version of stdatamodels, with the code in this PR I am able to reproduce the issue you described.
```python
jsonschema.exceptions.ValidationError:      array([[1., 1., 1., ..., 1., 1., 1.],
[1., 1., 1., ..., 1., 1., 1.],
[1., 1., 1., ..., 1., 1., 1.],
...,
[1., 1., 1., ..., 1., 1., 1.],
[1., 1., 1., ..., 1., 1., 1.],
[1., 1., 1., ..., 1., 1., 1.]], dtype=float32) is not valid under any of the given schemas

     Failed validating 'anyOf' in schema['properties']['data']:
         {'$schema': 'http://stsci.edu/schemas/yaml-schema/draft-01',
...
```
Am I understanding this correctly that the issue is that _tagged_tree_transform is converting the tagged dict array object (with a source describing the fits hdu for the array data) into a ndarray (returning hdu.data). ASDF then attempts to validate what it expects is a tagged tree and fails when it encounters the ndarray?

braingram:
Thanks for the details.

I opened a PR against your branch with some changes that so far seem to pass all tests (including the jwst ones) and doesn't require the addition of _tagged_tree_transform to asdf.
https://github.com/eslavich/stdatamodels/pull/1/files

It does require reaching into the internal `NDArrayType._source` attribute. Also, I believe that opening without lazy load would result in an exception as the attempt to resolve the source (e.g. fits:HDU) would attempt to open an external file (as this appears like a url). However this option doesn't appear to be overridden anywhere in stdatamodels.

@eslavich what are your thoughts on the approach in the PR?

braingram:
I had to restart the regression tests.
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/2185/
The new run is using my branch: https://github.com/braingram/stdatamodels/tree/feature/alternate_asdf_in_fits which has the same changes as the source branch for this PR but also has updated tags (to allow the version to exceed the requirement in jwst to prevent the regtests from installing stdatamodels from pypi).

braingram:
Regtests 1285 showed 12 failures which are the same as the failures in 1284 (run against the pypi version of stdatamodels) and same as failures in other recent test runs.

So this PR looks good to merge to me.

@WilliamJamieson how do you want to time the merge of this PR with #98 and #112?

WilliamJamieson:
I think should wait #98 until this one is merged too.

That way it can just migrate all the changes to the pyproject.toml at once.

nden (approved):
If the comment on the dependency of these changes on ASDF is not valid any more it should be removed to not confuse later reviewers.

braingram:
> If the comment on the dependency of these changes on ASDF is not valid any more it should be removed to not confuse later reviewers.

Thanks! I put a strike through that part of the conversation. Let me know if any other clean up would be helpful.

PR #112, opened 2023-01-18
JP-3076: Move `jwst.datamodels` to `stdatamodels.jwst.datamodels`

WilliamJamieson:
This is a copy at the f9618fcb0c52b7ef768e0e2edcb6dd52f46235d0 of `jwst`

This PR is part of moving `jwst.datamodels` out of `jwst` and into `stdatamodels`. This is so that `jwst` pipeline results can be opened without having to install the entire `jwst` pipeline.

Note that this version of the datamodels moved are those as of commit [f9618fcb0c52b7ef768e0e2edcb6dd52f46235d0](https://github.com/spacetelescope/jwst/commit/f9618fcb0c52b7ef768e0e2edcb6dd52f46235d0) of `jwst`.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)


WilliamJamieson:
@zacharyburnett and @hbushouse I am consistently seeing in all my runs these 5 errors (including in all the CI here):
```python
FAILED src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py::test_crds_selectors_vs_datamodel[fgs] - FileNotFoundError: [Errno 2] No such file or directory: '/Users/wjamieson/crds_cache/mappings/jwst/jwst_1039.pmap'
FAILED src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py::test_crds_selectors_vs_datamodel[miri] - FileNotFoundError: [Errno 2] No such file or directory: '/Users/wjamieson/crds_cache/mappings/jwst/jwst_1039.pmap'
FAILED src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py::test_crds_selectors_vs_datamodel[nircam] - FileNotFoundError: [Errno 2] No such file or directory: '/Users/wjamieson/crds_cache/mappings/jwst/jwst_1039.pmap'
FAILED src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py::test_crds_selectors_vs_datamodel[niriss] - FileNotFoundError: [Errno 2] No such file or directory: '/Users/wjamieson/crds_cache/mappings/jwst/jwst_1039.pmap'
FAILED src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py::test_crds_selectors_vs_datamodel[nirspec] - FileNotFoundError: [Errno 2] No such file or directory: '/Users/wjamieson/crds_cache/mappings/jwst/jwst_1039.pmap'
```

This includes when I run all of these tests as they exist in JWST itself. Please let me know if this is expected or how to fix this.

WilliamJamieson:
@stscieisenhamer for some reason I cannot tag you for review

stscieisenhamer:
I'm not priveleged enough to review such things officially. But will do anyways 😄

WilliamJamieson:
Note that I don't think its a good idea to `squash and merge` all of this PR. I don't mind squashing parts of it, but I want to leave the copy of code (and docs) from jwst to stdatamodels as distinct commits in the history so that one can jump exactly between the two projects as the there is a lot of history of these models in `jwst`.

zacharyburnett:
It looks like the only issue remaining is the Codecov upload

WilliamJamieson:
> It looks like the only issue remaining is the Codecov upload

There is not much that can be done about the coverage deficit. This PR exactly reproduces the tests from jwst that are scoped purely for the datamodels. There is good "coverage" for the datamodels via there use in each of the pipeline steps, but those tests are not directly testing the code here itself. Whether or not this is real "coverage" of the datamodels themselves is debatable. 

I feel writing tests arbitrarily to "fix coverage deficits" is outside the scope of these changes as this PR is intended as purely a transfer of the datamodels from JWST to stdatamodels.

WilliamJamieson:
@zacharyburnett it looks like it was an intermittent issue (Codecov timed out on the upload), because it seems to have worked when I re-ran the tests.

For the coverage issue It looks like we run the `jwst` tests and upload the coverage of `stdatamodels` by these tests too. Once spacetelescope/jwst#7439 is merged, `jwst` will actively use these models. Right now the `jwst` used in these tests does not have its datamodels removed so it will not actually be using the code as it has been moved here.

I think I can temporarily "fix" this (for coverage purposes) by pointing the toxenv for that test at the branch for spacetelescope/jwst#7439. I can then revert this change after that PR has been merged. Thoughts?

zacharyburnett:
> I think I can temporarily "fix" this (for coverage purposes) by pointing the toxenv for that test at the branch for [spacetelescope/jwst#7439](https://github.com/spacetelescope/jwst/pull/7439). I can then revert this change after that PR has been merged. Thoughts?

I would apply the temp patch to this PR, see that it works, then revert it within this branch and squash the two commits together (cancelling it out) before merging

WilliamJamieson:
I think the coverage configuration is not correct, it seems to be picking up files outside these changes for coverage. Since that is outside the scope of these changes I suggest we revisit that at a later date.

nden (approved):

setup.cfg
```diff
@@ -27,6 +27,7 @@ install_requires =
     psutil>=5.7.2
     numpy>=1.16
     astropy>=5.0.4
+    stcal>=1.3.2,<1.4
```
nden, line 30: Hm. I didn't realize stcal becomes a dependence. It works for now and if necessary it's easy to break this dependence.

WilliamJamieson: Looks like there is nothing stopping me from moving the the two modules from `stcal` to here (they are not used by Roman). I have looked closely at the best way to do this.

So that I don't have to rework this PR and #113 again, I think the best bet is to "temporally" accept this dependency as part of this PR (and #113), then come back and move the few bits out of `stcal` we need as a new PR.

The main reason I want to do it this way is I want to maintain continuity of the changes from `jwst` to here. Namely, I include the hash of the exact commit I copied from `jwst` to here so that when following the history one can switch repos as they follow it backwards. I would rather not have to fiddle with the history in this PR to handle moving the `stcal` bits first, then updating jwst (which will require adjustments to the datamodels), and then use the new starting point for where I copy the datamodels from.

hbushouse (approved):
Looks OK to me.  This kind of major restructuring is going to require extensive testing, not just by regression and unit tests, but also real people banging away on the models interactively, via notebooks, etc., to make sure it all still works properly.

PR #116, opened 2023-02-06
consolidate test and check jobs

zacharyburnett:
This PR consolidates and simplifies the CI workflow.

image

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,9 @@ on:
 
 env:
   CRDS_SERVER_URL: https://jwst-crds.stsci.edu
-
+  CRDS_PATH: /tmp/crds_cache
+  CRDS_CLIENT_RETRY_COUNT: 3
+  CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
 
 jobs:
   test:
@@ -56,7 +58,6 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
           cache-dependency-path: 'pyproject.toml'
-      - run: echo "CRDS_PATH=$HOME/crds_cache" >> $GITHUB_ENV
       - if: contains(matrix.toxenv, 'test')
         run: >
           echo "pmap=$(
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,5 +58,5 @@ jobs:
         - macos: test-xdist
           python-version: 3.11
         - linux: test-xdist-cov
+          coverage: codecov
         - linux: test-jwst-xdist-cov
-      coverage: 'codecov'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,13 @@ jobs:
           CRDS_PATH: /tmp/crds_cache
           CRDS_CLIENT_RETRY_COUNT: 3
           CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
+  check:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
+    with:
+      envs: |
+        - linux: check-style
+        - linux: check-security
+        - linux: build-dist
   test:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     needs: [ crds_context ]
@@ -43,9 +50,6 @@ jobs:
         CRDS_PATH: /tmp/crds_cache
         CRDS_SERVER_URL: https://jwst-crds.stsci.edu
       envs: |
-        - linux: check-style
-        - linux: check-security
-        - linux: build-dist
         - linux: test-xdist
           python-version: 3.9
         - macos: test-xdist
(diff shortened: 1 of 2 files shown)


zacharyburnett:
could we also use OpenAstronomy workflows here?

WilliamJamieson:
This can be done with pre-commit.ci (the style and bandit checks) and as a fairly simple open astronomy workflow (this can do everything if we don't want to use the pre-commit.ci bot)

braingram:
> Can we set up the open astronomy workflow to cache CRDS?

The open astronomy reusable workflow supports caching.
https://github.com/OpenAstronomy/github-actions-workflows#cache-key
When attempting to get this to work with asdf I wasn't able to sort out how to generate a suitable key (such as the crds pmap name) due to the inability to add steps to the reusable workflow. One possibility would be to have a job run that gets the pmap name and passes this to the openastronomy workflow but I never tested this.

WilliamJamieson:
Tagging @ConorMacBride and @Cadair to comment on OpenAstronomy workflows.

zacharyburnett:
alright, I've set up a output job and successfully sent the `pmap` to the inputs (https://github.com/spacetelescope/stdatamodels/actions/runs/4177718609/jobs/7235639961#step:1:66); however, I'm not sure how to set environment variables within the job; specifically, I need to set `CRDS_PATH` but it does not pass through from the global `env:` definition

ConorMacBride:
Hi @zacharyburnett, due to a [limitation in GitHub Actions](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations) env vars cannot be passed through to a reusable workflow. However, I've actually just added a workaround for the OpenAstronomy publishing (but not tox) workflow! https://github.com/OpenAstronomy/github-actions-workflows/pull/56

I'll have a look at adding this workaround to the tox workflow. I think it should possible, just slightly more complex than for the publishing workflow. Will let you know when it's available.

zacharyburnett:
> Hi @zacharyburnett, due to a [limitation in GitHub Actions](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations) env vars cannot be passed through to a reusable workflow. However, I've actually just added a workaround for the OpenAstronomy publishing (but not tox) workflow! [OpenAstronomy/github-actions-workflows#56](https://github.com/OpenAstronomy/github-actions-workflows/pull/56)
> 
> I'll have a look at adding this workaround to the tox workflow. I think it should possible, just slightly more complex than for the publishing workflow. Will let you know when it's available.

ok, thank you!

.github/workflows/ci.yml
```diff
@@ -13,72 +13,49 @@ on:
...
-        uses: codecov/codecov-action@v3
-        with:
-          file: ./coverage.xml
-          flags: unit
-          fail_ci_if_error: true
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
+    needs: [ crds_context ]
+    with:
+      cache-path: /tmp/crds_cache
+      cache-key: crds-${{ needs.crds_context.outputs.pmap }}
```
ConorMacBride, line 42: I've opened a PR (https://github.com/OpenAstronomy/github-actions-workflows/pull/107) to add support for env vars being passed through. I want to get more feedback on the new API before merging, but you can try it out on my branch if you like:
```suggestion
    uses: ConorMacBride/github-actions-workflows/.github/workflows/tox.yml@set-env-tox
    needs: [ crds_context ]
    with:
      cache-path: /tmp/crds_cache
      cache-key: crds-${{ needs.crds_context.outputs.pmap }}
      setenv: |
        CRDS_PATH: /tmp/crds_cache
```

.github/workflows/ci.yml
```diff
@@ -13,72 +13,51 @@ on:
...
+      setenv: |
+        CRDS_PATH: /tmp/crds_cache
+      envs: |
+        - linux: check-style
+        - linux: check-security
+        - linux: build-dist
+        - linux: test-xdist
+          python-version: 3.9
```
braingram, line 50: Overall looks much cleaner.

Is this dropping testing for 3.8?

zacharyburnett: I'd say yes, it's time to drop `3.8`, but we can keep it in if everyone wants

braingram: Is the motivation from numpy dropping support in April? I don't have any objection, just hoping to better understand.

Also, should we bump the minimum version in the pyproject.toml to keep that in sync with the testing?

zacharyburnett:
looks like https://github.com/OpenAstronomy/github-actions-workflows/pull/107 was approved, so that should make this ready for review

hbushouse (approved):

braingram:
This PR dropped testing for python 3.8 (which JWST still supports). Was this intended?
https://github.com/spacetelescope/stdatamodels/blob/cee65666f7525df68e52a6cb50a358ed804e8470/.github/workflows/ci.yml#L52-L66
If so, we should increase the required python version to 3.9 as discussed https://github.com/spacetelescope/stdatamodels/pull/143

PR #117, opened 2023-02-06
fixes to run tests outside of CI

braingram:
~~#112 introduced a few tests to access CRDS. These expect a CRDS_PATH environment variable and can take considerable time to download data.~~

~~This PR modifies the log line accessing CRDS_PATH (to fall back to a value of None if CRDS_PATH is not defined) and adds a pytest mark (off by default) that controls running of the crds tests. The tox tests are modified to enable this new crds mark.~~

~~Edited: This PR adds a no_crds pytest option to disable tests that require downloading 500MB of files and~~ addresses a few failures related to local testing that resulted in test failures when:
- not run on the vpn (or on the vpn but without a defined CRDS_PATH environment variable)
- not run on a machine with a pre-existing cached CRDS context

~~A description of the mark is added to the readme.~~

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,16 @@
 [![codecov](https://codecov.io/gh/spacetelescope/stdatamodels/branch/master/graph/badge.svg?token=TrmUKaTP2t)](https://codecov.io/gh/spacetelescope/stdatamodels)
 
 
-Provides `DataModel`, which is the base class for data models implemented in the JWST and Roman calibration software.
+Provides `DataModel`, which is the base class for data models implemented in the JWST calibration software.
+
+
+## Unit Tests
+
+A few unit tests require downloading (~500MB) data from CRDS. CRDS must be configured for these tests to pass
+(see the [CRDS User Guide](https://jwst-crds.stsci.edu/static/users_guide/index.html)
+for more information). Minimally (if not on the stsci vpn where the default path of
+`/grp/crds/cache` is available) you will need to set `CRDS_PATH`.
+
+```bash
+export CRDS_PATH=/tmp/crds_cache/jwst_ops
+```
--- a/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py
+++ b/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py
@@ -6,6 +6,15 @@ import pytest
 import warnings
 from ..util import NoTypeWarning
 
+os.environ["CRDS_SERVER_URL"] = "https://jwst-crds.stsci.edu"
+
+# the CRDS_SERVER_URL environment variable must be set before
+# crds is imported
+import crds  # noqa: E402
+from crds.client.api import cache_references, dump_files  # noqa: E402
+from crds.core.exceptions import IrrelevantReferenceTypeError  # noqa: E402
+
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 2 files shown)


WilliamJamieson (changes requested):
I don't think this is the correct approach. The CRDS related tests were migrated directly from JWST with no changes. As such they are fundamentally JWST tests and decisions where made there to have them always run.

WilliamJamieson:
This appears likely to be in response to the ASDF downstream failures on stdatamodels.

This should be corrected in ASDF by enabling CRDS in the test suite like we do for JWST.

braingram:
I switched the name of the mark to no_crds. By default it will attempt to run the crds tests.

These require downloading ~500MB of data and additional environment setup to allow the tests to pass. I added some documentation about this setup to the README based on the wording in jwst.

The tests consistently failed locally if I was not connected to the vpn (where `/grp/crds/cache` already contains the necessary data). The failures were attributable to:
- no CRDS_PATH variable (I changed the log line in the test to not assume this variable is set)
- need for a manual crds sync (the test now syncs with `dump_files`)
- an issue with locally importing crds in the test (I moved this import outside the test for now with a comment on why it's no longer local, but I think there is a bug in crds in handling StrConfigItem). This could also be solved by removing the `jail_environ` fixture from these tests but it is unclear to me if that is needed for other reasons.

Having the option to disable these slow and data hungry tests in an otherwise fast test suite would make local testing and development easier. This option would also be useful for non-ASDF packages that want to add stdatamodels tests to their ci (like jdaviz, specutils, or other packages that want to support AsdfInFits). @WilliamJamieson if you feel strongly otherwise I will remove the pytest mark from this PR and edit it to include only the fixes for the local failures and the documentation.

WilliamJamieson:
Just like jwst the assumption is going to have to be that crds is setup and configured locally. This means the CRDS environment variables will have to be set correctly. The reason why this isn't working without the VPN is because CRDS falls back on assuming you are on the VPN and looking for the remote mounts that enables on your file system.

As far as the test markers themselves are concerned, those tests should not be disabled by default. They were part of the standard test suite when the code was in jwst for good reason, as development of the datamodels is closely tied to CRDS references. Specific discussions about whether or not these CRDS tests are strictly necessary is really up to the JWST developers.

@hbushouse what do you think should be done?

braingram:
> 



> Just like jwst the assumption is going to have to be that crds is setup and configured locally.

Is this documented already in stdatamodels? If not, do the additions in this PR explain this new assumption?

> This means the CRDS environment variables will have to be set correctly. The reason why this isn't working without the VPN is because CRDS falls back on assuming you are on the VPN and looking for the remote mounts that enables on your file system.

When not on the VPN and with correct CRDS environment variables the reference files fail to download for all of the tests. 

> 
> As far as the test markers themselves are concerned, those tests should not be disabled by default. 

They are not disabled by default.

WilliamJamieson:
> This option would also be useful for non-ASDF packages that want to add stdatamodels tests to their ci (like jdaviz, specutils, or other packages that want to support AsdfInFits).

I'm very confused about why any of these packages would want to run the stdatamodels tests as part of their CI. They depend on stdatamodels, but stdatamodels does not depend on them. For practical reasons those packages should be assuming that stdatamodels is functional (or at the very least passing its own tests) when running their CI.

We may want to consider running the jdaviz or specutils tests as part of the stdatamodels CI simply so we know that stdatamodels changes do not break those packages.

braingram:
> I'm very confused about why any of these packages would want to run the stdatamodels tests as part of their CI. They depend on stdatamodels, but stdatamodels does not depend on them. For practical reasons those packages should be assuming that stdatamodels is functional (or at the very least passing its own tests) when running their CI.

Yeah that was my mistake. As you rightly pointed out it's unlikely they will add stdatamodels tests to their CI.

Would this PR be acceptable if I separate the mark into a different PR so that this one only contains the fixes and documentation for the tests?

I am also not opposed to these tests being moved back to jwst.

WilliamJamieson:
> Would this PR be acceptable if I separate the mark into a different PR so that this one only contains the fixes and documentation for the tests?

I don't think the `mark` should be part of these changes. That is a completely separate question from resolving having the tests run in a way that is repeatable.

braingram:
The mark is now in a separate PR: https://github.com/spacetelescope/stdatamodels/pull/118

@WilliamJamieson are there other changes you're requesting for this PR?

WilliamJamieson (approved):

PR #118, opened 2023-02-09
add pytest no-crds option to disable tests that require crds

braingram:
This PR adds a no-crds pytest option to disable tests that require downloading ~500MB of files from CRDS.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
The option is off by default so that when `pytest` is run with no options, tests against crds will be run (and data will be downloaded when needed). The crds tests are only skipped if a user explicitly passes the `--no-crds` option.

This is useful because the test suite without the crds tests (on my machine) take ~48 seconds. With downloading data the crds tests alone take 218 seconds (490 seconds if using the vpn shared cache).

The crds tests contribute 0% to the test coverage.

Including this `no-crds` option will allow for much faster local testing.

nden (approved):

braingram:
Thanks!

PR #114, opened 2023-02-01
Add helper functions to aid in migration of AsdfInFits from ASDF

braingram:
This PR adds an asdf_in_fits submodule with 2 functions:
- open
- write

These are intended as near replacements for:
- opening an AsdfInFits file
- writing an AsdfInFits file

The signatures of the replacement functions were determined by surveying AsdfInFits usage outside of stdatamodels/jwst. These uses include:
- [hcipy](https://github.com/ehpor/hcipy)
  - read: call to [`asdf.fits_embed.AsdfInFits.open(filename)`](https://github.com/ehpor/hcipy/blob/058c473b31ae094f44ad1a3e445f6773c18f0d8a/hcipy/util/io.py#L140)
  - write: call to [`ff = asdf.fits_embed.AsdfInFits(hdulist, tree); ff.write_to(filename, …)`](https://github.com/ehpor/hcipy/blob/058c473b31ae094f44ad1a3e445f6773c18f0d8a/hcipy/util/io.py#L187-L188)
- [specutils](https://github.com/astropy/specutils)
  - read: call to [`with asdf.open(filename) as af`](https://github.com/astropy/specutils/blob/fc04cd6c948698c70ec9bd15b3354b58e4dbbcc2/specutils/io/default_loaders/jwst_reader.py#L438-L452)
  - read: call to [`with asdf.open(filename) as af`](https://github.com/astropy/specutils/blob/fc04cd6c948698c70ec9bd15b3354b58e4dbbcc2/specutils/io/default_loaders/jwst_reader.py#L558-L559)
  - write: call to [`ff = asdf.fits_embed.AsdfInFits(hdulist, tree); ff.write_to(filename, …)`](https://github.com/astropy/specutils/blob/fc04cd6c948698c70ec9bd15b3354b58e4dbbcc2/specutils/io/default_loaders/tests/test_jwst_reader.py#L409)
- [jdaviz](https://github.com/spacetelescope/jdaviz/)
  - read: call to [`with AsdfInFits.open as af`](https://github.com/spacetelescope/jdaviz/blob/20b224e616f61f0fd4245807560d7025aefc6867/jdaviz/configs/imviz/plugins/parsers.py#L232-L234)
[...]

PR #119, opened 2023-02-15
re-add id to crds-context step

braingram:
re-add the crds-context id that was removed in https://github.com/spacetelescope/stdatamodels/commit/7690332bed152b16e36076afcf5bbb04362bafa8

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
Tested after re-adding `crds-context` to get a correct cache key and the same errors occurred as those in https://github.com/spacetelescope/stdatamodels/pull/114

Removing the crds cache altogether also results in the same erors:
https://github.com/spacetelescope/stdatamodels/actions/runs/4187846272/jobs/7258231109#step:8:1098

PR #122, opened 2023-02-16
fix sphinx link for asdf in fits documentation

braingram:
The missing colon made it not appear as a std:label and made creating a sphinx link difficult in https://github.com/asdf-format/asdf/pull/1337

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
There is no urgency to this change (I think I was able to work around it in the asdf pr).

PR #124, opened 2023-02-17
JP-2113: Update NIRCam WFSS transforms for field dependence (v6)

tapastro:
Resolves [JP-2113](https://jira.stsci.edu/browse/JP-2113)

Closes #

This PR provides higher-order polynomial transform models for the NIRCam WFSS grisms. It has a coupled PR in jwst (https://github.com/spacetelescope/jwst/pull/7018).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_nircam_grism.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_nircam_grism.schema.yaml
@@ -15,37 +15,49 @@ allOf:
           Nircam Grism wavelength dispersion model, instance of astropy.modeling.Model
         type: array
         items:
-          type: object
+          type: array
+          items:
+            type: object
     dispx:
         description: |
           Nircam Grism row dispersion model, instance of astropy.modeling.Model
         type: array
         items:
-          type: object
+          type: array
+          items:
+            type: object
     dispy:
         description: |
           Nircam Grism column dispersion model, instance of astropy.modeling.Model
         type: array
         items:
-          type: object
+          type: array
+          items:
+            type: object
     invdispl:
         description: |
           Nircam Grism inverse wavelength dispersion model, instance of astropy.modeling.Model
         type: array
         items:
-          type: object
+          type: array
+          items:
+            type: object
     invdispx:
         description: |
           Nircam Grism inverse row dispersion model, instance of astropy.modeling.Model
         type: array
         items:
-          type: object
+          type: array
+          items:
+            type: object
     invdispy:
         description: |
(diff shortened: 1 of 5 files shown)


src/stdatamodels/jwst/datamodels/schemas/specwcs_nircam_grism.schema.yaml
```diff
@@ -15,37 +15,49 @@ allOf:
           Nircam Grism wavelength dispersion model, instance of astropy.modeling.Model
         type: array
         items:
-          type: object
+          type: array
+          items:
+            type: object
```
WilliamJamieson, line 20: This will likely be incompatible with existing files as they will contain a single instance of the object rather than an array containing a single object. This may cause a validation error when reading such files.

If there is a validation error with existing files then
```suggestion
          oneOf:
            - type: object
            - type: array
               items:
                 type: object
```
Should resolve that error. This will apply to all the similar changes in this file.

src/stdatamodels/jwst/transforms/converters/jwst_models.py
```diff
@@ -20,19 +20,22 @@ class NIRCAMGrismDispersionConverter(TransformConverterBase):
     def from_yaml_tree_transform(self, node, tag, ctx):
         from stdatamodels.jwst.transforms import models
 
-        _fname = getattr(models, node["class_name"])
+        _fname = getattr(models, node["model_type"])
```
WilliamJamieson, line 23: Additional changes maybe necessary if my suggested changes to the schema are needed. Namely, it may need to branch on single instances vs lists of instances.

WilliamJamieson (commented):
I noticed a potential issue with the proposed schema changes.

Basically one needs to check that currently written jwst files can be opened without issue with the changes proposed in this PR. If there are issues, then my suggestions below may help fix those problems.

src/stdatamodels/jwst/datamodels/schemas/specwcs_nircam_grism.schema.yaml
tapastro: When written (a few months ago), there was no intent to preserve backwards compatibility as the prior versions were barely "science-valid". I don't know if this stance has changed, but other parts of the code are also implicitly backwards incompatible and would break if used with older references. Will be something to discuss once this is closer to review.

nden (approved):
This has been in testing for quite some time now. I suggest we merge it so we can work out any issues in jwst.

pllim:
I think this broke Jdaviz. See #159

PR #125, opened 2023-02-24
Prevent deprecation issue from importing from tests

WilliamJamieson:

Closes #

Importing from another package's tests is strictly frowned upon as the tests are considered part of the private interface of the package. `asdf-astropy` is moving some stuff in its tests in astropy/asdf-astropy#170, which `stdatamodels` is importing in its tests. To be nice `asdf-astropy` is including a deprecation warning, which still causes one of the `stdatamodels` tests to fail see: https://github.com/astropy/asdf-astropy/actions/runs/4265199400/jobs/7424250651. This PR fixes that issue before it can arise.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/transforms/converters/tests/test_models.py
+++ b/src/stdatamodels/jwst/transforms/converters/tests/test_models.py
@@ -4,8 +4,13 @@ import numpy as np
 import os
 import warnings
 from astropy.modeling.models import Shift, Rotation2D, Const1D
-from asdf_astropy.converters.transform.tests.test_transform import (
-    assert_model_roundtrip)
+
+from astropy.utils import minversion
+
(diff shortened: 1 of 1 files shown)


zacharyburnett (approved):
LGTM, looks more resilient than before. Is there a need to abstract the `assert_model_roundtrip` to be used by more model packages?

WilliamJamieson:
> LGTM, looks more resilient than before. Is there a need to pull out the `assert_model_roundtrip` or similar functions into a common package to be used by more model packages?

Why? these are only needed by packages which depend on `asdf-astropy`. astropy/asdf-astropy#170 makes sure all these helper functions are in a public module.

PR #132, opened 2023-03-01
JP-3075: Using dynamic_mask function inversion to properly save ReferenceFileModels

kmacdonald-stsci:
… if present.

Resolves [JP-3075](https://jira.stsci.edu/browse/JP-3075)

This PR addresses round tripping issues for datamodels that use `dq_def`.  The `dq_def` table details how DQ flags are compressed for saving and how to be uncompressed when opened.  When opened, datamodels automatically call `dynamic_mask` to properly uncompress DQ arrays.   However, the inverse was not true during save, which meant that if these datamodels got saved, the uncompressed DQ array got saved to disk, so that when opened again the decompression failed.

An `inv` parameter got added to the `dynamic_mask` function in STCAL to compress uncompressed DQ arrays.  This inversion is used for datamodels that use `dq_def`.  In particular, in `stdatamodels` the inversion is used for ReferenceFileModel during save to ensure the uncompressed DQ array does not get saved, but is properly transformed using the `dq_def` table before saving.

The ReferenceFileModel uses the `dq_def` attribute.  A `save` method got created to call the `dynamic_mask` function with inversion to properly save DQ arrays.  A test was added to ensure proper roundtripping of DQ arrays. 

The STCAL PR https://github.com/spacetelescope/stcal/pull/145 needs to merged before this PR, since this PR depends on the `dynamic_mask` function update which is defined in STCAL.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


src/stdatamodels/jwst/datamodels/tests/test_models.py
```diff
@@ -373,3 +374,36 @@ def test_ramp_model_zero_frame_by_dimensions():
...
+
+    fname = os.path.join(ROOT_DIR, bname)
+    new_fname = os.path.join(tmp_path, nbname)
+
+    diff = None
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        # Open MaskModel, making sure it has non None dq and dq_def, then save it
+        with MaskModel(fname) as mask:
+            mask.save(new_fname)
```
WilliamJamieson, line 397: Is this supposed to fail if it as a `None` `dq` or `dq_def`? or is this just supposed to prevent the warning from appearing?

WilliamJamieson (changes requested):
One small question, otherwise the changes look good.

Also, please fix the merge conflict I would like to see the CI run "successfully".

src/stdatamodels/jwst/datamodels/tests/test_models.py
kmacdonald-stsci: This is just to prevent the warnings from appearing.

kmacdonald-stsci:
> One small question, otherwise the changes look good.
> 
> Also, please fix the merge conflict I would like to see the CI run "successfully".

Conflict resolved.

WilliamJamieson:
Looks like the new test is failing.

kmacdonald-stsci:
> Looks like the new test is failing.

Yes.  As mentioned in above, an STCAL PR needs to be merged before this PR.  The STCAL PR is needed for this PR.

WilliamJamieson:
So #134 is moving the `dynamicdq` module to this package. I think we should wait to merge this PR until #134 is merged. Then you can combine your changes from spacetelescope/stcal#145 into this PR.

WilliamJamieson:
@kmacdonald-stsci #134 has now been merged, can you move your changes from spacetelescope/stcal#145 into this PR?

WilliamJamieson (changes requested):
LGTM, however please add a change log entry

WilliamJamieson (approved):

src/stdatamodels/jwst/datamodels/reference.py
```diff
@@ -41,6 +41,23 @@ def validate(self):
             self.print_err(f'Model.meta is missing values for {to_fix}')
         super().validate()
 
+    def save(self, path, dir_path=None, *args, **kwargs):
+        """
+        Save data model.  If the 'dq' and 'dq_def' they need special
```
hbushouse, line 46: minor nit, but I think there's a typo in this line. Need a verb or something after "If the dq and dq_def ...", i.e. "exist"?

hbushouse:
Are we agreed that this is ready to merge now?

WilliamJamieson:
> Are we agreed that this is ready to merge now?

I think its ready.

kmacdonald-stsci:
> Are we agreed that this is ready to merge now?

Yes.

PR #134, opened 2023-03-01
JP-3110 Move `dqflags` and related from `stcal` to `stdatamodels`

### LINKED ISSUE #133: Remove stcal as dependency
https://jira.stsci.edu/browse/JP-3110

*Moving the datamodels from jwst to stdatamodels required adding stcal as a dependency for stdatamodels. This should not be necessary as this code is only needed for the datamodels in jwst.*

WilliamJamieson:
Resolves [JP-3110](https://jira.stsci.edu/browse/JP-3110)

This PR moves:

- `stcal.dqflags`
- `stcal.dynamicdq`
- `stcal.basic_utils`

To `stdatamodels`. These modules appear to only be used by `jwst`.

This is to facilitate making `stdatamodels` independent of the JWST pipeline code, so that users of `jwst` pipeline products can open those products using tools like `jdaviz` without having to install the entire JWST pipeline. Instead, `stdatamodels` contains all the necessary code to unpack the asdf-in-fits objects embedded in some jwst pipeline products.

Fixes #133

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,12 @@
 1.3.0 (unreleased)
 ==================
 
-- 
+Other
+-----
+
+- Move the ``dqflags`` and related code from ``stcal`` to this package
+  so that the ``stcal`` dependency can be dropped. [#134]
+
 
 1.2.0 (2023-03-02)
 ==================
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
     "asdf>=2.14.1",
     "asdf-astropy>=0.3.0",
     "psutil>=5.7.2",
-    "numpy>=1.16",
+    "numpy>=1.18",
     "astropy>=5.0.4",
 ]
 dynamic = [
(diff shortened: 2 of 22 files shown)


pllim:
This fixes #133

tox.ini
```diff
@@ -51,7 +51,7 @@ extras =
 deps =
     xdist: pytest-xdist
     cov: pytest-cov
-    jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
+    jwst: jwst[test] @ git+https://github.com/WilliamJamieson/jwst.git@refactor/use_stdatamodels_dq
```
hbushouse, line 54: I'm assuming this is just temporary for testing and we're not going to leave it this way in the master branch?

hbushouse (approved):
Looks reasonable to me. Just one question.

tox.ini
WilliamJamieson: Correct, something like this is necessary until spacetelescope/jwst#7480 is merged because the upper pin on stdatamodels in jwst causes the changes in the stdatamodels PRs to overridden by the install of jwst.

nden (approved):

PR #138, opened 2023-03-03
fix coverage measurement

braingram:
The coverage measurement does not appear to be working correctly.

For example, this function:
https://github.com/spacetelescope/stdatamodels/blob/dc0da6d55da8e6251ef04b55551d14246c488a48/src/stdatamodels/asdf_in_fits.py#L13

is covered by this test:
https://github.com/spacetelescope/stdatamodels/blob/dc0da6d55da8e6251ef04b55551d14246c488a48/tests/test_asdf_in_fits.py#L35

The test does appear to be running (confirmed locally) and coverage reports run locally outside tox (with a development install of the package) show increased coverage.

However this function does not appear to have coverage in the CI.
https://app.codecov.io/gh/spacetelescope/stdatamodels/blob/master/src/stdatamodels/asdf_in_fits.py

I tested a few of the solutions mentioned here, which describe how pytest will test the installed package yet coverage is measured against the current directory (for our current configuration):
https://stackoverflow.com/questions/58696476/tox-0-coverage

and the 'usedevelop' option appears to improve the coverage measurement.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,12 +52,14 @@ deps =
     xdist: pytest-xdist
     cov: pytest-cov
     jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
+use_develop=
+    cov: true
 commands_pre =
     pip freeze
 commands =
     pytest \
     xdist: -n auto \
-    cov: --cov=. --cov-config=pyproject.toml --cov-report=term-missing --cov-report=xml \
+    cov: --cov=src --cov-config=pyproject.toml --cov-report=term-missing --cov-report=xml \
     jwst: --pyargs jwst --ignore-glob=*/scripts/* \
     # TODO: fix bug with `.finalize()` in `jwst.associations`
     jwst: --ignore-glob=*/associations/tests/test_dms.py \
--- a/tox.ini
+++ b/tox.ini
@@ -52,8 +52,9 @@ deps =
     xdist: pytest-xdist
     cov: pytest-cov
     jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
-use_develop=
-    cov: true
+package=
+    cov: editable
+    !cov: wheel
 commands_pre =
     pip freeze
 commands =
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,3 +148,8 @@ norecursedirs = [
 
 [tool.setuptools_scm]
 write_to = "src/stdatamodels/_version.py"
+
+[tool.coverage.run]
+omit = [
+    '*/tests/*',
+]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ minversion = "4.6"
 doctest_plus = true
 doctest_rst = true
 text_file_format = "rst"
-addopts = "--open-files"
+addopts = "--open-files --color=yes"
 testpaths = [
     "tests",
     "src/stdatamodels/jwst",


braingram:
I'm not sure if this is best strategy but it does appear to fix the issue with the example:
https://github.com/spacetelescope/stdatamodels/actions/runs/4326528146/jobs/7554058631#step:12:178

zacharyburnett:
I think I remember something about ASDF messing with coverage when running `pytest-cov`

WilliamJamieson:
> I think I remember something about ASDF messing with coverage when running `pytest-cov`
> 
> EDIT: here: [asdf-format/asdf@`3c83a49`/tox.ini#L27-L29](https://github.com/asdf-format/asdf/blob/3c83a49de6ab7455a4cd2bcc4aea65704a582db3/tox.ini#L27-L29)
> 
> ```ini
> # coverage run must be used because the pytest-asdf plugin will interfere
> # with proper coverage measurement due to the order pytest loads its
> # entry points.
> commands =
>     coverage: coverage run --source=asdf --rcfile={tox_root}/pyproject.toml -m \
>     pytest \
> ```
> 
> Is `pytest-asdf` used here?

This is only relevant to `asdf` itself. Its because `asdf` gets loaded by the `pytest-asdf` plugin, which gets loaded before `pytest-cov` gets loaded.

tox.ini
```diff
@@ -52,12 +52,14 @@ deps =
     xdist: pytest-xdist
     cov: pytest-cov
     jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
+use_develop=
+    cov: true
```
zacharyburnett, line 56: would this also work with `package`? 
```ini
package: 
    cov: editable
    !cov: wheel
```

zacharyburnett (approved):
LGTM

WilliamJamieson:
To make the coverage more accurate you might want to consider configuring `coverage` to omit some files from consideration, see: https://github.com/asdf-format/asdf/blob/3c83a49de6ab7455a4cd2bcc4aea65704a582db3/pyproject.toml#L123-L152

tox.ini
braingram: That does appear to work. Is that preferable?

braingram:
Also, jwst CI failure expected because of the version conflict and pkg_resources use in stpipe: https://github.com/spacetelescope/stpipe/pull/84

> To make the coverage more accurate you might want to consider configuring `coverage` to omit some files from consideration, see: https://github.com/asdf-format/asdf/blob/3c83a49de6ab7455a4cd2bcc4aea65704a582db3/pyproject.toml#L123-L152

Good idea. Any suggestions for files in `src` to ignore?

WilliamJamieson:
> Good idea. Any suggestions for files in `src` to ignore?

Really we should be ignoring any of the test files, so `./tests` and then any of the test directories in `src`. At some point those test directories can probably be factored out into the `./tests` directory. Currently they are just an artifact of the move of `jwst.datamodels` and `jwst.transforms`.

WilliamJamieson:
Also, if you do end-up modifying the `pyproject.toml` file could you sneak an `--color=yes` into the `pytest` configuration's `adopts`? Like: https://github.com/asdf-format/asdf/blob/3c83a49de6ab7455a4cd2bcc4aea65704a582db3/pyproject.toml#L121

I find this handy so that if I browse to an on-going job I can tell if something has already failed in the job because the output will start displaying in red text. This makes it easy to tell if something has gone wrong for long CI jobs.

If this is too out of scope I can make it a separate PR.

tox.ini
braingram: The suggested `package` option seems more readable. I switched it to that and the coverage job passed (with similar coverage, differences attributable to files omitted from coverage https://github.com/spacetelescope/stdatamodels/pull/138#issuecomment-1454028292)

PR #141, opened 2023-03-09
import `tomli` as `tomllib` for `docs` builds before Python 3.11

zacharyburnett:
`tomli` is implemented read-only as `tomllib` in Python 3.11

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -6,8 +6,10 @@ from datetime import datetime
 import importlib
 
 import stsci_rtd_theme
-import tomli
-
+if sys.version_info < (3, 11):
+    import tomli as tomllib
+else:
+    import tomllib
 
 def setup(app):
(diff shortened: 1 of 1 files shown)


braingram:
Changes look good to me.

While looking at what python version is used, I'm seeing readthedocs (based on the build log) uses python 3.8. and the minimum supported python version for this package says it's 3.6:
https://github.com/spacetelescope/stdatamodels/blob/2a418cb46833f30f2690141efba913ccecbff0f3/pyproject.toml#L4
but we only test down to 3.8:
https://github.com/spacetelescope/stdatamodels/blob/2a418cb46833f30f2690141efba913ccecbff0f3/.github/workflows/ci.yml#L48

I'll open an issue for that.

PR #143, opened 2023-03-09
drop support for Python 3.8

### LINKED ISSUE #142: Minimum supported python version is out-of-date
The minimum supported python version for this package says it's 3.6:
https://github.com/spacetelescope/stdatamodels/blob/2a418cb46833f30f2690141efba913ccecbff0f3/pyproject.toml#L4
but we only test down to 3.8:
https://github.com/spacetelescope/stdatamodels/blob/2a418cb46833f30f2690141efba913ccecbff0f3/.github/workflows/ci.yml#L48
Found while reviewing: #141

zacharyburnett:
Closes #142 

This PR drops Python 3.8 and ups the required Python version of this package.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "stdatamodels"
 description = "Core support for DataModel classes used in calibration pipelines"
-requires-python = ">=3.6"
+requires-python = ">=3.9"
```
braingram, line 4: JWST currently supports 3.8:
https://github.com/spacetelescope/jwst/blob/master/setup.cfg#L26

Is there anything in stdatamodels that is incompatible with python 3.8? The CI was passing for this version so I'd vote for keeping it.

braingram (changes requested):

pyproject.toml
zacharyburnett: that is fair, but when dependencies drop support for 3.8 then users might have issues with the rest of our dependencies

braingram: Agreed, we should re-check when we bump dependency versions. At the moment they all appear to support 3.8.

WilliamJamieson: The next releases of astropy and numpy will not support 3.8. `numpy` has already dropped 3.8 support in its development branch and `astropy` is working towards doing that as well.

WilliamJamieson: However, since `stdatamodels` needs to be released sooner rather than later, the min version of python should be 3.8, not 3.9

nden (approved):
I think it's time we merge this.

braingram:
> I think it's time we merge this.

Should 3.8 be dropped in jwst first? It still supports 3.8 and tests it in the CI.
https://github.com/spacetelescope/jwst/blob/f7f2fdb17ca2ae96b6c37fee9727fbf89dd572b6/.github/workflows/ci.yml#L67-L70

zacharyburnett:
> > I think it's time we merge this.
> 
> Should 3.8 be dropped in jwst first? It still supports 3.8 and tests it in the CI. https://github.com/spacetelescope/jwst/blob/f7f2fdb17ca2ae96b6c37fee9727fbf89dd572b6/.github/workflows/ci.yml#L67-L70

Good point, I made a PR here: https://github.com/spacetelescope/jwst/pull/7552

braingram:
Also jdaviz now depends on stdatamodels and supports and tests against 3.8:
https://github.com/spacetelescope/jdaviz/blob/6a2892c1c321547be7ea11756fadb086387a49b3/setup.cfg#L18
although I'm not seeing anything in the jdaviz ci that tests against development versions so dropping 3.8 for stdatamodels shouldn't present an issue for them until the next stdatamodels release.

braingram (approved):
LGTM.

WilliamJamieson:
> Also jdaviz now depends on stdatamodels and supports and tests against 3.8: [spacetelescope/jdaviz@`6a2892c`/setup.cfg#L18](https://github.com/spacetelescope/jdaviz/blob/6a2892c1c321547be7ea11756fadb086387a49b3/setup.cfg#L18) although I'm not seeing anything in the jdaviz ci that tests against development versions so dropping 3.8 for stdatamodels shouldn't present an issue for them until the next stdatamodels release.

Tagging @pllim for comment about `jdaviz`.

pllim:
As discussed offline, Jdaviz won't be affected as long as older version still work for our purpose. Though we might bump it soon anyway because we need something from astropy v5.3 that is not related to stdatamodels, so don't worry about us but thanks for asking!

PR #149, opened 2023-03-15
JP-3005: Update core schema for new step status keyword in metadata

tapastro:
Resolves [JP-3005](https://jira.stsci.edu/browse/JP-3005)

Closes #

This PR adds the cal_step keyword to denote whether pixel replacement was run, skipped, etc. It appears that I don't have many permissions here, as I can't add milestones/labels or request reviewers.

**Checklist**
- [ ] updated relevant tests
- [x] updated relevant documentation
- [x] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
@@ -2402,6 +2402,11 @@ properties:
             type: string
             fits_keyword: S_PHOTOM
             blend_table: True
+          pixel_replace:
+            title: Pixel Replacment
+            type: string
+            fits_keyword: S_PXREPL
+            blend_table: True
           ramp_fit:
             title: Ramp Fitting
             type: string
--- a/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
@@ -2402,7 +2402,7 @@ properties:
             type: string
             fits_keyword: S_PHOTOM
             blend_table: True
-          pixel_replace:
+          pixel_replacement:
             title: Pixel Replacment
             type: string
             fits_keyword: S_PXREPL
--- a/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
@@ -2402,7 +2402,7 @@ properties:
             type: string
             fits_keyword: S_PHOTOM
             blend_table: True
-          pixel_replacement:
+          pixel_replace:
             title: Pixel Replacment
             type: string
             fits_keyword: S_PXREPL
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ Other
 
 - Add units to BARTDELT and HELIDELT jwst keywords in datamodels schema. [#147]
 
+- Add pixel replacement step keyword to jwst core schema. [#149]
+
 1.3.0 (2023-03-13)
 ==================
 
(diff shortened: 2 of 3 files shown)


src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2402,6 +2402,11 @@ properties:
             type: string
             fits_keyword: S_PHOTOM
             blend_table: True
+          pixel_replace:
```
hbushouse, line 2405: not a big deal, but could go ahead and just spell out the full name to `pixel_replacement`

hbushouse (approved):
LGTM

src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2402,6 +2402,11 @@ properties:
             type: string
             fits_keyword: S_PHOTOM
             blend_table: True
+          pixel_replacement:
```
hbushouse, line 2405: Ack. I see from https://github.com/spacetelescope/jwst/pull/7398/files that the actual name of the step is in fact "pixel_replace" (not replace*ment*), so I take back my previous suggestion and this should be changed back to just "pixel_replace". Sorry about the hassle.

tapastro: No worries, flipped it back.

hbushouse (approved):

hbushouse:
> @tapastro and @hbushouse It looks like the failure in the jwst CI job is expected (due to the changes in this PR): https://github.com/spacetelescope/stdatamodels/actions/runs/4622302983/jobs/8174828584?pr=149#step:10:192
> 
> However the failing jwst job is preventing this from being merged because of the branch protections. Any objection to removing the jwst job passing as a requirement for PRs in this package?

I would not permanently disable the jwst job as a requirement for passing. I/we (those of us with appropriate privilege) can just override that requirement on a case-by-case basis when deemed appropriate (like this one). I just check the magic box that says "Merge without waiting for requirements to be met" and we're good to go.

hbushouse:
Oh no, that would be WAY too dangerous!  ;-)

braingram:
I was in the mood for some Kenny Loggins :)

@tapastro are you now able to use the 'merge without waiting for requirements' option. I added you with 'write' permissions to this repo. Looking over the github docs I don't think this is high enough permissions but figured I'd check. If you're not able to see the options appear to be:
- add you to the stdatamodels-maintainers team, which will mean you receive notifications due to your inclusion in the CODEOWNERS file.
- add you as an admin to this repository, which will give you additional permissions beyond allowing for bypassing branch protections
- specifically add you as allowed to bypass branch protections

I'm also happy to just merge this PR if the added permissions seem too complicated for how often you expect to make these types of changes. I'm happy with any of the above options, the first seems reasonable to me if you expect to make additional changes to stdatamodels.

tapastro:
> @tapastro are you now able to use the 'merge without waiting for requirements' option?

No, though it does look like I could merge if the requirements passed. Thanks!

braingram:
> No, though it does look like I could merge if the requirements passed. Thanks!

Thanks for checking. @tapastro I added you to the stdatamodels-maintainer team. Mind giving it one more try?

hbushouse:
I would like to hold off just a bit before merging this, because as soon as we do, anyone using stdatamodels from github/master may run into trouble until necessary updates in the jwst package have also been merged. Let's wait till the PR over in JWST is ready and then we'll merge them both at nearly the same time.

tapastro:
> > No, though it does look like I could merge if the requirements passed. Thanks!
> 
> Thanks for checking. @tapastro I added you to the stdatamodels-maintainer team. Mind giving it one more try?

I do appear to have unnatural powers now. 😟

PR #157, opened 2023-04-20
Updates nirspec cubepar reference to use driz wavelength table

jemorrison:
Resolves [JP-3047](https://jira.stsci.edu/browse/JP-3047)

This PR updates the jwst cubepars reference file data model to include wavelength arrays for the drizzle weighting method.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/datamodels/ifucubepars.py
+++ b/src/stdatamodels/jwst/datamodels/ifucubepars.py
@@ -36,13 +36,13 @@ class NirspecIFUCubeParsModel(ReferenceFileModel):
     ifucubepars_high_emsm_wavetable : numpy table
          default IFU cube high resolution emsm  wavetable
 
-    ifucubepars_prism_driz_wavetable : numpy table
+    ifucubepars_prism_driz_wavetable : numpy float32 array
          default IFU cube prism drizzle wavetable
 
-    ifucubepars_med_driz_wavetable : numpy table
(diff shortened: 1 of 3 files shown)


hbushouse (commented):
Model updates look OK. Needs a change log entry.

src/stdatamodels/jwst/datamodels/ifucubepars.py
```diff
@@ -35,6 +35,15 @@ class NirspecIFUCubeParsModel(ReferenceFileModel):
 
     ifucubepars_high_emsm_wavetable : numpy table
          default IFU cube high resolution emsm  wavetable
+
+    ifucubepars_prism_driz_wavetable : numpy table
```
nden, line 39: What is a `numpy table`?

jemorrison:
I can still NOT get stdatamodels pip installed. I can not fully test this change until the error is one.
DO NOT MERGE YET. I need to test and then we need to get the new cubepars file into CRDS. 

I still get an error: ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
jwst 1.10.2.dev16+gc1c575c15.d20230426 requires stdatamodels<1.4.1,>=1.4.0, but you have stdatamodels 1.4.1.dev3+g95570d5 which is incompatible.

PR #162, opened 2023-05-04
Update the NIRSpec Cubepar reference file to read in drizzle wavelength arrays

jemorrison:
Is used to resolve [JP-3047](https://jira.stsci.edu/browse/JP-3047)

This PR makes updates to the NIRSpec cubepar reference file data model to read in 3 new extensions. These extensions have names of MULTICHAN_PRISM_DRIZZLE, MULTICHAN_MED_DRIZZLE, and  MULTICHAN_HIGH_DRIZZLE.
These extensions are used to build ifu cubes when weighting='drizzle'. In particular these arrays are used when the grating/filters in the association are not the same and the wavelength range will then be non-linear. For NIRSpec all the gratings need to be either all  prism, all  medium, or all high resolution (i.e no mixture of medium and high resolution gratings are allowed). 
According to the type of grating (prism, medium, or high) the wavelength array is read in from the new extensions and used when creating an IFU cube. 

The changes to the jwst pipeline that support these updates to the cubepars reference file are in JWST pipeline #7559

We need to  merge this PR before the JWST PR #7559. I have tested that if the cubepar reference file does not have these wavelength arrays the code will still work. The arrays are just not read in. But we need to coordinate getting the new reference file in the CRDS and merging the JWST code.  

**Checklist**
- [X] added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,8 +6,12 @@ Other
 
 - Provide second-order polynomial transforms for NIRCam WFSS grisms. [#124]
 
+
 - Deprecate ``stdatamodels.jwst.datamodels.DataModel`` in favor of
   ``stdatamodels.jwst.datamodels.JwstDataModel``. [#160]
+
+- Add wavelength tables for NIRSpec Drizzle cubepars reference file [#157]
+
 
 1.4.0 (2023-04-19)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@ Other
 - Deprecate ``stdatamodels.jwst.datamodels.DataModel`` in favor of
   ``stdatamodels.jwst.datamodels.JwstDataModel``. [#160]
 
-- Add wavelength tables for NIRSpec Drizzle cubepars reference file [#157]
+- Add wavelength tables for NIRSpec Drizzle cubepars reference file [#162]
 
 
 1.4.0 (2023-04-19)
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -6,9 +6,13 @@ Other
 
 - Provide second-order polynomial transforms for NIRCam WFSS grisms. [#124]
 
+
 - Deprecate ``stdatamodels.jwst.datamodels.DataModel`` in favor of
   ``stdatamodels.jwst.datamodels.JwstDataModel``. [#160]
 
+- Add wavelength tables for NIRSpec Drizzle cubepars reference file [#157]
```
hbushouse, line 13: The PR number here is wrong. Should be 162, not 157. That's why the CI check is failing.

hbushouse (approved):
This looks good to me now and we've verified through regression tests using the corresponding updates in jwst that it appears to work OK and is benign when used with existing cubepars ref files. So this should be safe to merge.

hbushouse:
The failing CI test is due to unrelated issues with the stdatamodels updates for NIRCam grism transforms.

hbushouse:
@braingram and @tapastro are you OK with merging this? I only ask because you're the `stdatamodels-maintainers`

tapastro:
> @braingram and @tapastro are you OK with merging this? I only ask because you're the `stdatamodels-maintainers`

Still not sure how I got added to that... Looks good to me.

tapastro (approved):

braingram (approved):
If it looks good to @hbushouse and works with JWST this looks good to me.

PR #153, opened 2023-04-11
Fix open file handles which were not being caught by `pytest-openfiles`

WilliamJamieson:
`pytest-openfiles` was not catching all the open file-handles left during the tests. These handles were often blocked by
other exception handling. The removal of `pytest-openfiles` and replacement with `ResourceWarning` catching pointed this issue out in the form of `PytestUnraisableExceptionWarning`s, because other exceptions were being caught `pytest` could not raise the `ResourceWarning` as an exception.

This PR fixes all the new file handles being caught and turns `PytestUnraisableExceptionWarning`s into errors to prevent this issue in the future.

**Checklist**
- [ ] added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] updated relevant milestone(s)
- [ ] added relevant label(s)

PR #166, opened 2023-06-07
 update to JWST/MIRI MRS photom data model

jemorrison:
Partially Resolves [JP-3224](https://jira.stsci.edu/browse/JP-3224)

Closes #

This PR updates the JWST/MIRI MRS photom reference file to read in the time dependent corrections. 

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/datamodels/photom.py
+++ b/src/stdatamodels/jwst/datamodels/photom.py
@@ -102,19 +102,19 @@ class MirMrsPhotomModel(ReferenceFileModel):
         An array-like object containing pixel-by-pixel size values, in units of
         square arcseconds (arcsec^2).
 
-   timecoeff_ch1: numpy table
+    timecoeff_ch1: numpy table
         A table of time and wavelength dependent throughput corrections
         for channel 1
 
-   timecoeff_ch2: numpy table
+    timecoeff_ch2: numpy table
         A table of time and wavelength dependent throughput corrections
         for channel 2
 
-   timecoeff_ch3: numpy table
+    timecoeff_ch3: numpy table
         A table of time and wavelength dependent throughput corrections
         for channel 3
 
-   timecoeff_ch4: numpy table
+    timecoeff_ch4: numpy table
         A table of time and wavelength dependent throughput corrections
         for channel 4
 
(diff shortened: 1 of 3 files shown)


jemorrison:
I have updated the CHANGE LOG but and pushed it to this PR but it is not coming up as being part of the PR.
I can see it in my jemorrison:miri_photom_model branch 
Also fixed some flake8 changes and pushed them and the file is also not getting into this PR.
Something strange is going on.

braingram:
> I have updated the CHANGE LOG but and pushed it to this PR but it is not coming up as being part of the PR. I can see it in my jemorrison:miri_photom_model branch Also fixed some flake8 changes and pushed them and the file is also not getting into this PR. Something strange is going on.

braingram:
> I have updated the CHANGE LOG but and pushed it to this PR but it is not coming up as being part of the PR. I can see it in my jemorrison:miri_photom_model branch Also fixed some flake8 changes and pushed them and the file is also not getting into this PR. Something strange is going on.

Not sure what's happening either. I closed and reopened this and it seems to have picked up the commits. If it happens again it might help to just open a new PR as it may just be a temporary issue.

jemorrison:
@braingram Thanks

jemorrison:
@braingram  Do you understand the CI failure with test_downstream ?

braingram:
> > @braingram Do you understand the CI failure with test_downstream ?
> 
> :-/ The errors appear to be related to open files not being closed. But I don't see how that could be related to this PR (perhaps other repositories were updated). I don't see any change in JWST that would explain the errors so it could also be a bad run on github. I re-ran the job and let's see if it has similar failures. If it does, it might help to compare the dependencies with a previously successful run.

I think these might be errors that have been occurring for several months. The first appearance I can see is:
https://github.com/spacetelescope/stdatamodels/actions/runs/4703066425/jobs/8341111754
and these errors might be related to:
https://github.com/spacetelescope/stdatamodels/pull/153

jemorrison:
So does that mean we can ignore that CI failure and merge anyway ?

src/stdatamodels/jwst/datamodels/photom.py
```diff
@@ -101,6 +101,23 @@ class MirMrsPhotomModel(ReferenceFileModel):
     pixsiz : numpy array
         An array-like object containing pixel-by-pixel size values, in units of
         square arcseconds (arcsec^2).
+
+   timecoeff_ch1: numpy table
```
hbushouse, line 105: All of these new attribute name entries need to be indented 1 more space to line up with all previous entries. Also need a whitespace before the colon in each entry.

hbushouse (approved):

PR #171, opened 2023-06-13
remove deprecated models

braingram:
As this is removing deprecated features this should technically trigger a 2.0 release. As I believe the removal is only to remove warnings from test suite runs I don't imagine this is urgent. However I don't see any reason not to move to 2.0.

Removed -> use this instead
  DrizProductModel -> ImageModel
  DataModel -> JwstDataModel
  MultiProductModel -> MultiSlitModel
  MIRIRampModel -> RampModel

See comment below for cross references to jwst and crds.
jwst: https://github.com/spacetelescope/stdatamodels/pull/171#issuecomment-1591511582
crds: https://github.com/spacetelescope/stdatamodels/pull/171#issuecomment-1591781317

Dropping these models required updating the minimum crds to 11.17.1:
https://github.com/spacetelescope/crds/releases/tag/11.17.1

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
`jwst.datamodels` currently import `DataModel` which is causing the failure in the downstream jwst job.
https://github.com/spacetelescope/stdatamodels/actions/runs/5260129931/jobs/9506625097?pr=171#step:10:309
That failure could be fixed by: https://github.com/spacetelescope/jwst/pull/7605 (which will should automatically sync the models in the two packages).

braingram:
The current jwst downstream errors I believe will be addressed in:
https://github.com/spacetelescope/jwst/pull/7609

tapastro (approved):
Looks good to me - presumably 1.6.1 will be released shortly, and separately from 2.0.0, leading to the two unreleased headers in the changelog?

braingram:
> Looks good to me - presumably 1.6.1 will be released shortly, and separately from 2.0.0, leading to the two unreleased headers in the changelog?

Yeah I wasn't quite sure the best way to handle that. The existing changes in the 1.6.1 heading presumably weren't enough to trigger a 2.0.0. I think some options are:

1. release 1.6.1, update this PR to fix the changelog conflict, merge this PR and put all new changes in a 2.0.0 release
2. update this PR to move all changes to 2.0.0 and merge this PR prior to the next release

Given the timing of a JWST release (which doesn't require this PR but might require the other changes in 1.6.1) I'm inclined to go for option '1' and hold off on merging this PR (I can convert it to draft if that's easiest).

braingram:
Making this a draft until after the next jwst release.

emolter:
This would resolve ticket [JP-3585](https://jira.stsci.edu/browse/JP-3585); please ensure this is closed when the PR is merged

braingram:
I'd say let's start the 2.0.0 process. We have a milestone started:
https://github.com/spacetelescope/stdatamodels/milestone/13
that can be expanded when we identify more features/issues.

pyproject.toml
```diff
@@ -49,7 +49,7 @@ test = [
     "psutil",
     "pytest>=4.6.0",
     "pytest-doctestplus",
-    "crds>=11.16.14",
+    "crds>=11.17.1",
```
emolter, line 52: why the bump in CRDS version?

emolter (approved):
LGTM

pyproject.toml
braingram: Thanks for giving this a look.

Earlier version of crds use `DataModel`. See:
https://github.com/spacetelescope/crds/releases/tag/11.17.1
and:
https://github.com/spacetelescope/crds/pull/938

PR #174, opened 2023-06-22
OIFITS compatible schema and AmiOIModel

braingram:
~Opening this as a draft to trigger the CI to see how the jwst downstream handles the new model (in part to test: https://github.com/spacetelescope/jwst/pull/7607)~
Resolves [JP-1713](https://jira.stsci.edu/browse/JP-1713)

Adds a schema `oifits.schema.yaml` to create OIFITS compatible files to be used for a new `AmiOIModel`.

The added test generates a file that is compatible with OIFITS as tested with `oifits-check` from [oifitslib](https://github.com/jsy1001/oifitslib) which 'generally' agrees with the [online OIFITS checker from JMMC](http://oival.jmmc.fr/index.html).

100% OIFITS compliance testing is not be possible with the schema (as valid values for some items depend on values for other items, a dependency that cannot be checked with the schema). Additionally, the schema is constructed based off of the example files (which included OI_VIS OI_VIS2 and OI_T3 tables) and allow for one of each of those tables (and requires that at least one be defined).

There was one keyword 'conflict' with the JWST core schema and OIFITS. Both use the DATE-OBS keyword. JWST appears to use this to store a date whereas OIFITS mentions a date and time. However the wording in OIFITS does not specifically require a date time and the OIFITS checkers seem ok with a date. To avoid complications with existing JWST steps this keyword is being left as a date.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)


braingram:
JWST regression tests running at: https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/756/pipeline/202

braingram:
Regression tests showed 23 failures in `jwst.ramp_fitting.tests` @hbushouse are these expected for other reasons?

hbushouse:
> Regression tests showed 23 failures in `jwst.ramp_fitting.tests` @hbushouse are these expected for other reasons?

Yes, it's because some ramp_fitting updates were recently merged into stcal/main, but updates to the unit tests sitting over in jwst/ramp_fitting/tests/ have not yet been merged, so those unit tests are producing expected differences.

hbushouse (approved):
Looks good to me overall. I guess we'll need to confer with the NIRISS AMI folks as to whether it's OK to proceed with only having a date value in the DATE-OBS keyword.

braingram:
> Looks good to me overall. I guess we'll need to confer with the NIRISS AMI folks as to whether it's OK to proceed with only having a date value in the DATE-OBS keyword.

Thanks for the review!

Based off of https://github.com/spacetelescope/jwst/issues/5351#issuecomment-1611667377 merging seems like a reasonable next step.

PR #176, opened 2023-06-28
JP-2506: Updating How the ZEROFRAME Array gets Allocated

kmacdonald-stsci:
Resolves [JP-2506](https://jira.stsci.edu/browse/JP-2506)

This PR addresses the ZEROFRAME array for the RampModel getting incorrectly allocated with dimensions (ngroups, nrows, ncols), when it should have dimensions (nints, nrows, ncols).  This RampModel array gets automatically allocated when referenced.  The dimensions for all arrays of dimension `ndim` is that it should have the dimensions `dims[-ndims:]` where `dims` is the dimensions of the primary array.  This assumption is incorrect.  There is now a special handler for the ZEROFRAME array to ensure it gets allocated with the proper dimensions.

This changed doesn't cause any CI failures in `STCAL`, `JWST`, nor `stdatamodels`.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Other
 -----
 
+- Update the allocation of the ZEROFRAME array for the RampModel. [#176]
+
 - Added two new header keywords to the JWST core schema exposure section: PRIMECRS and
   EXTNCRS, which are used to record the rate of primary cosmic rays and extended cosmic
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
Looks good.

PR #179, opened 2023-07-07
JP-3302: Add TARGCAT and TARGDESC keywords

hbushouse:
Resolves [JP-3302](https://jira.stsci.edu/browse/JP-3302)

Closes https://github.com/spacetelescope/jwst/issues/7710

This PR adds definitions for the 2 new JWST keywords "TARGCAT" and "TARGDESC" to the core schema.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,14 @@
+1.7.1 (unreleased)
+==================
+
+Other
+-----
+
+- Added two new header keywords to the JWST core schema target section:
+  TARGCAT and TARGDESC, which record the target category and description
+  as given by the user in the APT. [#179]
+
+
 1.7.0 (2023-06-29)
 ==================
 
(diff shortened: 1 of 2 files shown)


tapastro (approved):
LGTM - though will APT descriptions exceed the allowable length of a header entry? If so, do we need to truncate here, or are datamodels smart enough to figure that out?

braingram:
Docs failures are due to sphinx-asdf incompatibility with 3.11 and how these docs use the asdf directive.
https://github.com/asdf-format/sphinx-asdf/pull/72

I opened a PR to switch docs builds to 3.10 for this package until the above fix can be applied:
https://github.com/spacetelescope/stdatamodels/pull/180

So far I'm only seeing stdatamodels affected (not jwst, etc) but please let me know if other packages show docs failures.

hbushouse:
> LGTM - though will APT descriptions exceed the allowable length of a header entry? If so, do we need to truncate here, or are datamodels smart enough to figure that out?

The FITSIO routines handle that automatically by either truncating or turning the value into a multi-line record (that almost always happens for the S_REGION keyword value, which is always very long).

PR #180, opened 2023-07-07
build docs with python 3.10

braingram:
sphinx-asdf 0.2.0 fails for some uses in python 3.11 (see: https://github.com/asdf-format/sphinx-asdf/pull/72).

This PR temporarily switches the docs to build in 3.10 until an update can be released for sphinx-asdf.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/docs/rtd_environment.yaml
+++ b/docs/rtd_environment.yaml
@@ -1,8 +1,8 @@
-name: rtd311
+name: rtd310
 channels:
  - conda-forge
  - defaults
 dependencies:
- - python=3.11
+ - python=3.10
  - pip
  - graphviz


tapastro (commented):
Has this been tested with e.g. PR#178 to see if it fixes the issue?

braingram:
> Has this been tested with e.g. PR#178 to see if it fixes the issue?

Thanks for taking a look! Locally I can build with 3.10 (but not 3.11). I opened a test PR that applies the changes in this PR and #178 to see if read the docs builds.

The build passed see build log here: https://readthedocs.org/projects/stdatamodels/builds/21236522/

tapastro (approved):
Ah, I thought the failure was a result of that PR, but I see that it's a general issue with existing documentation - is that right? Regardless, this PR seems reasonable enough.

braingram:
> Ah, I thought the failure was a result of that PR, but I see that it's a general issue with existing documentation - is that right? Regardless, this PR seems reasonable enough.

Yes! I just released sphinx-asdf (used by stdatamodels to build docs) with some changes. One change (a regex modification) combined with an updated dependency makes sphinx-asdf fail for certain uses in python 3.11 (so far the only one I've seen is in stdatamodels).

Thanks for the review. Hopefully I can clean up the mess soon.

PR #184, opened 2023-07-18
remove unused V23ToSkyConverter from jwst.transforms extension 1.0.0

braingram:
V23ToSkyConverter is supported in the jwst.transforms asdf extension version 0.7.0 (see [0.7.0 manifest](https://github.com/spacetelescope/stdatamodels/blob/fbd97cc7ab5a675d36e42db25218f8bffec8f494/src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-0.7.0.yaml#L88)) but is not currently supported in 1.0.0 (see [1.0.0 manifest](https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-1.0.0.yaml))

This removes V23ToSkyConverter from the 1.0.0 extension (which already does not support the tag) and in all prior versions of asdf this converter was ignored.

Previously asdf ignored any Converter that lists tags that are not in the manifest. This behavior is undocumented and may conflict with future asdf changes. asdf is changing Converter.select_tag to allow Converter deferral (see https://github.com/asdf-format/asdf/pull/1561) where a converter can support no tags but still support types (where the converter will convert the object being serialized to a new type which is handled by a different converter).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/transforms/extensions.py
+++ b/src/stdatamodels/jwst/transforms/extensions.py
@@ -8,35 +8,44 @@ from .converters.jwst_models import (Gwa2SlitConverter, Slit2MsaConverter, Logic
                                      CoordsConverter, V23ToSkyConverter)
 
 
-JWST_TRANSFORM_CONVERTERS = [
-    CoordsConverter(),
-    Gwa2SlitConverter(),
-    Slit2MsaConverter(),
-    LogicalConverter(),
-    NirissSOSSConverter(),
-    RefractionIndexConverter(),
-    Rotation3DToGWAConverter(),
-    MIRI_AB2SliceConverter(),
-    NIRCAMGrismDispersionConverter(),
-    NIRISSGrismDispersionConverter(),
-    GratingEquationConverter(),
-    SnellConverter(),
-    V23ToSkyConverter(),
-]
-
 # The order here is important; asdf will prefer to use extensions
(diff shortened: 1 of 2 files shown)


braingram:
Regression tests running at: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/788/

src/stdatamodels/jwst/transforms/extensions.py
```diff
@@ -8,35 +8,44 @@
...
+            LogicalConverter(),
+            NirissSOSSConverter(),
+            RefractionIndexConverter(),
+            Rotation3DToGWAConverter(),
+            MIRI_AB2SliceConverter(),
+            NIRCAMGrismDispersionConverter(),
+            NIRISSGrismDispersionConverter(),
+            GratingEquationConverter(),
+            SnellConverter(),
+            V23ToSkyConverter(),
```
tapastro, line 48: Should this still be here?

nden (approved):
The 0.7 converters are probably not used any more in any data files. 
We can probably run a search on crds files and see if there are any left that use 0.7 versions.
Other than that the archive should only have files using 1.0.
Something to discuss with Howard.

PR #193, opened 2023-08-15
clean up some test warnings and turn warnings into errors in ci

braingram:
This PR cleans up the warnings in the test suite (these were almost all from use of `np.empty` which often produced values that fell outside of the range of the destination datatype when the empty arrays were assigned to a datamodel).

This PR also turns all warnings into errors in the test suite (with 1 exception where `RuntimeWarning` with a message `numpy.ndarray size changed` is ignored as it's produced by importing astropy code).

If there's any objection to turning all warnings into errors I'm happy to remove that part of this PR.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,12 +133,11 @@ testpaths = [
     "src/stdatamodels/jwst",
 ]
 filterwarnings = [
-    # This error replaces pytest-openfiles
-    "error::ResourceWarning",
-    # Sometimes turning ResourceWarnings into errors creates an unraisable exception
-    #   e.g. when pyest catches another exception, this will block the turning of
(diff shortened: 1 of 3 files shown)


braingram:
Changing this to draft until the issue with coverage, xdist and a warning can be sorted.

nden (approved):
LGTM
I'm OK with turning warnings into errors. This will force us to look at them when they appear.

tapastro (approved):
Seems reasonable to me - maybe the story would be different if it immediately failed unit tests 🙂.

braingram:
Thanks for weighing in! I'm sure failures will come soon :)

PR #201, opened 2023-08-29
switch schema refs that use tag to uris

braingram:
asdf will soon warn for refs that are not uris. This PR changes the few remaining `$ref`s that contain tags to uris.

The tag `tag:stsci.edu:asdf/transform/transform-1.1.0` resolves to `http://stsci.edu/schemas/asdf/transform/transform-1.1.0`. All uses of this tag were replaced with the corresponding uri.

Closes #

This PR addresses ...

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,8 @@ Other
 
 - Add ``AmiLgFitModel`` class and schema [#199]
 
+- Switch schema refs from tags to equivalent uris [#201]
+
 
 1.8.0 (2023-08-24)
 ==================
--- a/src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/coords-0.7.0.yaml
+++ b/src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/coords-0.7.0.yaml
@@ -28,7 +28,7 @@ examples:
           model_type: unitless2directional
 
 allOf:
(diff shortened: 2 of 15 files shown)


braingram:
Regression tests running at: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/899/

hbushouse (approved):
Looks OK to me. This has no effect on existing jwst code, right?

braingram:
> Looks OK to me. This has no effect on existing jwst code, right?

Correct!

tldr; asdf internally converts `tag:stsci.edu:asdf/transform/transform-1.1.0` to `http://stsci.edu/schemas/asdf/transform/transform-1.1.0` due to a known "quirk" which will be fixed in asdf 3.0

It's a bit convoluted but old asdf extensions allowed defining a `tag_mapping` (which allowed converting tag string like `tag:stsci.edu:asdf/transform/transform-1.1.0` to url strings like `http://stsci.edu/schemas/asdf/transform/transform-1.1.0`). Due to an internal "quirk" of asdf this tag mapping is applied when a schema is loaded (see: https://github.com/asdf-format/asdf/issues/752) which allowed tag strings to be used as jsonschema `$ref`s. The refs updated here all rely on the `tag_mapping` provided by the asdf `BuiltinExtension` (which will be removed in asdf 3.0). asdf 3.0 will also start warning when a tag is used for a `$ref`.

PR #203, opened 2023-08-31
JP-3244: Add dither pointing RA/Dec to schema

tapastro:
Resolves [JP-3244](https://jira.stsci.edu/browse/JP-3244)

This PR addresses issues in extraction window centering for MIRI LRS fixed slit observations. Target RA/Dec information is not necessarily accurate, and an alternative was desired. The dither offset from slit center is a known and fixed quantity - we can use this with the transforms and wcsinfo metadata to store the RA/Dec of the dither offset position for extraction centering. It must be stored rather than calculated on the fly because the WCS frames necessary for calculation from existing metadata no longer exist after the product is resampled, prior to extraction.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ Other
 
 - Add ``AmiLgFitModel`` class and schema [#199]
 
+- Add ``DITH_RA`` and ``DITH_DEC`` to JWST core schema metadata,
+  to be used in spectral extraction window centering. [#203]
+
 
 1.8.0 (2023-08-24)
 ==================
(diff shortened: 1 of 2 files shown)


hbushouse:
I know this is related to the work being done for [JP-3244](https://jira.stsci.edu/browse/JP-3244) to fix extraction aperture centering for NIRSpec FS and MOS, but there are also several existing tickets explicitly for these kinds of extraction problems in MIRI LRS, such as [JP-2858](https://jira.stsci.edu/browse/JP-2858) and [JP-3021](https://jira.stsci.edu/browse/JP-3021), so referencing those here too.

tapastro:
The initial goal was to have them used for internal purposes only, but they weren't passed through resample_spec if not listed in the schema. I suppose they could be useful to users to see where (modulo a transform) the pipeline placed the extraction center for these exposures. The current PR implementation only populates these for MIRI_LRS-FIXEDSLIT. Apart from a difference in frame, they are not too dissimilar from source_xpos/ypos, but those keywords are defined for slit models, which MIRI LRS does not make use of.

hbushouse (approved):

PR #207, opened 2023-09-11
remove redundant check in rebuild_fits_rec_dtype

### LINKED ISSUE #206: Redundant condition in `rebuild_fits_rec_dtype`
As mentioned in comment: https://github.com/spacetelescope/stdatamodels/pull/205#discussion_r1320844064
`rebuild_fits_rec_dtype` contains a check for 'bad' fits_rec dtypes which appears to contain a useless comparison:
https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/util.py#L327
In this case the first condition will always be true if the second condition is true (as fits does not support unsigned integers). The first condition can likely be removed.

braingram:
unsigned integers are unsupported by FITS and will be saved as signed integers with an offset (by astropy) so there is no need to check that the stored type is signed if the dtype presented to the user is unsigned.

Fixes #206 

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/util.py
+++ b/src/stdatamodels/util.py
@@ -333,7 +333,7 @@ def rebuild_fits_rec_dtype(fits_rec):
     for field_name in dtype.fields:
         table_dtype = dtype[field_name]
         field_dtype = fits_rec.field(field_name).dtype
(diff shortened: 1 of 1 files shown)


braingram:
The removed check does not appear to be redundant.
See comments on #206

PR #216, opened 2023-09-21
TST: Python 3.12

pllim:
Resolves [JP-nnnn](https://jira.stsci.edu/browse/JP-nnnn)
Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes #

This PR addresses ...

**Checklist**
- [ ] added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] updated relevant milestone(s)
- [ ] added relevant label(s)

PR #217, opened 2023-09-21
remove use of pkg_resources

braingram:
pkg_resources is deprecated. It's use in stdatamodels is limited to parsing the asdf version to check if it's at least 2.6. The pyproject.toml dependencies set a lower pin on asdf version 2.15.0 making the check in the code unnecessary:
https://github.com/spacetelescope/stdatamodels/blob/27c397415ad2d96a1149f492b7906c49862fc7c2/pyproject.toml#L15

This PR removes the use of pkg_resources.

This is causing issues for 3.12 testing as seen in: https://github.com/spacetelescope/stdatamodels/pull/216

Regression tests passed with no errors: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/967/

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -3,7 +3,6 @@ from functools import partial
 import hashlib
 import inspect
 import io
-from pkg_resources import parse_version
 import re
 import warnings
 import weakref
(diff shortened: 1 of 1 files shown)


braingram:
Failures are unrelated and appear to be caused by a brand new crds context for jwst made operational today:
https://jwst-crds.stsci.edu/context_table/jwst_1132.pmap

PR #221, opened 2023-09-22
use PyPI upload workflow from OpenAstronomy

zacharyburnett:
Resolves [SCSB-108](https://jira.stsci.edu/browse/SCSB-108)

This PR changes the PyPI upload process to use the OpenAstronomy workflow instead of the STScI one. The workflow in this PR is configured to test the build with each pull request push, and then publish upon an official release.

**Checklist**
- [x] ~added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)~
- [x] ~updated relevant tests~
- [x] ~updated relevant documentation~
- [x] added relevant label(s)


.github/workflows/build.yml
```diff
@@ -0,0 +1,15 @@
...
+  release:
+    types: [ released ]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    with:
+      upload_to_pypi: ${{ github.event_name == 'released'}}
```
braingram, line 13: Thanks for updating this workflow. Should this be `release` instead of `released`? I attempted to look through the github docs but failed to find anything definitive.

braingram, line 13:
```suggestion
      upload_to_pypi: ${{ github.event_name == 'release'}}
```
Based on the test here:
https://github.com/braingram/test/actions/runs/6300990973/job/17105025277#step:2:29
I think this should be 'release'. I wish I could point to docs somewhere that describes this but haven't been able to find any. Any good references @zacharyburnett ?

braingram (changes requested):
Thanks for making this update. One suggestion on the new workflow changing `released` to `release`.

.github/workflows/build.yml
WilliamJamieson, line 13:
```suggestion
      upload_to_pypi: ${{ (github.event_name == 'release') && (github.event.action == 'released') }}
```

@braingram is correct, however we should only be publishing full releases to PyPi, not just anything which is a release even such as editing an existing release's description. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release.

braingram: See https://github.com/spacetelescope/stdatamodels/pull/221#discussion_r1336045600 for a further improvement.

braingram (approved):
LGTM @WilliamJamieson any further changes?

PR #223, opened 2023-09-29
add workflow_dispatch to ci.yml

braingram:
**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
The CI is red because the MIRI apcorr reference files on CRDS were updated to include a CHANNEL parameter which does not match the schema.

WilliamJamieson (approved):

PR #232, opened 2023-10-30
allow control of memmapping for asdf files

braingram:
This PR allows control of memory mapping when opening an asdf file using `DataModel(asdf_filename)`. Control of memory mapping already exists for FITS files (through the `memmap` argument which is reused here).

asdf is making a few changes to memory mapping. This PR will help to prepare stdatamodels for those changes. Specifically, asdf will no longer force closure of memory mapped view of arrays (which can and often results in a segfault when an array access fails during an attempt to access the closed file). Instead, asdf will release all references to the opened memory map and rely on the garbage collector to close the file (see this asdf PR for more details: https://github.com/asdf-format/asdf/pull/1668). This works in most situations but is complicated in a few stdatamodels tests which test for the number of open files without first destroying the model. This PR (by using the default `memmap=False`) allows the effected tests to pass as they no longer use memory mapping.

Regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1025/
show only existing 8 unrelated failures (due to niriss crds file updates).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@ Changes to API
 - Deprecate ``cast_arrays`` argument to ``from_fits_hdu`` and
   ``cast_fits_arrays`` argument to ``Datamodel.__init__`` [#214]
 
+- Use ``DataModel.__init__`` ``memmap`` argument when opening ASDF
+  files [#232]
+
 Other
 -----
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
LGTM

PR #234, opened 2023-11-06
map JWST to OIFITS keywords prior to validate

braingram:
As reported by @rcooper295 using `AmiOIModel.update` results in a `ValidationError`. This is due to the mapping of JWST to OIFITS keywords only during `AmiOIModel.on_save` (which is not called during `update`).

This PR updates `AmiOIModel` to map JWST to OIFITS keywords before calling `validate` (which is called during `update` and also fixes an issue that without this PR `AmiOIModel.validate` would also similarly fail).

Unit tests were added to verify that `AmiOIModel.update` and `AmiOIModel.validate` run successfully for an OIFITS compliant model.

Regression tests run with no errors: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1047/

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 Bug Fixes
 ---------
 
--
+- Fixed ``ValidationError`` during ``AmiOIModel.update`` [#234]
 
 Changes to API
 --------------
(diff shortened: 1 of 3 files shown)


rcooper295:
I tested the changes and they fixed the `ValidationWarning` I encountered when trying to use `AmiOIModel.update`. Thanks Brett!

nden (approved):
LGTM

PR #235, opened 2023-11-08
JP-3455: Changes to the MIRI imager PHOTOM datamodel

mlibralato:
Related to [JP-3455](https://jira.stsci.edu/browse/JP-3455)

This PR is related to proposed changes to the PHOTOM step for the MIRI Imager (and coronographs). We add a time-dependent correction to the PHOTOM value. A new datamodel is needed for the PHOTOM reference file.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/mirimg_photom.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/mirimg_photom.schema.yaml
@@ -23,3 +23,15 @@ allOf:
         datatype: float32
       - name: uncertainty
         datatype: float32
+- type: object
+  properties:
+    correction_table:
+      title: Temporal correction to the PHOTOM value
+      fits_hdu: TIMECORR
+      datatype:
+      - name: amplitude
+        datatype: float32
+      - name: tau
+        datatype: float32
+      - name: t0
+        datatype: float32
--- a/src/stdatamodels/jwst/datamodels/schemas/mirimg_photom.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/mirimg_photom.schema.yaml
@@ -25,9 +25,9 @@ allOf:
         datatype: float32
 - type: object
   properties:
-    correction_table:
+    timecoeff:
       title: Temporal correction to the PHOTOM value
-      fits_hdu: TIMECORR
+      fits_hdu: TIMECOEFF
       datatype:
       - name: amplitude
         datatype: float32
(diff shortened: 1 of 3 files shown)


src/stdatamodels/jwst/datamodels/schemas/mirimg_photom.schema.yaml
```diff
@@ -23,3 +23,15 @@ allOf:
         datatype: float32
       - name: uncertainty
         datatype: float32
+- type: object
+  properties:
+    correction_table:
```
hbushouse, line 28: I know it's been said that a foolish consistency is the hobgoblin of little minds, but for the sake of consistency with the MRS time-dependent entries in the photom ref file, would it be a lot of work to change the title here to "timecoeff" and the fits_hdu to "TIMECOEFF"? That would of course require changes in the photom step code too, but I think it's a pretty simple substitution. The name "correction" is kind of generic and hence it's not obvious at first sight what kind of correction it is.

mlibralato: I can do that and I have no particular opinion on the naming. So you are suggesting to change "Correction_table" to timecoeff"? What about the second name? Fyi, it is technically not the exact same correction the MRS does, but it is OK.

hbushouse: Yes, they could both be named "timecoeff". The first refers to the name used in the data model and the 2nd is the value for the EXTVER header keyword in FITS files.

hbushouse:
This needs an entry added to the CHANGES.rst file.

src/stdatamodels/jwst/datamodels/schemas/mirimg_photom.schema.yaml
mlibralato: Done both here and in the jwst pipeline PR.

hbushouse:
If someone has tested this for backwards compatibility with existing MIRI photom ref files (i.e. do the old ref files load OK using the updated schema), which I think they should, because the absence of the new extension will just be ignored, then we can go ahead and merge this.

mlibralato:
> If someone has tested this for backwards compatibility with existing MIRI photom ref files (i.e. do the old ref files load OK using the updated schema), which I think they should, because the absence of the new extension will just be ignored, then we can go ahead and merge this.

@hbushouse I confirmed that I tested the changes with the old datamodels and the pipeline works (as you said, if the extension is not present, the code is executed as before).

PR #244, opened 2023-12-21
drop use of json_id in walk_and_modify callbacks

braingram:
This PR removes uses of the optional `json_id` argument for callbacks passed to `asdf.treeutil.walk_and_modify`.

In `fits_support` this argument was unused and removal has no effect.

In `schema_editor` it is likely better to not use this argument (it's use might lead to bugs like in https://github.com/asdf-format/asdf/pull/1716).

For some extra context, `json_id` contains the string value under the key "id" of the most closely nested dictionary for the contained object processed in the callback. This is useful for draft 4 jsonschema so any reference fragments can be resolved using the uri stored in the "id" (see [this schema for an example id](https://github.com/spacetelescope/stdatamodels/blob/44a5691d56e001f5f1a9c54167481fb73be14813/src/stdatamodels/jwst/datamodels/schemas/abvegaoffset.schema.yaml#L4)). Keyword files do not follow draft 4 and do not contain an "id" with a uri. Within the schema_editor code, the top level keyword files are combined:
https://github.com/spacetelescope/stdatamodels/blob/44a5691d56e001f5f1a9c54167481fb73be14813/src/stdatamodels/jwst/datamodels/schema_editor.py#L231-L242
before calling 'resolve_references' to resolve references to other keyword files:
https://github.com/spacetelescope/stdatamodels/blob/44a5691d56e001f5f1a9c54167481fb73be14813/src/stdatamodels/jwst/datamodels/schema_editor.py#L244
using the directory as the 'uri' for resolving these references:
https://github.com/spacetelescope/stdatamodels/blob/44a5691d56e001f5f1a9c54167481fb73be14813/src/stdatamodels/jwst/datamodels/schema_editor.py#L393-L399
so that references can be found relative to the directory containing the keyword json files. The presence of an "id" key would break this resolution if it did not contain the directory that contains the keyword json files.

Locally the `test_schema_editor` jwst regression tests pass. Regression test run:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1117/
shows only unrelated errors (due to new emicorr reference file and engdb failure, both of which are occurring on jwst main and the random "missing_msa_fail/nofail").

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,8 @@ Bug Fixes
 Changes to API
 --------------
 
--
+- Remove ``json_id`` argument use for callbacks passed
+  to ``asdf.treeutil.walk_and_modify`` [#244]
 
 Other
 -----
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -357,7 +357,7 @@ def _get_validators(hdulist):
 
 
 def _save_from_schema(hdulist, tree, schema):
-    def datetime_callback(node, json_id):
+    def datetime_callback(node):
         if isinstance(node, datetime.datetime):
             node = time.Time(node)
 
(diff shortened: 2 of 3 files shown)


tapastro (approved):
Looks good, and I tested the keyword comparison workflow on this branch and the schema comparison had no issues.

braingram:
Thanks for the review!

PR #247, opened 2024-01-17
prepare for rename to main

braingram:
This PR updates the CI and docs to refer to a renamed "main" branch.

As no downstream package appears to reference a specific branch for stdatamodels (and instead uses the default branch) we should be able to:
- rename the default branch to "main"
- merge this PR

It's possible codecov may need to be updated to use "main" (if it's not using the default branch).

Github does quite a number of helpful things when the branch is renamed (see https://github.com/github/renaming#renaming-existing-branches) so if acceptable I would propose we rename the branch, merge this PR, and deal with any fallout (although I do not expect any).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: test
 on:
   push:
     branches:
-      - master
+      - main
       - '*.x'
     tags:
       - '*'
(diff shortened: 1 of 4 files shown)


zacharyburnett (approved):
LGTM!

nden (approved):
LGTM

PR #248, opened 2024-01-22
JP-3449: Add TMEASURE keyword to core schema

emolter:
Resolves [JP-3449](https://jira.stsci.edu/browse/JP-3449)

Closes [spacetelescope/jwst#8036]

This PR addresses adding TMEASURE to the JWST core schema.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,9 @@ Other
 - Add ``ngroups`` keyword to ``mask`` schema to match
   parkeys on crds [#249]
 
+- Updated JWST core datamodel schema to include the new
+  TMEASURE keyword for measurement time. [#248]
+
 1.9.0 (2023-12-11)
 ==================
 
--- a/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
@@ -862,6 +862,12 @@ properties:
             fits_hdu: SCI
             blend_rule: sum
             blend_table: True
+          measurement_time:
+            title: "[s] Measurement time"
+            type: number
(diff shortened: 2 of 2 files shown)


braingram (approved):
Schema changes look good to me. Thanks for making this update!

Is there a jwst PR to populate and/or use the new keyword?

emolter:
I am working on JP-3465 but it isn't ready yet; here's a [link to the draft PR](https://github.com/spacetelescope/jwst/pull/8212).

hbushouse:
@braingram This keyword will be populated by SDP code during level-1b processing, just like all the other time-related keywords. There is an existing JSDP ticket to compute the value and populate it, and a JWSTKD ticket to add the definition to the official JWST keyword dictionary. So all bases are covered.

hbushouse:
All the CI failures seem to be due to the same test "test_crds_selectors_vs_datamodel[miri]" and appear to be having a problem with model.meta.exposure.ngroups. No idea why this change would be causing that failure. I've checked the new lines in the core.schema.yaml lines closely and I don't see any syntax errors that would cause other failures in the schema.

tapastro:
> All the CI failures seem to be due to the same test "test_crds_selectors_vs_datamodel[miri]" and appear to be having a problem with model.meta.exposure.ngroups. No idea why this change would be causing that failure. I've checked the new lines in the core.schema.yaml lines closely and I don't see any syntax errors that would cause other failures in the schema.

This was caused by a CRDS update that didn't have a matching datamodel update, addressed in #251.

hbushouse (approved):
Everything looks good now. All CI tests pass.

PR #258, opened 2024-02-02
Add asdf resources test and update schemas for new transform schemas

braingram:
This PR adds a test for "resources" (schemas, manifests, metaschemas) registered with asdf. The test checks:

- that every resource in the various resource directories is registered with asdf
- that the "id" in the resource matches the "uri" used to register the resource

There are a few issues with the resources that this test revealed:
```
===================================================================================================== short test summary info ======================================================================================================
FAILED test_integration.py::test_resource_id[resource19] - AssertionError: id[http://stsci.edu/schemas/jwst_datamodel/nrsmos_apcorr.schema] for resource[/Users/bgraham/projects/src/stdatamodels/src/stdatamodels/jwst/datamodels/schemas/nrsfs_apcorr.schema.yaml] did not return the co...
FAILED test_integration.py::test_resource_id[resource34] - AssertionError: id[http://stsci.edu/schemas/jwst_datamodel/abvegamag_offset.schema] for resource[/Users/bgraham/projects/src/stdatamodels/src/stdatamodels/jwst/datamodels/schemas/abvegaoffset.schema.yaml] was not registered...
FAILED test_integration.py::test_resource_id[resource82] - AssertionError: id[http://stsci.edu/schemas/jwst_datamodel/keyword_lampmode.schema] for resource[/Users/bgraham/projects/src/stdatamodels/src/stdatamodels/jwst/datamodels/schemas/keyword_lampstate.schema.yaml] did not retur...
```

These are all due to the "id" in the resource not matching the "uri" used to register the resource (although the errors differ if the "id" used matches a different "uri" as is the case for nrsfs_apcorr.schema.yaml and keyword_lampstate.schema.yaml). The "uri" is constructed based on the location of the file by the `DirectoryResourceMapping` used to register the resource. For the datamodels schemas:
https://github.com/spacetelescope/stdatamodels/blob/b7c7bcd83b7a32d908e3f63c5dece313b3611f7a/src/stdatamodels/jwst/datamodels/integration.py#L34-L37
using `nrsfs_apcorr.schema.yaml` as an example. This schema will be registered using the uri `http://stsci.edu/schemas/jwst_datamodel/nrsfs_apcorr.schema` which does not match the id in the file (which uses `nrsmos` instead of `nrsfs`).

The issues fixed with this PR are somewhat innocuous as they would only cause problems if the fixed schemas contained `$ref`s (which none do).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,8 @@
 - Change ``integration_number`` from int16 to int32 in ``group``
   schema. [#283]
 
+- Fix datamodel schema ids for abvegaoffset, keyword_lampmode, nrsfs_apcorr [#258]
+
 
 1.10.0 (2024-02-29)
 ===================
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ test = [
     "pytest-doctestplus",
     "crds>=11.16.14",
     "scipy>=1.5",
+    "importlib_resources; python_version < '3.10'"
 ]
 docs = [
     "sphinx",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ classifiers = [
 ]
 dependencies = [
     "asdf>=2.15.0",
+    "asdf-transform-schemas>=0.5.0",
     "asdf-astropy>=0.3.0",
     "psutil>=5.7.2",
     "numpy>=1.18",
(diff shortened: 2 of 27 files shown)


braingram:
Regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1295/
show 4 expected errors from changes in https://github.com/spacetelescope/jwst/pull/8336 and no failures due to the changes in this PR.

braingram:
As discussed on slack, I added a test that schema and tag versions match. To avoid delays or changes to this PR I added the test in a draft PR based off the changes of this PR:
https://github.com/spacetelescope/stdatamodels/pull/286

PR #264, opened 2024-02-15
Update MultiSpec schema to provide keywords for recording extraction characteristics

jemorrison:
Partially resolves [JP-3169](https://jira.stsci.edu/browse/JP-3169)
<!Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/RCAL-nnnn) -->

Is needed to close spacetelescope/jwst#7773

This PR updates the MultiSpec model to include the xstart, xstop, ystart, ystop values (if they are defined) so they can be recorded in the FITS header

The extraction x center and y center were already defined with keyword values of EXTR_X, EXTR_Y.
Since the extraction locations are related to these values I picked
EXTRXSTR: x start of extraction region
EXTRXSTP: x stop of extraction region
EXTRYSTR: y start of extraction region
EXTRYSTP: y stop of extraction region

But these value can be changed if a better set can be found

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
@@ -84,6 +84,26 @@ allOf:
             type: number
             fits_keyword: EXTR_Y
             fits_hdu: EXTRACT1D
+          extraction_xstart:
+            title: X axis start of extraction region
+            type: number
+            fits_keyword: EXTRXSTR
+            fits_hdu: EXTRACT1D
+          extraction_xstop:
+            title: X axis end of extraction region
+            type: number
+            fits_keyword: EXTRXSTP
+            fits_hdu: EXTRACT1D
+          extraction_ystart:
+            title: Y axis start of extraction region
+            type: number
+            fits_keyword: EXTRYSTR
+            fits_hdu: EXTRACT1D
+          extraction_ystop:
+            title: Y axis end of extraction region
+            type: number
+            fits_keyword: EXTRYSTP
+            fits_hdu: EXTRACT1D
           shutter_state:
             title: All (open and close) shutters in a slit
             type: string
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,10 +25,6 @@ Other
   options to ``DarkModel`` and ``MIRIDarkModel``. Add the array extension
   to the ``RampModel``, for tracking the average dark current. [#265]
 
-- Add new JWST keywords ``EXTRXSTR``, ``EXTRXSTP``, ``EXTRYSTR``, and ``EXTRYSTP``
-  to the ``MultiSpecModel`` schema for extracted 1-D spectra. [#264]
-
-
 1.9.1 (2024-01-25)
 ==================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,10 @@ Other
   options to ``DarkModel`` and ``MIRIDarkModel``. Add the array extension
   to the ``RampModel``, for tracking the average dark current. [#265]
 
+- Add ``EXTRXSTR``, ``EXTRXSTP``, ``EXTRYSTR``, and ``EXTRYSTP`` keywords
+  to the jwst ``MultiSpec`` schema. [#264]
+
+
 1.9.1 (2024-01-25)
 ==================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,10 @@ Other
   options to ``DarkModel`` and ``MIRIDarkModel``. Add the array extension
   to the ``RampModel``, for tracking the average dark current. [#265]
 
+- Add new JWST keywords ``EXTRXSTR``, ``EXTRXSTP``, ``EXTRYSTR``, and ``EXTRYSTP``
+  to the ``MultiSpecModel`` schema for extracted 1-D spectra. [#264]
+
+
 1.9.1 (2024-01-25)
 ==================
 


jemorrison:
@hbushouse @tapastro  I could not add you to the list of reviewers.  Are the names of the FITS Keywords for these parameters ok to you. This PR will need to be merged before the extract 1d code updates will be able to use them

hbushouse:
We already have a couple of similar sets of keywords that give region limits:

SUBSTRT1, SUBSTRT2, SUBSIZE1, SUBSIZE2 - which record the location and size of subarrays relative to the full detector
SLTSTRT1, SLTSTRT2, SLTSIZE1, SLTSIZE2 - which record the location and size of 2-D cutouts performed in extract_2d

These work a bit differently, in that instead of including explicit start/stop values, they only specify the start location and then record the size.

So if we want keywords that record start/stop values, we can't just duplicate these existing sets by simply changing the prefix (i.e. use "EXTR" instead of "SUB" and "SLT"). But I'm wondering if we should at least conform to the same use of axis numbers in the keywords, instead of using "x" and "y"?  Unfortunately we already have a precedent set with the "EXTR_X" and  "EXTR_Y" keywords that are used to record the center of the circular apertures used to make 1-D extractions from IFU cubes. So ... not sure what the best naming convention should be for these new keywords.

Not a big fan of the fact that "EXTRXSTR" and "EXTRXSTP" are so similar looking until you see the difference in the very last letter. But I honestly don't have many better suggestions. So I guess, after all of this, I'm OK with the proposed keyword names.

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
             type: number
             fits_keyword: EXTR_Y
             fits_hdu: EXTRACT1D
+          xstart:
```
hbushouse, line 87: In keeping with the datamodel names "extraction_x" and "extraction_y" (above), I suggest we make these names `extraction_xstart`, `extraction_xstop`, etc.

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
...
+            title: x start of extraction region
```
hbushouse, line 88:
```suggestion
            title: X axis start of extraction region
```

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
...
+            type: number
+            fits_keyword: EXTRXSTR
+            fits_hdu: EXTRACT1D
+          xstop:
+            title: x end of extraction region
```
hbushouse, line 93:
```suggestion
            title: X axis end of extraction region
```

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
...
+            type: number
+            fits_keyword: EXTRXSTP
+            fits_hdu: EXTRACT1D
+          ystart:
+            title: y start of extraction region
```
hbushouse, line 98:
```suggestion
            title: Y axis start of extraction region
```

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
...
+            type: number
+            fits_keyword: EXTRYSTR
+            fits_hdu: EXTRACT1D
+          ystop:
+            title: y end of extraction region
```
hbushouse, line 103:
```suggestion
            title: Y axis end of extraction region
```

hbushouse (commented):
Needs a change log entry

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
hbushouse, line 87:
```suggestion
          extraction_xstart:
```

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
             type: number
             fits_keyword: EXTR_Y
             fits_hdu: EXTRACT1D
+          xstart:
+            title: X axis start of extraction region
+            type: number
+            fits_keyword: EXTRXSTR
+            fits_hdu: EXTRACT1D
+          xstop:
```
hbushouse, line 92:
```suggestion
          extraction_xstop:
```

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
...
+            title: X axis end of extraction region
+            type: number
+            fits_keyword: EXTRXSTP
+            fits_hdu: EXTRACT1D
+          ystart:
```
hbushouse, line 97:
```suggestion
          extraction_ystart:
```

src/stdatamodels/jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -84,6 +84,26 @@ allOf:
...
+            title: Y axis start of extraction region
+            type: number
+            fits_keyword: EXTRYSTR
+            fits_hdu: EXTRACT1D
+          ystop:
```
hbushouse, line 102:
```suggestion
          extraction_ystop:
```

hbushouse (approved):

PR #268, opened 2024-02-29
safely convert FITS_rec for non-schema data

braingram:
Convert ``FITS_rec`` instances read from old files where a hdu was linked in the old schema (but is no longer linked) when rewriting files.

Resolves [JP-3555](https://jira.stsci.edu/browse/JP-3555)

Regression tests run:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1250/

As this is aimed to fix warnings like:
```
  /data1/jenkins/workspace/RT/JWST-devdeps/miniconda/envs/tmp_env0/lib/python3.11/site-packages/asdf/yamlutil.py:282: AsdfConversionWarning: A ndarray subclass (<class 'astropy.io.fits.fitsrec.FITS_rec'>) was converted as a ndarray. This behavior will be removed from a future version of ASDF. See https://asdf.readthedocs.io/en/latest/asdf/config.html#convert-unknown-ndarray-subclasses
```
which by-default do not produce errors. The log output of the regression test run will need to be manually checked (as I did not start this run with `-W error::asdf.exceptions.AsdfConversionWarning`.

No warnings about ndarray subclass conversions are in the above regtest run log.

1 common and unrelated error occurred for `[stable-deps] test_duplicate_names – jwst.associations.tests.test_level3_duplicate` failing to find a log message (likely due to stpipe logging issues).

1 related warning can be seen in the regtest run:
```
jwst/regtest/test_fgs_guider.py::test_fgs_guider[exptype_fgs_id_stack-cal]
/data1/jenkins/workspace/RT/JWST-Developers-Pull-Requests/miniconda/envs/tmp_env2/lib/python3.11/site-packages/asdf/yamlutil.py:333: AsdfConversionWarning: tag:stsci.edu:asdf/core/asdf-1.0.0 is not recognized, converting to raw Python data structure
```
This is caused by the regtest input file:
https://bytesalad.stsci.edu/ui/repos/tree/General/jwst-pipeline/dev/fgs/level1b/exptype_fgs_id_stack_uncal.fits?clearFilter=true
containing outdated and invalid asdf
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.2.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.0.0
```
The file reports `ASDF_STANDARD 1.2.0` yet uses the `asdf-1.0.0` tag which should be `asdf-1.1.0` in this [version of the standard](https://github.com/asdf-format/asdf-standard/blob/7d75d4341a9d73b27a2d3d7c1327ef6267b4c80a/resources/manifests/asdf-format.org/core/core-1.2.0.yaml#L7). The same error can be seen by opening the file with `stdatamodels`. Re-saving the file fixes this issue.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -504,14 +504,21 @@ def to_fits(tree, schema, hdulist=None):
     if _ASDF_EXTENSION_NAME in hdulist:
         del hdulist[_ASDF_EXTENSION_NAME]
 
-    hdulist.append(_create_asdf_hdu(tree))
+    # Do not pass "extra_fits" to _create_asdf_hdu
+    # as we've already saved the contents above in _save_extra_fits
+    hdulist.append(_create_asdf_hdu({k: v for k, v in tree.items() if k != 'extra_fits'}))
 
     return hdulist
 
 
 def _create_asdf_hdu(tree):
     buffer = io.BytesIO()
-    asdf.AsdfFile(tree).write_to(buffer)
+    # convert all FITS_rec instances to numpy arrays, this is needed as
+    # some arrays loaded from the fits data for old files may not be defined
+    # in the current schemas. These will be loaded as FITS_rec instances but
+    # not linked back (and safely converted) on write if they are removed
+    # from the schema.
+    asdf.AsdfFile(util.convert_fitsrec_to_array_in_tree(tree)).write_to(buffer)
     buffer.seek(0)
 
     data = np.array(buffer.getbuffer(), dtype=np.uint8)[None, :]
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -504,9 +504,7 @@ def to_fits(tree, schema, hdulist=None):
     if _ASDF_EXTENSION_NAME in hdulist:
         del hdulist[_ASDF_EXTENSION_NAME]
 
-    # Do not pass "extra_fits" to _create_asdf_hdu
-    # as we've already saved the contents above in _save_extra_fits
-    hdulist.append(_create_asdf_hdu({k: v for k, v in tree.items() if k != 'extra_fits'}))
+    hdulist.append(_create_asdf_hdu(tree))
 
     return hdulist
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,10 @@
 
 - Drop support for python 3.9 [#287]
 
+- Convert ``FITS_rec`` instances read from old files where a
+  hdu was linked in the old schema (but is no longer linked)
+  when rewriting files. [#268]
+
 
 1.10.0 (2024-02-29)
 ===================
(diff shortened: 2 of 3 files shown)


braingram:
CI failures are due to newly available `emicorr` reference file. Addressed in: https://github.com/spacetelescope/stdatamodels/pull/269

braingram:
@emolter This should fix the `AsdfConversionWarning`s you found in the jwst tests.

src/stdatamodels/fits_support.py
```diff
@@ -497,14 +497,21 @@ def to_fits(tree, schema, hdulist=None):
     if _ASDF_EXTENSION_NAME in hdulist:
         del hdulist[_ASDF_EXTENSION_NAME]
 
-    hdulist.append(_create_asdf_hdu(tree))
+    # Do not pass "extra_fits" to _create_asdf_hdu
+    # as we've already saved the contents above in _save_extra_fits
+    hdulist.append(_create_asdf_hdu({k: v for k, v in tree.items() if k != 'extra_fits'}))
```
braingram, line 502: The contents of `extra_fits` are saved a few lines prior:
https://github.com/spacetelescope/stdatamodels/blob/6b62413268edace5295e82e62911875e65b99629/src/stdatamodels/fits_support.py#L491
This line strips the contents of `extra_fits` prior to saving. Note that the contents are already stripped on read:
https://github.com/spacetelescope/stdatamodels/blob/6b62413268edace5295e82e62911875e65b99629/src/stdatamodels/fits_support.py#L625-L628

braingram: I'm removing this as testing this with other files revealed an odd behavior of `extra_fits` where it sometimes does not load. Until I understand why it sometimes does not load I'd rather leave the duplicate data in place.

braingram:
I converted this back to draft until I can run another regtest run after removing the fix for the duplicate `extra_fits` data (see [7d0787a](https://github.com/spacetelescope/stdatamodels/pull/268/commits/7d0787a44df63aa9a78bee90944d8044a0223477)). That will need to be addressed separately.

braingram:
To provide a bit more context on this PR.

This PR adds a (hopefully) final pass safely converting `FITS_rec` instances in the ASDF tree before it's saved. If not safely converted `FITS_rec` instances can do problematic things (like corrupting the input data array or output file, see https://github.com/spacetelescope/stdatamodels/issues/280 for one example). This PR is using the "safe" conversion code that exists in`stdatamodels`:
https://github.com/spacetelescope/stdatamodels/blob/8bbc9e85678ca6dca2d0e8d9b118ad9fa42f0078/src/stdatamodels/util.py#L322
to deal with a case that was revealed during jwst testing. Specifically, when a fits file is opened that contains a table that is not defined by the datamodel schema. The non-schema-defined table will end up in `extra_fits` in the ASDF tree. On write, `stdatamodels` will duplicate this data, writing it once to a FITS extension and also including it in the ASDF extension (serialized to ASDF) (see https://github.com/spacetelescope/stdatamodels/issues/271 for more details on this duplication). Without this PR, the table/`FITS_rec` in the ASDF tree will be passed to asdf in a dangerous state (not safely converted by `stdatamodels`). For asdf 3.0 (to maintain backwards compatibility) asdf will see that `FITS_rec` is a subclass of `ndarray` and treat it as a normal `ndarray` (which can lead to data corruption) and issue a warning. With this PR the `FITS_rec` will be safely converted to an array and the warning and data corruption risks (at least from the asdf side) will be avoided.

PR #270, opened 2024-03-01
deprecate skip_fits_update

braingram:
`skip_fits_update` is [unused in jwst](https://github.com/search?q=repo%3Aspacetelescope%2Fjwst+skip_fits_update&type=code)
as is [SKIP_FITS_UPDATE](https://github.com/search?q=repo%3Aspacetelescope%2Fjwst+SKIP_FITS_UPDATE&type=code).

This PR deprecates their usage. Their eventual removal will allow `stdatamodels` to reduce data duplicated between the fits headers/hdus and the `ASDF` extension (including `extra_fits` which is entirely duplicated in both).

This is a step towards the eventual removal of `skip_fits_update`, see: https://github.com/spacetelescope/stdatamodels/issues/271

**Checklist**
- [x] added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] updated relevant milestone(s)
- [x] added relevant label(s)

PR #276, opened 2024-03-04
provide existing asdf file to validate

braingram:
During validation on assignment `check_value` a call is made to `asdf.schema.validate`. This PR speeds up that call by passing in a `ctx` `AsdfFile` instance (one used a few lines prior).

This should speed up all validated `DataModel` assignments including those performed during `open` (discussed: https://github.com/spacetelescope/stdatamodels/pull/270). For example opening:
https://bytesalad.stsci.edu/ui/repos/tree/General/jwst-pipeline/dev/nircam/image/jw01538046001_03105_00001_nrcblong_rate.fits?clearFilter=true
with `skip_fits_update=False` (which will result in assignment to all parts of the model linked to the fits data) without this PR takes 177 ms. With this PR opening the same model takes 72 ms.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,8 @@
 1.10.1 (unreleased)
 ===================
 
-- 
+- Provide existing ``AsdfFile`` instance to ``validate`` to
+  speed up assignment validation ``check_value``. [#276] 
 
 1.10.0 (2024-02-29)
 ===================
--- a/src/stdatamodels/validate.py
+++ b/src/stdatamodels/validate.py
@@ -152,7 +152,7 @@ def _check_value(value, schema, ctx):
         else:
             validators = YAML_VALIDATORS
 
-        asdf_schema.validate(value, schema=schema, validators=validators)
+        asdf_schema.validate(value, ctx=ctx._asdf, schema=schema, validators=validators)
 
 
 def _error_message(path, error):


braingram:
Regtests ran with only 2 unrelated errors:
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/1261/tests
Both errors are due to missing log messages due to stpipe logging.

tapastro (approved):
Not sure I'm qualified/educated enough to understand this, but I can at least approve it 🙂

braingram:
Thanks!

The "short" of it (let me know if "long" is wanted) is that for asdf to validate something against the schemas it may need to resolve additional schemas and tags. To do this resolution, an `AsdfFile` instance is needed (see the `ctx` argument to `validate`: https://asdf.readthedocs.io/en/latest/api/asdf.schema.validate.html).

Since the call to `validate` in the line changed in this PR:
https://github.com/spacetelescope/stdatamodels/blob/e1ebc483f574f03057e2474bb5cbef38f101735c/src/stdatamodels/validate.py#L155
does not provide a `ctx`, a new `AsdfFile` instance is created which is more expensive than reusing one. As every `DataModel` has an `AsdfFile` instance (`model._asdf`) this PR uses that instance, passing it in as `ctx`, similar to what  is done a few lines prior:
https://github.com/spacetelescope/stdatamodels/blob/e1ebc483f574f03057e2474bb5cbef38f101735c/src/stdatamodels/validate.py#L148
for `custom_tree_to_tagged_tree` (which requires a `ctx` argument: https://asdf.readthedocs.io/en/latest/api/asdf.yamlutil.custom_tree_to_tagged_tree.html).

PR #277, opened 2024-03-04
JP-3465: removed product exposure time TEXPTIME

emolter:
Partially resolves [JP-3465](https://jira.stsci.edu/browse/JP-3465)

This PR addresses removal of the TEXPTIME (product exposure time) keyword, since it is functionally redundant with XPOSURE and EFFEXPTM and is not used by any high-level data processing scripts we know about.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,8 @@
 1.10.1 (unreleased)
 ===================
 
-- 
+- Remove ``TEXPTIME`` keyword from the JWST core datamodel schema
+  because it duplicates the information of ``XPOSURE``. [#277]
 
 1.10.0 (2024-02-29)
 ===================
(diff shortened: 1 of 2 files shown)


braingram:
It looks like https://github.com/spacetelescope/jwst/pull/8212 removes the 2 uses in jwst.

Given the separate packages how should this PR be coordinated to avoid errors?

The current jwst dev pins stdatamodels to:
`stdatamodels>=1.10.0,<1.11.0`
all released version have pins `<1.10` so this PR (which is targeting 1.10.1) will not impact any released versions of jwst. I think that means if we:
- merge the jwst PR first, this should fix the CI here
- merge this PR

will we see any issues?

hbushouse (approved):
LGTM

hbushouse:
https://github.com/spacetelescope/jwst/pull/8212 has been merged, so it should be safe to merge this PR now. We'll need to update stdatamodels pins over in jwst once there's a new stdatamodels release.

PR #283, opened 2024-03-12
JP-3558: update integration_number dtype in group schema

emolter:
Resolves [JP-3558](https://jira.stsci.edu/browse/JP-3558)

Closes [#8323](https://github.com/spacetelescope/jwst/issues/8323)

This PR prevents values from wrapping around the limit of the int16 range in observations with a large number of integrations.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,9 @@
 - Decrease size of ``SPECTYP`` and ``TARGET`` columns in
   ``OI_TARGET`` table of oifits schema to 16 characters. [#281]
 
+- Change ``integration_number`` from int16 to int32 in ``group``
+  schema. [#283]
+
 
 1.10.0 (2024-02-29)
 ===================
--- a/src/stdatamodels/jwst/datamodels/schemas/group.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/group.schema.yaml
@@ -9,7 +9,7 @@ properties:
     fits_hdu: GROUP
     datatype:
     - name: integration_number
-      datatype: int16
+      datatype: int32
     - name: group_number
       datatype: int16
     - name: end_day


braingram:
The change looks good to me.

Is there a regression test run?

Also, is this used in jwst? If so would you point me to the usage? I failed in my (albeit brief) search.

emolter:
@braingram This was triggered by [this ticket](https://jira.stsci.edu/browse/JP-3460), related to issues with the time stamps in tso3 white light curves.  The datamodel keyword int_times is used several times in the files tso_photometry.py and white_light.py.

Regression test run started [here](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1307/).

braingram:
Thanks for the reply. Perhaps I'm missing something but doesn't the ticket refer to a different schema:
https://github.com/spacetelescope/stdatamodels/blob/41d4a43c1bec37d097a25866c1a76d93460ee306/src/stdatamodels/jwst/datamodels/schemas/int_times.schema.yaml#L9-L12

stscirij:
Yes, as far as I know they should have the same information.  The tso code uses the information from the INT_TIMES extension.  However the same information is in the GROUP extension, and the datatype in that table is i2 (16 bit).  So that also needs to be updated so that it can hold the 32 bit value that is now being used to fill the column in the INT_TIMES extension.  I'll submit a new JSDP ticket to cover that work.

hbushouse (approved):
Changes look simple enough. I'm assuming the regtest results will tell us whether this change is backwards compatible with existing files that have this table column defined as int16 in the GROUP extension (I'm assuming it'll just promote the values to int32 when loading the file into a datamodel).

emolter:
There are a bunch of [regtest failures](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1307/#showFailuresLink), but it looks like the steps ran fine, and all of them just require OKify on the truth files.  All 51 of them read:
```
     Data contains differences:
       Extra column integration_number of format J in a
       Extra column integration_number of format I in b
```
At what stage should that OKify be run?  I guess as soon as this gets merged?

hbushouse:
I agree that regtest results look fine (just the expected differences in the data type for the column in the GROUP table extension) and they also prove that the code change is backwards compatible with old FITS files that use the old column data type.

hbushouse (approved):

PR #286, opened 2024-03-18
Add test that tag and schema versions match for transform manifests

braingram:
https://github.com/spacetelescope/stdatamodels/pull/258 almost missed a mismatch between a newly added schema and a tag listed in the manifest.

This PR adds a test to avoid future misses by testing that all schema versions match tag versions. This is not generally required but is the case for the jwst/stdatamodels transforms manifest.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


tapastro (approved):
Thanks for adding the test!

Edit: just needs a rebase before merge.

PR #288, opened 2024-03-19
increase asdf version, switch _force_raw_types for load_yaml

braingram:
Increases minimum asdf version to 3.1.0 (inline with jwst: https://github.com/spacetelescope/jwst/blob/master/pyproject.toml#L22) and switches `_force_raw_types` usage to `load_tree`.

Also updates one test that fails with ASDF standard 1.6.0 as the default due to an incorrect assumption about the tag version:
https://github.com/spacetelescope/stdatamodels/blob/50751a8f91a4cd8a685dfb7bc692a4fe30c08f18/src/stdatamodels/jwst/transforms/converters/tests/test_models.py#L76-L77
assumes the tag will be `1.0.0`. With ASDF standard 1.6.0 the tag will be `1.1.0`:
https://github.com/spacetelescope/stdatamodels/blob/50751a8f91a4cd8a685dfb7bc692a4fe30c08f18/src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-1.1.0.yaml#L46-L47

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,8 @@
   ``skip_fits_update`` was ``False`` and the FITS headers
   will always be read [#270]
 
+- Increase minimum required asdf version [#288]
+
 
 1.10.0 (2024-02-29)
 ===================
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 dependencies = [
-    "asdf>=2.15.0",
+    "asdf>=3.1.0",
     "asdf-transform-schemas>=0.5.0",
     "asdf-astropy>=0.3.0",
     "psutil>=5.7.2",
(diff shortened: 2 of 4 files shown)


tapastro (commented):
Approve the minimum asdf bump - but I don't quite follow the change from _force_raw_types to load_yaml. Does this address a bug, improve performance, prep for a deprecation, or something else entirely?

braingram:
Thanks for taking a look. I should have added more details.

https://github.com/asdf-format/asdf/pull/1700 introduced `load_yaml` to try and replace the (many) uses of the private argument `_force_raw_types`. It was included in asdf 3.1.0 and since jwst increased the minimum version we can follow suit with this PR and swap out all the uses. So sort of dealing with a deprecation as now asdf can consider removing that private argument to clean up `asdf.open`.

tapastro (approved):
Looks good, appreciate the clarification! Sorry for the accidental re-request, @zacharyburnett - misclick 😬

Issue #42, opened 2020-11-18
ListNode methods __getslice__, __setslice__, __delslice__ should be removed


jdavies-st:
The methods `__getslice__`, `__setslice__`, `__delslice__` are all Python2 holdovers and are deprecated in Python 3.  Any functionality needed in these for stdatamodels should be captured in `__getitem__`, `__setitem__` and `__delitem__`.

See

https://docs.python.org/2.0/ref/sequence-methods.html

Generally, they should be handled like:

```python
    def __getitem__(self, i):
        if isinstance(i, slice):
            # Implement what __getslice__ would have done here
            pass
```

Currently, none of these methods are tested in `stdatamodels`, but some do get tested in the `jwst` package it seems.

Would be good to decide if they're needed.

PR #302, opened 2024-05-09
remove python 2 slicing api

braingram:
Fixes: #42

These methods are [no longer supported as of python 3](https://docs.python.org/3.9/whatsnew/3.0.html?highlight=__getslice__#operators-and-special-methods).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


zacharyburnett (approved):
down with Python 2! Long live Python 3!

braingram:
Hopefully we can remove all the 2.0 API before 4.0 is out 😜

PR #306, opened 2024-07-18
replace usages of ``copy_arrays`` with ``memmap``

zacharyburnett:
follow-on to https://github.com/asdf-format/asdf/pull/1797

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 2.0.1 (unreleased)
 ===================
 
-- 
+- replace usages of ``copy_arrays`` with ``memmap`` [#306]
 
 2.0.0 (2024-06-24)
 ===================
--- a/src/stdatamodels/jwst/datamodels/util.py
+++ b/src/stdatamodels/jwst/datamodels/util.py
@@ -128,7 +128,7 @@ def open(init=None, guess=True, memmap=False, **kwargs):
             return ModelContainer(init, **kwargs)
 
         elif file_type == "asdf":
-            asdffile = asdf.open(init, copy_arrays=not memmap)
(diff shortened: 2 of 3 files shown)


braingram (changes requested):
Thanks!

Based on https://github.com/spacetelescope/jwst/pull/8660/files the "copy_arrays" usage is very limited in jwst and I see no uses that pass it to `DataModel.__init__`. It's possible a user might be doing this and it also makes sense to me to remove "copy_arrays" as an allowed keyword argument (eventually).

Would you add a deprecation for passing "copy_arrays" to `DataModel.__init__` and a test for the deprecation? I think another UserWarning is called for here since it seems likely to happen in an interactive session. We can then remove "copy_arrays" for stdatamodels 3.0.0

zacharyburnett:
> Thanks!
> 
> Based on https://github.com/spacetelescope/jwst/pull/8660/files the "copy_arrays" usage is very limited in jwst and I see no uses that pass it to `DataModel.__init__`. It's possible a user might be doing this and it also makes sense to me to remove "copy_arrays" as an allowed keyword argument (eventually).
> 
> Would you add a deprecation for passing "copy_arrays" to `DataModel.__init__` and a test for the deprecation? I think another UserWarning is called for here since it seems likely to happen in an interactive session. We can then remove "copy_arrays" for stdatamodels 3.0.0

Alternatively, we can remove handling `copy_arrays` entirely and outsource that warning to the `AsdfWarning` that'll come up when it is passed to `asdf.open()`; see f5c46d761c3ea876bcd60843e3811a7d4b744a56

braingram (approved):
Much simpler :) Good idea!

PR #318, opened 2024-08-19
JP-1048: added unit test for dq_def roundtrip

emolter:
Resolves [JP-1048](https://jira.stsci.edu/browse/JP-1048)

No associated GitHub issue

This PR adds a simple unit test covering a long-since-fixed bug wherein the `dq_def` array was being corrupted after saving and re-loadong a `FlatModel` with `datamodels.save()` and `datamodels.open()`

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stdatamodels/jwst/datamodels/tests/test_fits.py
+++ b/src/stdatamodels/jwst/datamodels/tests/test_fits.py
@@ -5,7 +5,7 @@ import numpy as np
 import pytest
 
 from stdatamodels.jwst import datamodels
-from stdatamodels.jwst.datamodels import ImageModel, JwstDataModel, RampModel, SpecModel
+from stdatamodels.jwst.datamodels import ImageModel, JwstDataModel, RampModel, SpecModel, FlatModel
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


src/stdatamodels/jwst/datamodels/tests/test_fits.py
```diff
@@ -106,3 +106,25 @@ def test_units_roundtrip(tmp_path):
...
+    fn = tmp_path / "test_flat.fits"
+    flatmodel.save(fn)
+
+    flatmodel2 = datamodels.open(fn)
+    assert np.allclose(flatmodel2.data, flatdata)
+    for i, flag in enumerate(flatmodel2.dq_def):
+        for j in range(2):
+            # tuples do not directly assert True because datamodels.open() returns np.int32
+            # which is not technically of type int. So must check their equality directly
+            assert flag[j] == flatflags[i][j]
```
braingram, line 130:
```suggestion
    with datamodels.open(fn) as flatmodel2:
        assert np.allclose(flatmodel2.data, flatdata)
        for i, flag in enumerate(flatmodel2.dq_def):
            for j in range(2):
                # tuples do not directly assert True because datamodels.open() returns np.int32
                # which is not technically of type int. So must check their equality directly
                assert flag[j] == flatflags[i][j]
```

CHANGES.rst
```diff
@@ -3,6 +3,9 @@
 
 - replace usages of ``copy_arrays`` with ``memmap`` [#306]
 
+- Added unit test coverage for long-fixed bug where saving and re-loading
```
braingram, line 6: I'd say this doesn't need a changelog.

braingram (approved):
Thanks!

PR #320, opened 2024-08-23
JP-3588: Implement new reference file datamodel for PASTASOSS transformations

tapastro:
Partially resolves [JP-3588](https://jira.stsci.edu/browse/JP-3588)

This PR adds a new reference file datamodel for the PASTASOSS algorithm, replacing the existing ATOCA transformation algorithm. This effectively deprecates the spectrace and wavemap datamodels - TBD on how best to remove them from here and/or CRDS.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
🍝

src/stdatamodels/jwst/datamodels/schemas/soss_tracemodel.schema.yaml
```diff
@@ -0,0 +1,49 @@
...
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_exptype.schema
+- $ref: keyword_filter.schema
+- type: object
+  properties:
+    meta:
+      pwcpos_cmd:
+        title: Commanded position
+        type: float32
```
braingram, line 14:
```suggestion
      type: object
      properties:
        pwcpos_cmd:
          title: Commanded position
          type: number
```

src/stdatamodels/jwst/datamodels/schemas/soss_tracemodel.schema.yaml
```diff
@@ -0,0 +1,49 @@
...
+        pivot_x:
+          title: Pixel rotation point
+          type: float64
+        pivot_y:
+          title: Pixel rotation point
+          type: float64
+        trace:
+          title: The trace position
+          ndim: 2
+          datatype: float64
```
braingram, line 32:
```suggestion
      items:
        type: object
        properties:
          spectral_order:
            title: Spectral order
            type: number
            minimum: 1
          pivot_x:
            title: Pixel rotation point
            type: number
          pivot_y:
            title: Pixel rotation point
            type: number
          trace:
            title: The trace position
            ndim: 2
            datatype: float64
```

src/stdatamodels/jwst/datamodels/schemas/soss_tracemodel.schema.yaml
```diff
@@ -0,0 +1,49 @@
...
+        coefficients:
+          title: List of polynomial coefficients
+          type: float64
+        coefficient_labels:
+          title: Coefficient order specification
+          type: string
+        scale_extents:
+          title: Range of values for normalization
+          type: float64
+          ndim: 2
```
braingram, line 49:
```suggestion
        type: object
        properties:
          polynomial_degree:
            title: Polynomial degree of model
            type: number
          coefficients:
            title: List of polynomial coefficients
            type: number
          coefficient_labels:
            title: Coefficient order specification
            type: string
          scale_extents:
            title: Range of values for normalization
            type: float64
            ndim: 2
```

braingram:
The suggestions I added don't yet account for the `ndim` validators. I assume these properties are arrays? If so, the schema will have to be updated to accommodate that.

Would you add a unit test that makes an instance of the new model with expected junk data? That would help me understand what should go in those spots.

src/stdatamodels/jwst/datamodels/__init__.py
```diff
@@ -66,6 +66,7 @@
 from .saturation import SaturationModel
 from .segmap import SegmentationMapModel
 from .slit import SlitModel, SlitDataModel
+from .soss_tracemodel import SossTraceModel
```
braingram, line 69:
```suggestion
from .sosstracemodel import SossTraceModel
```
One nitpick... can we remove the `_` to keep it consistent with the other soss stuff?

src/stdatamodels/jwst/datamodels/schemas/pastasossmodel.schema.yaml
```diff
@@ -0,0 +1,86 @@
...
+- $ref: keyword_exptype.schema
+- $ref: keyword_filter.schema
+- type: object
+  properties:
+    meta:
+      type: object
+      properties:
+        pwcpos_cmd:
+          title: Commanded position
+          type: float32
```
braingram, line 16:
```suggestion
          type: number
```

braingram (changes requested):
There one item in the schema that causes the following to fail (which I think is worth adding as a generic test-all-model-schemas test):
```python
import asdf
s = dm.PastasossModel()
asdf.schema.check_schema(s.schema)
```
Otherwise looks good to me.

Note that by default all `ndim` validation is disabled due to [validate_arrays](https://github.com/spacetelescope/stdatamodels/blob/a49474ce4f5f63423ac20a82ce9dd053bf89d3cf/src/stdatamodels/model_base.py#L67) being default `False`. I have a hunch this is off as an attempt to avoid data loading or something but it is suprising to me that this disables ndim validation.

braingram (approved):
LGTM!

Issue #323, opened 2024-09-09
Add test to check all datamodel schemas


braingram:
As came up https://github.com/spacetelescope/stdatamodels/pull/320#pullrequestreview-2290621622

It could be useful to do some basic schema checks for the datamodels schemas. These could include:
- ~a call to `asdf.schema.check_schema`~ EDIT: this is already handled by pytest-asdf
- checking if all schema keywords are in expected and understood places

PR #324, opened 2024-09-10
Check datamodels against schema

tapastro:
Partially addresses #323 

This PR adds a test to check schemas against instance of jwst datamodels. Might want to add additional test to address second bullet of #323.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@
 - Added ``MEDIUMDEEP2`` and ``MEDIUMDEEP8`` to allowed readout patterns
   in JWST core schema, ``READPATT``, and ``PREADPATT``. [#315]
 
+- Add test to check datamodels against their schema. [#324]
+
 2.0.0 (2024-06-24)
 ===================
 
--- a/src/stdatamodels/jwst/datamodels/tests/test_schema.py
+++ b/src/stdatamodels/jwst/datamodels/tests/test_schema.py
@@ -1,8 +1,17 @@
 from asdf import schema as mschema
 import numpy as np
 from numpy.testing import assert_array_almost_equal
+import stdatamodels.jwst.datamodels as dm
 
-from stdatamodels.jwst.datamodels import JwstDataModel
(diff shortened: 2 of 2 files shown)


braingram:
Thanks for the PR. I am a little confused about our CI at the moment. Let me try to explain.

The asdf pytest plugin is configured to run with this repo:
https://github.com/spacetelescope/stdatamodels/blob/4e37d607f86059d995b95f506d255052b90aed01/pyproject.toml#L139
and the plugin runs [check_schema](https://github.com/asdf-format/asdf/blob/c5cb4daae91fd3e0d76b1a0a04cf36f205de6cdd/pytest_asdf/plugin.py#L141).

When I test this locally (by making a schema invalid) I see the expected failure (using modifications off main). So I think we are should already be testing the schemas in the way proposed in this PR. However, since the tests on https://github.com/spacetelescope/stdatamodels/pull/320 didn't fail for an invalid schema I think our github CI might be misconfigured (perhaps the tests are running from a different path).

I'll take a look but I think once we sort out the CI we should automatically have the schemas tests as is done in this PR.

braingram:
I opened a test PR with the pastasoss schema changed back to the `type: float64` and the CI fails for that schema (as it should):
https://github.com/spacetelescope/stdatamodels/pull/325
I'm not sure why we didn't see the failure before (or perhaps we didn't give it enough time to fail?).

Overall I think that means that we don't need a separate `check_schema` for the schemas (since pytest-asdf is already handling that for us). I'll update the issue.

braingram:
I should also say that I wouldn't be opposed to disabling pytest-asdf and adding our own schema tests (similar to what's in this PR). It would allow the eventual removal of pytest-asdf (although it is used in many other places). I don't think pytest-asdf is providing either enough functionality to be useful and given issues like what came up here (where it should have caught an error but didn't) doesn't give me much confidence.

tapastro:
Closing in favor of #327

PR #329, opened 2024-09-25
JP-2349: Add workaround to load old files with new moving_target_table schema

tapastro:
Resolves [JP-2349](https://jira.stsci.edu/browse/JP-2349)

This PR addresses validation issues encountered when opening `jwst` uncal files generated with DMS Ops builds <11.1 and using `stdatamodels`>=2.1.0. This most recent release of `stdatamodels` changed the number of columns in the `moving_target_table` schema, which leads to validation errors when opening any existing file with an old moving target extension. For any users attempting to load an outdated uncal file, this PR will check for the missing columns and add nan-filled columns if they are missing.

## Tasks

</details>

### DIFF
--- a/changes/329.feature.rst
+++ b/changes/329.feature.rst
@@ -1,0 +1,1 @@
+Add function to ``jwst.Level1bModel`` twhich fills moving target tables with missing columns
--- a/changes/329.feature.rst
+++ b/changes/329.feature.rst
@@ -1,1 +1,1 @@
-Add function to ``jwst.Level1bModel`` twhich fills moving target tables with missing columns
+Add function to ``jwst.Level1bModel`` which fills moving target tables with missing columns
--- a/src/stdatamodels/jwst/datamodels/level1b.py
+++ b/src/stdatamodels/jwst/datamodels/level1b.py
@@ -1,5 +1,4 @@
 from .model_base import JwstDataModel
-from astropy.io import fits
 import numpy as np
 from numpy.lib.recfunctions import merge_arrays
 
--- a/src/stdatamodels/jwst/datamodels/level1b.py
+++ b/src/stdatamodels/jwst/datamodels/level1b.py
@@ -53,8 +53,5 @@ class Level1bModel(JwstDataModel):
     """
     schema_url = "http://stsci.edu/schemas/jwst_datamodel/level1b.schema"
 
-    def __init__(self, init=None, **kwargs):
-        if isinstance(init, fits.HDUList):
-            init = _migrate_moving_target_table(init)
-
-        super(Level1bModel, self).__init__(init=init, **kwargs)
+    def _migrate_hdulist(self, hdulist):
+        return _migrate_moving_target_table(hdulist)
--- a/src/stdatamodels/jwst/datamodels/level1b.py
+++ b/src/stdatamodels/jwst/datamodels/level1b.py
@@ -1,7 +1,32 @@
 from .model_base import JwstDataModel
-
+from astropy.io import fits
+import numpy as np
+from numpy.lib.recfunctions import merge_arrays
 
 __all__ = ['Level1bModel']
+
+
+def _migrate_moving_target_table(hdulist):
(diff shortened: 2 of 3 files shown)


tapastro:
I'm running a set of regression tests for completeness: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1750/

But this made it through without a regression test failure, which would mean that we don't have a regression test that loads an uncal file with moving target data. This needs to be fixed in a yet-unwritten PR to `jwst` with a new test.

changes/329.feature.rst
```diff
@@ -0,0 +1 @@
+Add function to ``jwst.Level1bModel`` twhich fills moving target tables with missing columns
```
braingram, line 1:
```suggestion
Add function to ``jwst.Level1bModel`` which fills moving target tables with missing columns
```

src/stdatamodels/jwst/datamodels/level1b.py
```diff
@@ -27,3 +52,9 @@ class Level1bModel(JwstDataModel):
 
     """
     schema_url = "http://stsci.edu/schemas/jwst_datamodel/level1b.schema"
+
+    def __init__(self, init=None, **kwargs):
+        if isinstance(init, fits.HDUList):
+            init = _migrate_moving_target_table(init)
+
+        super(Level1bModel, self).__init__(init=init, **kwargs)
```
braingram, line 60: Similar to https://github.com/spacetelescope/stdatamodels/pull/183 would you add a call to `_migrate_moving_target_table` to the `_migrate_hdulist` method for this subclass? I think this should do it.
```suggestion
        super(Level1bModel, self).__init__(init=init, **kwargs)

    def _migrate_hdulist(self, hdulist):
        return _migrate_moving_target_table(hdulist)
```

braingram (commented):
Would you add a simple unit test for this that provides a table in the out-dated structure to verify it's updated? It can be largely copied from https://github.com/spacetelescope/stdatamodels/pull/183 (updated to the table structue for `MOVING_TARGET_POSITION`).

Thanks!

emolter:
>  @emolter modified the test's input file to include the new columns, which of course fixed the test but ignores the problems introduced for all other files.

Guess I learned the hard way that this is bad practice... sorry about that.  Should have thought it through more

src/stdatamodels/jwst/datamodels/level1b.py
tapastro: Happy to do that, but I didn't see it being called. Should the call in `__init__` be to `_migrate_hdulist` instead of `_migrate_moving_target_table`?

braingram: It's called during the `super` call when init is:
- a [hdulist](https://github.com/spacetelescope/stdatamodels/blob/234413f1209c50b9b103239279c813fa678dec81/src/stdatamodels/model_base.py#L217) this overlaps with what's currently in this PR (and what was in #183). I don't recall why it was added separately in the subclass `__init__` as the migration should be triggered in the `super` call for an hdulist.
- a [filename](https://github.com/spacetelescope/stdatamodels/blob/234413f1209c50b9b103239279c813fa678dec81/src/stdatamodels/model_base.py#L231) this is the main reason to add `_migrate_hdulist` since otherwise a call to `Level1bModel(filename)` won't trigger the migration.

Maybe the thing to do would be to:
- add a unit test
- drop the subclass `__init__`
- implement `_migrate_hdulist`

and see if that fixes the problem (or perhaps reveals why the previous PR also added the migration to the subclass).

braingram (approved):
LGTM. The CI failure is unrelated.

```
AssertionError: jwst_niriss_pastasoss_0001.asdf missing META.INSTRUMENT.EXP_TYPE required by jwst_niriss_pastasoss_0003.rmap
```

tapastro:
JWST regtests pass with this PR: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1755/

Merging!

PR #334, opened 2024-10-09
fix curl getting of context

braingram:
Similar to:
https://github.com/spacetelescope/crds/pull/1086

fixes the curl command to add the second required parameter.

## Tasks

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - id: crds_context
         run: >
           echo "pmap=$(
-          curl -s -X POST -d '{"jsonrpc": "1.0", "method": "get_default_context", "params": ["${{ env.OBSERVATORY }}"], "id": 1}' ${{ env.CRDS_SERVER_URL }}/json/ |
+          curl -s -X POST -d '{"jsonrpc": "1.0", "method": "get_default_context", "params": ["${{ env.OBSERVATORY }}", null], "id": 1}' ${{ env.CRDS_SERVER_URL }}/json/ |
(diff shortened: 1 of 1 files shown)


tapastro:
This looks good to me and I'm happy to merge, but out of curiosity I started a CI test. I didn't see any failures in past runs, and this run also passed: https://github.com/spacetelescope/stdatamodels/actions/runs/11264886442

I'm not sure I understand... are the downstream tests retrieving a context somehow?

braingram:
The setup here is a little different than jwst. A similar curl command is used to get the latest context but unlike jwst this repo doesn't set the `CRDS_CONTEXT` environment variable to that value (it probably should). The same as jwst the context is used to generate the cache "key" (the identifier github actions uses to store and retrieve cached files which is used to allow the tests to re-use an already fetched context). For the run you linked the key is `crds-None`: https://github.com/spacetelescope/stdatamodels/actions/runs/11264886442/job/31325721559#step:3:4
whereas for the run in this PR the key is `crds-jwst_1293.pmap`: https://github.com/spacetelescope/stdatamodels/actions/runs/11263496725/job/31321510542?pr=334#step:3:4
The caching helps with issues that can randomly pop up because of https://github.com/spacetelescope/crds/issues/930 and running the tests in parallel. If 1 test decides to start downloading a file and a second test starts that uses the same file the 2nd test will see a partial file and try to use it (leading to erroneous test failures). I think this can still happen with the current setup (when a new context is released) but I have not seen that type of failure recently.

tapastro (approved):

PR #336, opened 2024-10-16
PSF data model

jemorrison:
Supports [JP-251](https://jira.stsci.edu/browse/JP-251)

This PR adds a PSF reference data model to support PSF-based extraction for MIRS LRS FIXED slit data  in the JWST pipeline. 

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
@@ -38,11 +38,11 @@ allOf:
       fits_hdu: PSF
       default: 0.0
       ndim: 2
-      datatype: float32
+      datatype: float64
     wave:
       title: Wavelength image
       fits_hdu: WAVE
       default: 0
       ndim: 1
-      datatype: float32
+      datatype: float64
 


jemorrison:
@spacetelescope/stdatamodels-maintainers ready for review

jemorrison:
@braingram @emolter I could use a review of this code to get it into the pipeline to be used with the new reference file

braingram:
Is this in support of https://github.com/spacetelescope/jwst/pull/8967?
If so, would you help me find where this new model is used in that PR?
If not, is there an associated JWST PR?

Also, is there an example file?

melanieclarke:
> Is this in support of [spacetelescope/jwst#8967](https://github.com/spacetelescope/jwst/pull/8967)? If so, would you help me find where this new model is used in that PR? If not, is there an associated JWST PR?
> 
> Also, is there an example file?

That is the right PR for this change.  The new model is used here:
https://github.com/spacetelescope/jwst/blob/521fa4c0c6cb6b7cfcf7a07700d812accc67e60a/jwst/extract_1d/psf_profile.py#L92

Example file is on CRDS test:
https://jwst-crds-test.stsci.edu/browse/jwst_miri_psf_0001.fits

The PR for JP-251 has not yet been updated to match this format for the file though -- it currently expects an earlier version of the file, with a 1D wavelength array.

src/stdatamodels/jwst/datamodels/schemas/miri_lrs_psf.schema.yaml
```diff
@@ -0,0 +1,44 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/jwst_datamodel/miri_lrs_psf.schema"
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_exptype.schema
+- $ref: keyword_readpatt.schema
+- $ref: keyword_psubarray.schema
+- $ref: keyword_grating.schema
```
braingram, line 10: Should this be keyword_filter?
The example file on CRDS has:
`META.INSTRUMENT.FILTER	P750L`
but I don't see a defined `META.INSTRUMENT.GRATING`.

braingram:
> > Is this in support of [spacetelescope/jwst#8967](https://github.com/spacetelescope/jwst/pull/8967)? If so, would you help me find where this new model is used in that PR? If not, is there an associated JWST PR?
> > Also, is there an example file?
> 
> That is the right PR for this change. The new model is used here: https://github.com/spacetelescope/jwst/blob/521fa4c0c6cb6b7cfcf7a07700d812accc67e60a/jwst/extract_1d/psf_profile.py#L92
> 
> Example file is on CRDS test: https://jwst-crds-test.stsci.edu/browse/jwst_miri_psf_0001.fits
> 
> The PR for JP-251 has not yet been updated to match this format for the file though -- it currently expects an earlier version of the file, with a 1D wavelength array.

Thanks!

braingram:
It looks like the reference file is only on test at the moment. I suspect we will want to update `test_schema_against_crds.py` to include the new reference file datamodel/reftype (before this line if the reftype is 'psf'):
https://github.com/spacetelescope/stdatamodels/blob/cf3f85f854e20e85818273754b136fdf8d4d0251/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py#L155
If we make that update as part of this PR will it break this test? Also is there a way to run that test against `jwst-test` to confirm the update/model is working?

src/stdatamodels/jwst/datamodels/schemas/miri_lrs_psf.schema.yaml
jemorrison: Yes it probably should include filter.

jemorrison:
@melanieclarke  @braingram I would think that if the reference file is only in crds-test that if we update the test test_schema_against_crds.py it would break the system.
Can we update that test once this reference file is in the operation crds. This is not needed for the upcoming build so I would not want to break anything getting ready for that build.

jemorrison:
@melanieclarke Change the wavelength array back to 1 D

braingram:
> I will try to look at this today or tomorrow.

I tested this PR with `jwst-crds-test` and it fails due to the missing `reftype` mapping in the `test_schema_against_crds.py`. Adding the following allows it to pass with `jwst-crds-test` and `jwst-crds`.
```
+    'psf': dm.MiriLrsPsfModel,
     'psfmask': dm.PsfMaskModel,
```
@jemorrison would you update the above mentioned test with the change so that test won't start failing when ops is updated?

braingram:
> > I will try to look at this today or tomorrow.
> 
> I tested this PR with `jwst-crds-test` and it fails due to the missing `reftype` mapping in the `test_schema_against_crds.py`. Adding the following allows it to pass with `jwst-crds-test` and `jwst-crds`.
> 
> ```
> +    'psf': dm.MiriLrsPsfModel,
>      'psfmask': dm.PsfMaskModel,
> ```
> 
> @jemorrison would you update the above mentioned test with the change so that test won't start failing when ops is updated?

The `psf` reftype is now on ops. I added an ignore for this reftype in https://github.com/spacetelescope/stdatamodels/pull/371 that we can remove in this PR (along with the above mentioned changes).

braingram (changes requested):
Now that #371 is merged (with the "skip" for the `psf` reftype) when you update this PR would you remove the "skip" and update the test as mentioned in: https://github.com/spacetelescope/stdatamodels/pull/336#issuecomment-2539089676

Thanks!

melanieclarke:
I'm testing out what would need to be done to extend this work to NIRSpec, and I'm wondering if there is anything MIRI LRS specific about this model.  Would it make sense to name it something more general so we can use the same model for other modes?  Like maybe call it SlitPSFModel, include a dispersion axis, and rename the center_col keyword to cross_dispersion_center or something like that?

jemorrison:
@braingram I updated the name of the files and models. I added the new name to test_schema_against_crds.py.
I tried to 'pytest test_schema_against_crds.py '- but I got errors about astropy not being installed (and it is ) Do I need to do more than 'pytest test_schema_against_crds.py ' ?

braingram:
The test will be run as part of the CI (and also if you run pytest in your local checkout). I'm also puzzled by the astropy warning :-/ Maybe try a fresh install (or environment). Let me know if it's helpful to huddle to sort it out .

Would you remove these 3 lines:
https://github.com/spacetelescope/stdatamodels/blob/0d141e5360e04b83cc12b04f3997cd76c70d84a3/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py#L204-L206
with this PR we want to no longer skip testing `psf` reftypes.

src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py
```diff
@@ -163,6 +163,7 @@ def flatten(xs):
     'sirskernel': dm.SIRSKernelModel,
     'speckernel': dm.SpecKernelModel,
     'specprofile': dm.SpecProfileModel,
+    'specpsf': dm.SpecPsfModel,
```
braingram, line 166:
```suggestion
    'psf': dm.SpecPsfModel,
```

The reftype is `psf` on crds:
https://jwst-crds.stsci.edu/browse/jwst_miri_psf_0002.rmap
Is this being renamed to `specpsf`?

jemorrison:
@braingram  The test now seems to run ok in cl.
The datamodel was changed to specpsf. I misunderstood how the test was running and changed it there too, but that was wrong. The reftype is 'psf'

src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
```diff
@@ -0,0 +1,48 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/jwst_datamodel/specpsf.schema"
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_exptype.schema
+- $ref: keyword_readpatt.schema
+- $ref: keyword_psubarray.schema
+- $ref: keyword_filter.schema
```
braingram, line 10:
```suggestion
- $ref: keyword_readpatt.schema
- $ref: keyword_pexptype.schema
- $ref: keyword_psubarray.schema
- $ref: keyword_filter.schema
- $ref: keyword_pfilter.schema
```
Do the other keywords need the corresponding `P_` keywords (like `psubarray` has already)?

src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
```diff
@@ -0,0 +1,48 @@
...
+              type: number
+              fits_keyword: CENTROW
+- type: object
+  properties:
+    data:
+      title: The PSF image
+      fits_hdu: PSF
+      default: 0.0
+      ndim: 2
+      datatype: float32
```
braingram, line 41: The `PSF` and `WAVE` HDUs contain float64s for the example on crds:
https://jwst-crds.stsci.edu/browse/jwst_miri_psf_0001.fits
Is it an error that the current reference file has float64s or should these be float64?

jemorrison: Good catch I changed it to float64.

src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
jemorrison: ok. Since the only reference file we have now is MIRI LRS fix slit I think we can leave them out. NIRSpec might provide a spec psf reference file in the future so me might need to revisit this then.

braingram (approved):
LGTM.

emolter (approved):

PR #348, opened 2024-10-23
make IRS2Model inherit from ReferenceFileModel

### LINKED ISSUE #347: IRS2Model does not inherit from ReferenceFileModel
https://github.com/spacetelescope/stdatamodels/blob/503320f6325f7dd454dc2d7c177043ddf069c6f6/src/stdatamodels/jwst/datamodels/irs2.py#L7

This is described as a reference file but doesn't inherit from ReferenceFileModel.

braingram:
Closes #347

This changes the parent class for IRS2Model to ReferenceFileModel.

Testing locally I was able to (with this PR) open https://jwst-crds.stsci.edu/browse/jwst_nirspec_refpix_0017.fits and run "validate" without issue. Let me know if more tests (not covered by the CI and regtests) would be helpful.

Will need regression tests run (when they're working again). Opening for review in the meantime.

Regression tests all passed: https://github.com/spacetelescope/RegressionTests/actions/runs/11484201187/job/31961401428

## Tasks

</details>

### DIFF
--- a/changes/348.bugfix.rst
+++ b/changes/348.bugfix.rst
@@ -1,0 +1,1 @@
+Change IRS2Model parent class to ReferenceFileModel
--- a/src/stdatamodels/jwst/datamodels/irs2.py
+++ b/src/stdatamodels/jwst/datamodels/irs2.py
@@ -1,4 +1,4 @@
-from .model_base import ReferenceFileModel
+from .reference import ReferenceFileModel
 
 
 __all__ = ['IRS2Model']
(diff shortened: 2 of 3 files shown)


tapastro (approved):
Looks good! Strange that there was a mismatch on schema inheritance vs. python class inheritance and yet nothing appeared to complain - unless you caught this on an error?

braingram:
I caught this due to IRS2Model ending up in the keyword dictionary comparison (after ignoring all ReferenceFileModel subclasses).

I think it should be straightforward to add a unit test that checks that all reference file schema using models inherit from ReferenceFileModel.

As this PR is held up until regtests can be run I'll see if adding that unit test causes any other failures. If not we can sneak it into this PR or else it might make sense in a follow-up.

braingram:
Added unit test for inheritance in 730e710

braingram:
As all regtests passed and so did the added unit tests I'm considering this "good to go". Let me know if any follow-up is needed.

PR #361, opened 2024-11-12
add fits keywords to NRMModel schema

### LINKED ISSUE #359: NRMModel updates
Discussion originating on https://github.com/spacetelescope/stdatamodels/issues/342) with @braingram :

>I'm working on changing the AMI code to use values from the header of the NRM reference file rather than the hardcoded ones currently in [mask_definitions.py](https://github.com/spacetelescope/jwst/blob/main/jwst/ami/mask_definitions.py). When I load the reference file as an NRMModel, the keywords I need are in the extra_fits part of the tree, but instead of being [accessed like this](https://stdatamodels.readthedocs.io/en/latest/jwst/datamodels/models.html#extra-fits-keywords) they seem to be stored in nested lists and have to be accessed by their index [...] This whole issue may be irrelevant anyways; since these values will be important to the AMI code I was thinking it might make sense to have them be defined/required in the schema.

>The reference file keywords aren't listed in the keyword dictionary and are in different schemas in this package. Adding them to the schema is required to get them out of "extra_fits". Adding them will require picking an attribute name so feel free to suggest one (or we can just use the current variable names) in that issue.

We concluded the values we need should be added to the NRMModel schema and therefore the keyword dictionary. I will confirm with other AMI team members what we want to call the attribute names (if different from the current FITS keyword names).

braingram:
Closes #359

The NRM reference files contain FITS keywords useful to CAL.
https://jwst-crds.stsci.edu/browse/jwst_niriss_nrm_0001.fits

This PR adds those keywords to the schema so they are accessible via the datamodel interface:
- `model.flat_to_flat`
- `model.x_a1` etc...

Regression tests running at https://github.com/spacetelescope/RegressionTests/actions/runs/11802888443

## Tasks

</details>

### DIFF
--- a/changes/361.feature.rst
+++ b/changes/361.feature.rst
@@ -1,0 +1,1 @@
+Add NRM reference file keywords to schema.
--- a/src/stdatamodels/jwst/datamodels/schemas/nrm.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/nrm.schema.yaml
@@ -13,46 +13,46 @@ allOf:
       title: "[m] flat2flat hole size"
     x_a1:
       fits_keyword: NRM_X_A1
-      title: "[m] X coordinate (m) of NRM sub-aperture 1"
+      title: "[m] X coordinate of NRM sub-aperture 1"
     y_a1:
       fits_keyword: NRM_Y_A1
-      title: "[m] Y coordinate (m) of NRM sub-aperture 1"
+      title: "[m] Y coordinate of NRM sub-aperture 1"
     x_a2:
       fits_keyword: NRM_X_A2
-      title: "[m] X coordinate (m) of NRM sub-aperture 2"
+      title: "[m] X coordinate of NRM sub-aperture 2"
     y_a2:
       fits_keyword: NRM_Y_A2
-      title: "[m] Y coordinate (m) of NRM sub-aperture 2"
+      title: "[m] Y coordinate of NRM sub-aperture 2"
     x_a3:
       fits_keyword: NRM_X_A3
-      title: "[m] X coordinate (m) of NRM sub-aperture 3"
+      title: "[m] X coordinate of NRM sub-aperture 3"
     y_a3:
       fits_keyword: NRM_Y_A3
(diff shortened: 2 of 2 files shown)


braingram:
> I think this looks good! To make sure I understand, these will not be accessed via the `meta` tree but rather from the "top level" of the datamodel? I.e. `model.x_a1` rather than `model.meta.x_a1`?

Exactly! Feel free to give it a try with the current NRM reference file (and this branch) if you're curious.

rcooper295:
Thanks! I was able to try it with your branch and it looks good! I realized that there may be two additional keywords whose values are currently hardcoded but that we could read in from this model instead, but I need to confirm if they are actually the same as values we already have in the FITS header, and if so, what we should call the attributes. I should have an answer on that tomorrow.
(For more details, this is the message I sent the AMI team):

> mask_definitions.py contains:
> ```
> self.activeD = 6.559 * m  # webbpsf kwd DIAM  - not a 'circle including all holes'
> self.OD = 6.610645669291339 * m  # Full pupil file size, incl padding, webbpsf kwd PUPLDIAM
> ```
> while the NRM reference file currently contains:
> ```
> DIAM    =             6.559348 / Flat-to-flat distance across pupil in V3 axis  
> PUPLDIAM=             6.603464 / Full pupil file size, incl padding.            
> PUPL_CRC=             6.603464 / Circumscribing diameter for JWST primary
> ```      
> I'm not sure if `activeD` should be the same as `DIAM` (it is, at least up to the fourth decimal place) and if `OD` should be the same as `PUPLDIAM` (it isn't quite) but if these are referring to the same quantity we should probably make them consistent.

braingram:
I'll mark this as draft until we hear back about the extra keywords.

rcooper295:
I spoke to my AMI team members and chased down the source of these numbers (or what we should be using) and we'd like to add these two: 
```
 diameter:
      fits_keyword: DIAM
      title: "[m] Flat-to-flat distance across pupil in V3 axis"
pupil_circumscribed:
      fits_keyword: PUPL_CRC
      title: "[m] Circumscribing diameter for JWST primary"
```
Thanks for your patience!!

Issue #261, opened 2024-02-08
`fits-schema` metaschema only checks top level `fits_hdu` and `fits_keyword`


braingram:
The `fits-schema` metaschema used for several (but not all) of the datamodel schemas does not check `fits_hdu` or `fits_keyword` if they are present in the top-most portion of the schema.

For example, the following schema will correctly fail a call to `check_schema`:
```yaml
%YAML 1.1
---
$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
id: "http://stsci.edu/schemas/jwst_datamodel/test.schema"
type: object
fits_hdu:
  - "invalid list"
fits_keyword:
  - "invalid keyword"
```
```python
>>> asdf.schema.check_schema(schema)
ValidationError: ['invalid keyword'] is not of type 'string'

Failed validating 'type' in schema['allOf'][1]['properties']['fits_keyword']:
    {'type': 'string'}

On instance['fits_keyword']:
    ['invalid keyword']
```

However if the `fits_hdu` (or `fits_keyword`) is nested in the schema the `check_schema` will incorrectly pass:
```yaml
%YAML 1.1
---
$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
id: "http://stsci.edu/schemas/jwst_datamodel/test.schema"
type: object
properties:
  area_table:
    fits_hdu:
      - "invalid list"
    fits_keyword:
      - "invalid keyword"
```
```python
>>> asdf.schema.check_schema(schema)  # passes with no ValidationError
```

All uses of `fits_hdu` and `fits_keyword` that I'm able to find in `stdatamodels` is of the second type (so the metaschema provides no validation for any of these parts of the schemas). Note that the `fits_hdu` and `fits_keyword` handling by `stdatamodels` (mapping model properties and fits data) is unaffected by this metaschema issue. This is only a problem if the schemas are checked against their metaschemas (as only occurs if `check_schema` is called). We are not currently seeing any errors because `check_schema` is not currently called at runtime or during any tests.

braingram:
Metaschema removed in https://github.com/spacetelescope/stdatamodels/pull/378

PR #378, opened 2025-01-22
Remove fits metaschema

braingram:
The fits metaschema provides no validation for the current schemas (see https://github.com/spacetelescope/stdatamodels/issues/261). It also references an old version (1.0.0) of the asdf metaschema.

This PR updates the datamodels schemas to use the new asdf metaschema 1.1.0 instead of the fits metaschema and removes the now unused fits metaschema.

This impact of this change is minimal and restricted to pytest/CI runs for this package. The metaschema is only used when the schema is checked with [check_schema](https://asdf.readthedocs.io/en/latest/api/asdf.schema.check_schema.html#check-schema). `check_schema` validates a schema against the stated metaschema and [is not run in the pipeline](https://github.com/search?q=repo%3Aspacetelescope%2Fjwst%20check_schema&type=code) (as expected) and is [not run directly in this repository](https://github.com/search?q=repo%3Aspacetelescope%2Fstdatamodels%20check_schema&type=code). It is only run as [part of the pytest asdf plugin](https://github.com/asdf-format/asdf/blob/43e1947f01eecd5b770705f4d9794686a2532e9b/pytest_asdf/plugin.py#L134) which runs in the CI for this package.

Regtests running at: https://github.com/spacetelescope/RegressionTests/actions/runs/12917676424

## Tasks

</details>


braingram:
@melanieclarke I can't request you as a reviewer so I'll add a ping here. I'll try to sort out why I can't request you and fix that.

braingram:
The random CI failure for py312-xdist that popped up in the latest run is due to conflicting crds reference file downloads, crds not locking and verifying files: https://github.com/spacetelescope/crds/issues/930
Maybe for this package (since the CRDS usage is minimal outside of the jwst downstream tests) we could turn on the cache file lock without too much of a performance hit for the tests. However the downstream jwst tests would continue to randomly fail (like the CI failure here) unless we also enable the lock there (which would make the tests essentially run in 1 process). Either way the failure is unrelated to this PR.

braingram:
> This looks fine to me. Can you point me to the asdf 1.1.0 schema, for my education?

Thanks!

Here's 1.1.0: https://github.com/asdf-format/asdf-standard/blob/main/resources/schemas/stsci.edu/asdf/asdf-schema-1.1.0.yaml

tapastro:
Does the replacement of the unused fits metaschema with the asdf metaschema retain some functionality? Is it needed?

braingram:
> Does the replacement of the unused fits metaschema with the asdf metaschema retain some functionality? Is it needed?

To recap the huddle discussion.

- The metaschemas check the schemas and only do so during pytest/CI runs of stdatamodels. The changes here have no impact on datamodel validation.
- The current fits metaschema provides no additional functionality over asdf-1.0.0 (due to the lack of recursion of the property validators for `fits_keyword` and `fits_hdu`) so dropping it (and replacing it with the asdf metaschema) doesn't lose any functionality.
- asdf-1.1.0 vs asdf-1.0.0 added support for defining `float16` as a valid datatype (with respect to ASDF). There are no plans to use float16 and we'd probably need to not map these to FITS (as it doesn't support float16). We could stick with asdf 1.0.0 but since we're making the change now that 1.1.0 is available my vote is for the newer one.
- This PR doesn't add unit tests to validate `fits_hdu`/`fits_keyword` schema contents but this could be added in a follow up PR that could also address any potential issues uncovered (there's nothing validating the `fits_hdu`/`fits_keyword` schema keywords currently so it's possible there are differences between the structure in the current fits metaschema and the datamodel schema contents). Given that as far as we know all the datamodels work any bugs would be issues with the metaschema defined structure and addressed in the added unit tests.

tapastro (approved):
🧹 🗑️

PR #379, opened 2025-01-22
JP-3796: Store slit scales in slit datamodel

melanieclarke:
Partially resolves [JP-3796](https://jira.stsci.edu/browse/JP-3796)
Partially resolves [JP-1341](https://jira.stsci.edu/browse/JP-1341)

Adding a couple metadata values relevant to NIRSpec MOS data to the slit and spec datamodels.

The slit_xscale and slit_yscale attributes will hold the scale factor to convert from slit coordinates relative to the open MSA shutter area (DMS style) to coordinates relative to the full shutter pitch (PPS style).  These values will be derived from data in one of the NIRSpec WCS reference files (the MSA file), and are necessary to convert DMS-style slit coordinates, such as source_xpos and source_ypos, to sky coordinates (JP-1341).  Barshadow corrections also require an accurate slit_yscale value to scale correctly to large slits like NIRSpec MOS data in longslit mode (JP-3796).

The scale values are different for each MSA quadrant, so it would be helpful for users to also store the quadrant value for the slit.  The quadrant is already defined in the slit datamodel (without a FITS keyword) but is not defined at all in the spectral datamodel.  I added it to the spectral datamodel and added a FITS keyword for both.

Proposed new FITS keywords for SCI extensions in SlitModel or MultiSlitModel data or EXTRACT1D extensions in MultiSpecModel data are: MSAQUAD, SLTXSCL, SLTYSCL.  They should be populated only for NIRSpec MOS data, and should not need to be searchable in the archive.

## Tasks

</details>


melanieclarke:
Running initial regression tests here:
https://github.com/spacetelescope/RegressionTests/actions/runs/12918138451

emolter:
Code changes look good to me.  It looks like most of the ~100 regtest failures and 1 unit test failure were caused by mismatches in header info that are expected.  However, I'm not sure what to make of the fact that this changes data values in the `SB_ERROR` extension of some regtest data.  Why is that happening?

I didn't look too much in detail at why the unit test is failing, but is a small companion PR to change the test in jwst going to be necessary?

melanieclarke:
Sorry, it looks like I was not clear enough on this PR.  This is associated with JWST PR https://github.com/spacetelescope/jwst/pull/9085.  The changes needed to fix the JWST tests are in that PR, as well as the changes to read in and store the scales in the right places, and use them for barshadow correction in place of a hard-coded value.  That last part is why there are differences in the data values - the barshadow correction values are slightly different for all MOS and IFU data with this change.

emolter (approved):

tapastro (approved):

PR #382, opened 2025-01-28
MIRI LRS specwcs datamodel 

jemorrison:
CLOSING PR in favor of #393 

Helps to resolve[JP-3848](https://jira.stsci.edu/browse/JP-3848)

Closes #

This PR adds a datamodel for the reference file type, specwcs, for the LRS. It can be used for Fixed slit and slitless. 
The reference file for Fixed slit contains the V2, V3 corners of the slit. These values are not in the slitless reference file
and default to None when the reference files are read in using datamodels. 

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -13,7 +13,7 @@ allOf:
 - $ref: subarray.schema
 - type: object
   properties:
-    wave_table:
+    wavetable:
       title: Wavelengths and x, y locations of wavelengths
       fits_hdu: WAVETABLE
       datatype:
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -20,36 +20,47 @@ allOf:
       - name: x_center
         datatype: float32
         unit: pixels
+        title: x location of the center of a given wavelength element
       - name: y_center
         datatype: float32
         unit: pixels
+        title: y location of the center of a given wavelength element
       - name: wavelength
         datatype: float32
         unit: microns
+        title: central wavelength of the wavelength element at x_center, y_center
       - name: x0
         datatype: float32
         unit: pixels
+        title: x location of the upper left corner of a given wavelength element
       - name: y0
         datatype: float32
         unit: pixels
+        title: y location of the upper left corner of a given wavelength element
       - name: x1
         datatype: float32
         unit: pixels
+        title: x location of the upper right corner of a given wavelength element
       - name: y1
         datatype: float32
         unit: pixels
+        title: y location of the upper right corner of a given wavelength element
       - name: x2
         datatype: float32
         unit: pixels
+        title: x location of the lower right corner of a given wavelength element
       - name: y2
         datatype: float32
         unit: pixels
+        title: y location of the lower right corner of a given wavelength element
       - name: x3
         datatype: float32
         unit: pixels
+        title: x location of the lower left corner of a given wavelength element 
       - name: y3
         datatype: float32
         unit: pixels
+        title: y location of the lower left corner of a given wavelength element
 - type: object
   properties:
     meta:
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -1,0 +1,109 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
+id: "http://stsci.edu/schemas/jwst_datamodel/specwcs_miri_lrs.schema"
+title: MIRI LRS Spec Schema
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_pexptype.schema
+- $ref: keyword_exptype.schema
+- $ref: keyword_readpatt.schema
+- $ref: keyword_filter.schema
+- $ref: keyword_band.schema
+- $ref: subarray.schema
+- type: object
+  properties:
+    wave_table:
+      title: Wavelengths and x, y locations of wavelengths
+      fits_hdu: WAVETABLE
+      datatype:
+      - name: x_center
+        datatype: float32
+        unit: pixels
+      - name: y_center
+        datatype: float32
+        unit: pixels
+      - name: wavelength
+        datatype: float32
+        unit: microns
+      - name: x0
+        datatype: float32
+        unit: pixels
+      - name: y0
+        datatype: float32
+        unit: pixels
+      - name: x1
+        datatype: float32
+        unit: pixels
+      - name: y1
+        datatype: float32
+        unit: pixels
+      - name: x2
+        datatype: float32
+        unit: pixels
+      - name: y2
+        datatype: float32
+        unit: pixels
+      - name: x3
+        datatype: float32
+        unit: pixels
+      - name: y3
+        datatype: float32
+        unit: pixels
+- type: object
+  properties:
+    meta:
+      type: object
+      properties:
+        x_ref:
+          type: number
+          title: x coord of ref position of MIRIM_SLIT
+          default: pixels
+          fits_keyword: IMX
+        y_ref:
+          type: number
+          title: y coord of ref position of MIRIM_SLIT
+          default: pixels
+          fits_keyword: IMY
+        x_ref_slitless:
+          type: number
+          title: x coord of ref position of MIRIM_SLITLESS
+          default: pixels
+          fits_keyword: IMXSLTL
+        y_ref_slitless:
+          type: number
+          title: y coord of ref position of MIRIM_SLITLESS
+          default: pixels
+          fits_keyword: IMYSLTL
+        v2_vert1:
+          type: number
+          title: Slit vertex 1 in V2 frame
+          fits_keyword: V2VERT1
+        v2_vert2:
+          type: number
+          title: Slit vertex 2 in V2 frame
+          fits_keyword: V2VERT2
+        v2_vert3:
+          type: number
+          title: Slit vertex 3 in V2 frame
+          fits_keyword: V2VERT3
+        v2_vert4:
+          type: number
+          title: Slit vertex 4 in V2 frame
+          fits_keyword: V2VERT4
+        v3_vert1:
+          type: number
+          title: Slit vertex 1 in V3 frame
+          fits_keyword: V3VERT1
+        v3_vert2:
+          type: number
+          title: Slit vertex 2 in V3 frame
+          fits_keyword: V3VERT2
+        v3_vert3:
+          type: number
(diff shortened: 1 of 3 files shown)


jemorrison:
@drlaw1558 @melanieclarke 

Here is the first attempt at a data model for specwcs for MIRI slit data. This data model can be used for slit or slitless. 
It is rather simple. If it used on slitless data the v2, v3 corners of the slit do not exist in the reference file and default to be None.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,109 @@
...
+        unit: pixels
+      - name: y2
+        datatype: float32
+        unit: pixels
+      - name: x3
+        datatype: float32
+        unit: pixels
+      - name: y3
+        datatype: float32
+        unit: pixels
```
nden, line 52: It'd be helpful to add description (or title) for each of these fields. These can come as a minimum from the names in the FITS table.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -343,6 +343,43 @@ def to_fits(self):
...
+        slit vertex 3 in V3 frames
+    v3vert4 : float
+        slit vertex 4 in V3 frames
+    wavetable : numpy  2-D array
+        For each row in the slit hold  wavelength, and 
+        x center, ycenter, x and y box region cooresponding to the wavelength
+    """
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/specwcs_miri_lrs.schema"
+    reftype = "specwcs"
+
```
nden, line 382: If you want to be able to programmatically create a MIRILRSModel then you need to implement the `__init__` method.

A `populate_meta` method may be useful to make sure the required meta keywords are present.

Would yo consider renaming the model to `MiriLRSSpecwcs` to be consistent with the other simialr models?

jemorrison: @nden I will add __init__ and populate_meta. I will follow what the other models do in wcs_ref_models.py. I have never made a model like this. Could you explain the reason these methods are in the model. Is it to able to create 
a model in python - on the fly and not use a reference file. 

On the name I tried to name the model in a way that was consistent with the other models defined in wcs_ref_models.py.
All the models had the instrument capitalized and did not have Specwcs in the model name. Am I missing something ?

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
jemorrison: @nden The names in the fits file are exactly these names and no more information in fits file is provided. I will give a shot at a description and run that by Greg Sloan who I think made these reference files.

nden: Yes, you are right. The names are the same. I sent you a document with the description. I think it's good to add the descriptions because the meaning of the names is not obvious.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
nden: > Is it to able to create a model in python - on the fly and not use a reference file.

Yes, that's the only reason. If you think the model will be be initialized only by a file then it's not necessary to add an `__init__`.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
jemorrison: done

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
jemorrison: @nden I have a question about how I have set up the specwcs_miri_lrs.schema.
I put all the float values that are used - x, y location of the slit frame , v2 v3 vertices in the meta data of the schema.
I am wondering if I should pull that out of meta just have entries for each float value so I would access them 
model.x_ref instead of model.meta.x_ref. 
When I call __initi___ I could pass these values and in the populate_meta - it would just be things
like instrument and exp_type.
Is that more standard ?

jemorrison:
@braingram I am hoping you can help me. I have made a new datamodel MIRILrsModel (it is a specwcs model).
The model name is up for debate so I can change it to make it more standard, but my question is I set up a __init__ and
populate_meta method for the model. This is first time I have done it this way. I don't know how to call it and test it. 
I thought if I did this:
model2 = datamodels.MIRILrsModel()
The __init__ method would call populate_meta at least. 
But that does not seem to be the case.
Do you know how to test that _init_ or populate_meta is work.  I am just looking for a simple test I can run and then one where I set the value in the __initi__ step.

braingram:
> @braingram I am hoping you can help me. I have made a new datamodel MIRILrsModel (it is a specwcs model). The model name is up for debate so I can change it to make it more standard, but my question is I set up a **init** and populate_meta method for the model. This is first time I have done it this way. I don't know how to call it and test it. I thought if I did this: model2 = datamodels.MIRILrsModel() The **init** method would call populate_meta at least. But that does not seem to be the case. Do you know how to test that _init_ or populate_meta is work. I am just looking for a simple test I can run and then one where I set the value in the **initi** step.

The `__init__` and `populate_meta` look similar to the other models in `wcs_ref_models.py`. When I call `MIRILrsModel()` it does call `populate_meta`. Is this not the case for you?

braingram:
> @braingram I think I solved my problem with the class and populate_meta. Now I want to try and pass a wavetable to the _init_ method and see if it gets filled in correctly. Maybe this is over-kill, but since wavetable is a 2-D array and in the schema it is read in from the FITS extension WAVETABLE - how could I test passing this table in on initialization. Would I need to set up a class wavetable and fill in the values and then pass wavetable in when I define MIRILrsModel or something else ?

Thanks for the update.

Having the test (or tests) closely match the usage is jwst seems most useful. Is there a jwst PR using this new model? If it's opening a reference file from CRDS one option would be to construct a test that opens a "faked" reference file with the new datamodel. Here's a test that does something similar:
https://github.com/spacetelescope/stdatamodels/blob/f7587b7f80f54c011779c7dfaff1039cd9c8032b/src/stdatamodels/jwst/datamodels/tests/test_open.py#L87

I assume one other use will be for defining a reference file. In that case providing an array to the constructor, then saving the model, then verifying the contents seems like a good test to consider.

Finally if this is a reference model mapped to a CRDS reference file the `test_schema_against_crds` file could use an update (to use the new reference model):
https://github.com/spacetelescope/stdatamodels/blob/f7587b7f80f54c011779c7dfaff1039cd9c8032b/src/stdatamodels/jwst/datamodels/tests/test_schema_against_crds.py#L170

jemorrison:
@braingram I just pushed up  changes.
This is what I am trying to do - maybe it is not needed. I wrote this python script to just see if I could set the MIRI LRS specwcs model up rather than using a reference file in CRDS. 
from stdatamodels.jwst import datamodels
import jwst

class wavetable_test:
    x_center = np.array([1,2,3])
    y_center= np.array([4,5,6])
    wavelength = np.array([12.4,13.4, 14.5])
    x0 = np.array([1,2,3])
    x1 = np.array([1,2,3])
    x2 = np.array([1,2,3])
    x3 = np.array([1,2,3])
    y0 = np.array([1,2,3])
    y1 = np.array([1,2,3])
    y2 = np.array([1,2,3])
    y3 = np.array([1,2,3])
print(wavetable_test.x_center)
model = datamodels.MiriLRSSpecwcsModel(x_ref=430, y_ref=400, wavetable=wavetable_test)



I get an error when I run this:
File "/Users/morrison/Pipeline/testing/MIRI_LRS_3848/test_specwcs_datamodel.py", line 20, in <module>
    model = datamodels.MiriLRSSpecwcsModel(x_ref=430, y_ref=400, wavetable=wavetable_test)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/morrison/git/stdatamodels/src/stdatamodels/jwst/datamodels/wcs_ref_models.py", line 404, in __init__
    self.wavetable = wavetable
    ^^^^^^^^^^^^^^
  File "/Users/morrison/git/stdatamodels/src/stdatamodels/model_base.py", line 692, in __setattr__
    properties.ObjectNode.__setattr__(self, attr, value)
  File "/Users/morrison/git/stdatamodels/src/stdatamodels/properties.py", line 331, in __setattr__
    val = _cast(val, schema)
          ^^^^^^^^^^^^^^^^^^
  File "/Users/morrison/git/stdatamodels/src/stdatamodels/properties.py", line 47, in _cast
    if (_is_struct_array_schema(schema) and len(val) and
                                            ^^^^^^^^
TypeError: object of type 'type' has no len()

jemorrison:
@braingram 
I have such a mess now with the main getting the style checks updated. I can not figure out the conflicts. I think I will close this PR - and open another clean on and then add my changes to the new PR

braingram:
Thanks for sharing the example. That error is caused by the schema expecting a [structured array](https://numpy.org/doc/stable/user/basics.rec.html) for the `wavetable`. Let me know if more information would be helpful.

> @braingram
I have such a mess now with the main getting the style checks updated. I can not figure out the conflicts. I think I will close this PR - and open another clean on and then add my changes to the new PR

Sounds good. This repository should hopefully be done with major style changes (and now aligns with the in-progress changes in jwst). These can certainly be disruptive to ongoing work.

jemorrison:
@braingram This reference file is an existing retype - specwcs. So I don't think I need to add it to testing_againist_crds_schema. I added this model to the wcs_ref_models.py 
If I have misunderstood your question please let me know

braingram:
> @braingram This reference file is an existing retype - specwcs. So I don't think I need to add it to testing_againist_crds_schema. I added this model to the wcs_ref_models.py If I have misunderstood your question please let me know

Is the new `MiriLRSSpecwcsModel` model going to be used to open the `specwcs` reference file for MIRI LRS reference files instead of `SpecwcsModel`? If so than the test needs to be updated to use the new reference model for reference files for MIRI LRS (and use `SpecwcsModel` for other modes).

jemorrison:
@braingram Sorry I am not tracking.
Looking at the other specwcs reference models: NIRCAMGrismModel, NIRISSGrismModel. 
These models are not defined in  test_schema_against_cdrs.py
I looked at how they are used in the pipeline:
with NIRCAMGrismModel(reference_files['specwcs'])
with NIRCAMGrismModel(reference_files['specwcs'])

This is how I am using the MIRILRS one:
refmodel = MiriLRSSpecwcsModel(reference_files['specwcs'])

So do I need it in test_schema_against_crds.py ?

jemorrison:
@braingram I added a unit test for this new model. I wrote this to test it was working as expected - I am not sure if it is what you want for a unit test. 
Let me know if I need to change it

jemorrison:
Sorry again
This is obviously wrong because it fails. But what should it be
it seems the format is ref_type: model

 "specwcs": dm.SpecwcsModel, dm.MiriLRSSpecwcsModel,
 this does not work either:
   "specwcs": dm.SpecwcsModel,
    "specwcs": dm.MiriLRSSpecwcsModel,

PR #383, opened 2025-01-29
Align style checks with jwst

braingram:
https://github.com/spacetelescope/jwst/pull/9081 introduced major changes to the code style checks in jwst.

Since this package primarily serves jwst this PR updates the style checks here to more closely match those in jwst.

~Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/13039378253
show 1 error `test_exec_time_0_crs` which I'm chalking up to a slower-than-usual runner which took 11 instead of 10 seconds. I expect these may need to be re-run after review so I'll open this PR for review now.~
EDIT: see updated regtests comment: https://github.com/spacetelescope/stdatamodels/pull/383#issuecomment-2628321965

Branch protections will need to be updated after this PR is merged to check for pre-commit success.

## Tasks

</details>


.ruff.toml
```diff
@@ -0,0 +1,73 @@
...
+    "YTT",    # flake8-2020 (prevent some specific gotchas from sys.version)
+]
+ignore = [
+    "D100", # missing docstring in public module
+    "E741", # ambiguous variable name (O/0, l/I, etc.)
+    "UP008", # use super() instead of super(class, self). no harm being explicit
+    "UP015", # unnecessary open(file, "r"). no harm being explicit
+    "TRY003", # prevents custom exception messages not defined in exception itself.
+    "ISC001", # single line implicit string concatenation. formatter recommends ignoring this.
+    "PTH123", # use Path.open instead of open
```
braingram, line 54: There are lots of `open` uses in the code. I don't see why `Path(filename).open` is preferred.

.ruff.toml
```diff
@@ -0,0 +1,73 @@
...
+    "UP038", # isinstance with | instead of ,
```
braingram, line 55: As far as I'm aware `isinstance(foo, FooType | BarType)` and `isinstance(foo, (FooType, BarType))` are equivalent and the latter is faster.

.ruff.toml
```diff
@@ -0,0 +1,73 @@
...
+    # longer term fix
+    "B028", # warnings need stack level, this is non-trivial to determine
+    "B904", # raise ... from ?, when and where do we want to not use the default chain?
+    "NPY002",  # numpy rng, will require finding the right replacements
+    "S101",  # asserts are used in many non-test places
+    "SLF001",  # private member access, this is overly restrictive
```
braingram, line 61: I put this in the "fix later" category but I'm not convinced it's worth the effort for the number of what I would consider "false positives".

.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - id: detect-private-key
+      # - id: end-of-file-fixer
+      # - id: trailing-whitespace
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: 'v0.9.2'
+    hooks:
+      - id: ruff
+        args: ["--fix"]
+      - id: ruff-format
+  # - repo: https://github.com/numpy/numpydoc
```
braingram, line 25: Even after excluding `test.*` items this shows 2884 failures. My vote is to enable and attempt to address these in a follow up PR.

emolter: Sounds fine to me

src/stdatamodels/jwst/datamodels/tests/test_integration.py
```diff
@@ -101,16 +97,16 @@ def test_schema_refs_base(datamodel_schema_file):
...
-        assert 'http://stsci.edu/schemas/jwst_datamodel/referencefile.schema' not in seen_ids
-    elif 'http://stsci.edu/schemas/jwst_datamodel/referencefile.schema' in seen_ids:
-        assert 'http://stsci.edu/schemas/jwst_datamodel/core.schema' not in seen_ids
+    if "http://stsci.edu/schemas/jwst_datamodel/core.schema" in seen_ids:
+        assert "http://stsci.edu/schemas/jwst_datamodel/referencefile.schema" not in seen_ids
+    elif "http://stsci.edu/schemas/jwst_datamodel/referencefile.schema" in seen_ids:
+        assert "http://stsci.edu/schemas/jwst_datamodel/core.schema" not in seen_ids
     else:
-        assert False
+        assert False  # noqa: B011
```
emolter, line 112: Could this be replaced with a `raise AssertionError` as the rule suggests?

Or I guess it's fine to wait until the "fix later" assert removes happen, so nevermind

src/stdatamodels/jwst/datamodels/tests/test_multislit.py
```diff
@@ -105,27 +107,27 @@ def test_multislit_metadata():
...
         assert len(output.slits) == 4
 
     with fits.open(path, memmap=False) as hdulist:
         assert len(hdulist) == 6
 
     with MultiSlitModel(path) as model:
-        for i, slit in enumerate(model.slits):
+        for i, _ in enumerate(model.slits):  # noqa: B007
             pass
         assert i + 1 == 4
```
emolter, line 132: it isn't immediately clear to me what these lines are supposed to be doing, but I assume there is a reason for this?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -435,10 +508,10 @@ def validate(self):
             if self._strict_validation:
                 raise
             else:
-                warnings.warn(traceback.format_exc(), ValidationWarning)
+                warnings.warn(traceback.format_exc(), ValidationWarning, stacklevel=2)
 
-    def get_wfss_wavelength_range(self, filter, orders):
-        """ Retrieve the wavelength range for a WFSS observation.
+    def get_wfss_wavelength_range(self, filter, orders):  # noqa: A002
```
emolter, line 513: Can we replace the variable name filter with something else?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -843,7 +960,7 @@ def __init__(self, init=None, apertures=None, **kwargs):
 
     @property
     def aperture_names(self):
-        return [getattr(ap, 'aperture_name') for ap in self.apertures]
+        return [getattr(ap, "aperture_name") for ap in self.apertures]  # noqa: B009
```
emolter, line 963: why ignore instead of fixing?

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -106,69 +162,62 @@ class GrismObject(namedtuple('GrismObject', ("sid",
...
-                                               sky_bbox_ur=sky_bbox_ur,
-                                               sky_bbox_ul=sky_bbox_ul,
-                                               xcentroid=xcentroid,
-                                               ycentroid=ycentroid,
-                                               is_extended=is_extended,
-                                               isophotal_abmag=isophotal_abmag)
+    def __new__(
+        cls,
+        sid=None,
+        order_bounding={},  # noqa: B006
```
emolter, line 171: looks to me like these could be easily fixed by setting the default to `None` as suggested by the rule

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -672,34 +737,36 @@ def apply_poly(coeff_model, inputs, t):
...
+                    self.xmodels[order][0](x0, y0)
+                    + t0 * self.xmodels[order][1](x0, y0)
+                    + t0**2 * self.xmodels[order][2](x0, y0)
+                )
             elif len(self.xmodels[order][0].inputs) == 1:
                 xr = (dx - self.xmodels[order][0].c0.value) / self.xmodels[order][0].c1.value
                 return xr
             else:
-                raise Exception
+                raise Exception  # noqa: TRY002
```
emolter, line 769: Can this be made more specific instead of bare Exception?

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -820,24 +897,26 @@ def apply_poly(coeff_model, inputs, t):
...
+                    model[order][0](x0, y0)
+                    + t0 * model[order][1](x0, y0)
+                    + t0**2 * model[order][2](x0, y0)
+                )
             elif len(model[order][0].inputs) == 1:
                 xr = (dy - model[order][0].c0.value) / model[order][0].c1.value
                 return xr
             else:
-                raise Exception
+                raise Exception  # noqa: TRY002
```
emolter, line 919: same here

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1435,48 +1521,59 @@ def compute_refraction_index(lam, temp, tref, pref, pressure, kcoef, lcoef, tcoe
 
         # Convert to microns
         lam = np.asarray(lam * 1e6)
-        KtoC = 273.15  # kelvin to celsius conversion
+        KtoC = 273.15  # kelvin to celsius conversion  # noqa: N806
```
emolter, line 1524: can this be replaced by `k_to_c` or similar here?

src/stdatamodels/model_base.py
```diff
@@ -533,48 +540,40 @@ def save(self, path, dir_path=None, *args, **kwargs):
...
             path_head, path_tail = os.path.split(path)
-        base, ext = os.path.splitext(path_tail)
+        ext = Path(path_tail).suffix
         if isinstance(ext, bytes):
             ext = ext.decode(sys.getfilesystemencoding())
 
         if dir_path:
             path_head = dir_path
-        output_path = os.path.join(path_head, path_tail)
+        output_path = os.path.join(path_head, path_tail)  # noqa: PTH118
```
emolter, line 549: same here, can os.path instances be fixed?

tests/test_models.py
```diff
@@ -281,29 +282,27 @@ def on_save(self, init):
...
     assert "foo" not in model.meta._instance
-    model.save(tmp_path/"test.asdf")
+    model.save(tmp_path / "test.asdf")
     assert model.meta.foo == "bar"
 
 
-@pytest.mark.parametrize("ModelType", [DataModel, BasicModel, TableModel, TransformModel])
-def test_garbage_collectable(ModelType, tmp_path):
+@pytest.mark.parametrize("ModelType", [DataModel, BasicModel, TableModel, TransformModel])  # noqa: N803
+def test_garbage_collectable(ModelType, tmp_path):  # noqa: N803
```
emolter, line 290: can this be changed to `model_type`?

emolter (commented):
Thanks very much for putting this together, Brett!

The changes to the CI and pre-commit look good to me.  I had some suggestions for following up on this PR, but none of them would have to be handled now.

Most of my comments are related to places where I think it wouldn't be overly disruptive to fix the errors instead of `noqa: ignore`.

src/stdatamodels/filetype.py
```diff
@@ -15,24 +15,24 @@ def check(init):
     file_type: a string with the file type ("asdf", "asn", or "fits")
 
     """
-    supported = ('asdf', 'fits', 'json')
+    supported = ("asdf", "fits", "json")
 
     if isinstance(init, str):
-        path, ext = os.path.splitext(init)
-        ext = ext.strip('.')
+        path, ext = os.path.splitext(init)  # noqa: PTH122
```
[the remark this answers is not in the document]
braingram: I'm happy to open a separate issue for this.

The general suggestion would be to replace this with `Path.suffix` (since all we want is the final suffix here). However the code below has a second block for further stripping suffixes. We could update that to also use `Path.suffix`. I'd argue that if we're supposed to adopt `Path` that `Path.suffixes` is the cleaner choice here but would require more code refactoring. I think this makes sense in a separate PR.

src/stdatamodels/jwst/_kwtool/kwd.py
```diff
@@ -29,10 +29,10 @@ def _walk_kwd(kwd_path, root, path=None, keywords=None):
     if keywords is None:
         keywords = []
     if path is None:
-        path = tuple()
+        path = ()
     if not isinstance(root, (dict, list)):
         if not isinstance(root, (int, float, str)):
-            raise ValueError(f"Unknown type {type(root)} in keyword dictionary")
+            raise ValueError(f"Unknown type {type(root)} in keyword dictionary")  # noqa: TRY004
```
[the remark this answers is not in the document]
braingram: That would change the type of the error so anything catching `ValueError` would no longer work. I can open an issue to track this but I'd say changing the exception type is not a style change (and worth a separate PR).

src/stdatamodels/jwst/datamodels/tests/test_integration.py
braingram: I think `assert False` is more readable. Perhaps `B011` should be disabled for tests since the description of [why it's bad](https://docs.astral.sh/ruff/rules/assert-false/) doesn't apply to tests:
> Python removes assert statements when running in optimized mode (python -O), making assert False an unreliable means of raising an AssertionError.

src/stdatamodels/jwst/datamodels/tests/test_multislit.py
braingram: I see this as a bug in the rule, maybe we should disable it.

`i` is in scope and used on 132. Having a rule that suggests removing it would break the code.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
braingram: I think that's worth a different PR if we want to address it. I think `filter` is readable here and given the context is obvious to the developer and user that it's not the builtin `filter`.

tests/test_models.py
braingram: Since `ModelType` here is a class so I'm in favor of keeping the CamelCase.

src/stdatamodels/model_base.py
braingram: `output_path` is returned below (as a string). Rather than convert to a `Path` to do the join then back to a string to return I think it's simpler and more readble to keep it as a string.

src/stdatamodels/jwst/transforms/models.py
braingram: It could. However this line shows no test coverage. I'd rather err on the side of caution of add the noqa. New code can follow the convention but I'd like to see test coverage first (which is beyond the scope of this PR).

src/stdatamodels/jwst/transforms/models.py
braingram: Similar to https://github.com/spacetelescope/stdatamodels/pull/383#discussion_r1937770506 this shows no test coverage. We could add a new exception class and use it here but I'd like to see tests first.

src/stdatamodels/jwst/transforms/models.py
braingram: Same as https://github.com/spacetelescope/stdatamodels/pull/383#discussion_r1937771820

src/stdatamodels/dynamicdq.py
```diff
@@ -32,23 +33,22 @@ def dynamic_mask(input_model, mnemonic_map, inv=False):
 
     dq_table = input_model.dq_def
     # Get the DQ array and the flag definitions
-    if (dq_table is not None and
-        not np.isscalar(dq_table) and
-        len(dq_table.shape) and
-            len(dq_table)):
+    if dq_table is not None and not np.isscalar(dq_table) and len(dq_table.shape) and len(dq_table):
```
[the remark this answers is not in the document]
braingram: Is there a rule/style configuration that would enforce this?

.ruff.toml
```diff
@@ -0,0 +1,70 @@
...
+    "__pycache__",
+    "docs",
+    ".eggs",
+    "build",
+    "dist",
+    ".tox",
+    ".eggs",
+]
+line-length = 100
+target-version = "py310"
```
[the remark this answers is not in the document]
braingram: Currently stdatamodels supports 3.10:
https://github.com/spacetelescope/stdatamodels/blob/b49f5f437e49322c5fcc48bd47052779c6a066de/pyproject.toml#L4
Since jwst just dropped support today we can drop 3.10 but I'd say let's leave that to a different PR.

.github/workflows/ci.yml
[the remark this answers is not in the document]
braingram: xref: https://github.com/spacetelescope/stdatamodels/issues/386

src/stdatamodels/dynamicdq.py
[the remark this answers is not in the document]
emolter: I don't think so, but it just looked like the style fix made it harder to read.  I personally prefer the parentheses.  NBD though either way

src/stdatamodels/jwst/datamodels/tests/test_integration.py
emolter: I'm fine with disabling it for tests.  TBH I think I didn't realize I was in a test when I made the comment

src/stdatamodels/jwst/datamodels/tests/test_multislit.py
emolter: So far this rule has caught several of what I would consider "real" bad style in the various JWST PRs, so I'm in favor of keeping it.

You have to admit this is an unusual case, where there's nothing at all inside the `for` loop.  I'd suggest leaving it as-is

src/stdatamodels/jwst/transforms/models.py
emolter: This feels to me like a pretty safe one, since any exception would inherit from `Except`.

src/stdatamodels/jwst/datamodels/tests/test_integration.py
braingram: disabled in 4090a783ae3eea9c6ec859b30f74ca0b8ad62a00

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
emolter: I'm not sure why this would be put in a different PR, if it's being suggested by the style rules and this PR is fixing things that violate the style rules.

Yes, `filter` is readable, but so is, say, `filter_name` or `filt`.  I don't think the change would hurt anything.

src/stdatamodels/filetype.py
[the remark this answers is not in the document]
emolter: Fine with me if a separate issue is opened.

src/stdatamodels/jwst/datamodels/tests/test_multislit.py
braingram: Thanks! Yeah I say we keep the rule enabled.

src/stdatamodels/jwst/_kwtool/kwd.py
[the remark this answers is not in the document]
emolter: Fair enough RE the lack of tests.  We are changing things like this in JWST with the assumption that we have decent test coverage (although it isn't always true).

For this one and the few others where I agree the fix might be too risky, what would the issue look like?  Something like, "Add more unit tests to submodule X, and remove noqa: ignores where possible"?  Not really sure what the best thing would be there.

src/stdatamodels/jwst/transforms/models.py
emolter: Sounds good, see my other comment about what issue(s) to open to ensure this gets tracked

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
braingram: I guess I see function signatures as more public than variable names.

I changed it to `filter_name` in 28050746164a1296c32edbe2f18c5bd3f9ad090f

src/stdatamodels/jwst/transforms/models.py
braingram: What do you suggest we name the new exception and add to it's docstring?

src/stdatamodels/filetype.py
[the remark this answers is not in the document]
braingram: https://github.com/spacetelescope/stdatamodels/issues/387

src/stdatamodels/jwst/_kwtool/kwd.py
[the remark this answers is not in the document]
braingram: Thanks. Test coverage would help. Changing an exception type that might be handled by a downstream package (or other parts of the same package) would be worth a deprecation period with the old and new exceptions before switching to the new one.

src/stdatamodels/jwst/transforms/models.py
braingram: How does https://github.com/spacetelescope/stdatamodels/issues/388 look?

src/stdatamodels/jwst/transforms/models.py
emolter: Maybe something like `InvalidTransformError(f"xmodels for order {order} has invalid length {len(self.xmodels)}")`

But I see your point that we might want to run that past, say, Nadia or someone more expert with the code to ensure that name and docstring seem correct to her... and in that case it would make more sense as a separate PR.

src/stdatamodels/jwst/transforms/models.py
emolter: Thanks, sounds great to me

src/stdatamodels/jwst/transforms/models.py
emolter: Since this is also covered by the issue you just opened, I'd say we leave it as-is here and ignore my original comment.  Feel free to resolve if you agree

src/stdatamodels/jwst/_kwtool/kwd.py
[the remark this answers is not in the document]
emolter: That's fair

src/stdatamodels/jwst/transforms/models.py
braingram: Thanks for the suggestion. I opened https://github.com/spacetelescope/stdatamodels/issues/389 to track this.

src/stdatamodels/jwst/transforms/models.py
braingram: Maybe but that's a change in behavior. Using a modifiable argument allows things like caching values. It's unclear to me at a cursory glance why this is `{}` and others are `None`. Without digging further into it I'm not sure if that's what's happening here. This line is also uncovered. I'd rather go with the safer bet of adding a `noqa` than changing the behavior.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
braingram: This line isn't covered by any tests. I chose not to change the code since I'm not if the change would be tested.

src/stdatamodels/jwst/transforms/models.py
emolter: Ok. Happy to include this in fixes associated with https://github.com/spacetelescope/stdatamodels/issues/388

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
emolter: should I open another issue requesting the addition of unit tests for `wcs_ref_models.py`?

braingram: Thanks. Yeah, we might also consider a catch all issue that just says "add coverage for noqa lines". What do you think?

.pre-commit-config.yaml
braingram: Issue opened to track this: https://github.com/spacetelescope/stdatamodels/issues/385

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
emolter: I think that would be good, something like "Add unsafe style fixes, with test coverage"

emolter (commented):
I'm happy with this pending passing downstream JWST unit tests.  Probably wouldn't hurt to re-run the regression tests too out of an abundance of caution, since you accepted at least one of the "unsafe" fixes I suggested.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
braingram: How does https://github.com/spacetelescope/stdatamodels/issues/390 look?

braingram:
Rerunning regtests here: https://github.com/spacetelescope/RegressionTests/actions/runs/13039378253

braingram:
> Rerunning regtests here: https://github.com/spacetelescope/RegressionTests/actions/runs/13039378253

Regtests show only unrelated errors.

emolter (approved):
Code changes look good to me now, or at least, we have separate issues tracking additional changes.

Why is the Python 3.10 test failing now?  I don't see how it was modified by the changes here.

braingram:
The 3.10 test failure:
https://github.com/spacetelescope/stdatamodels/actions/runs/13115248313/job/36587914704?pr=383#step:10:282
is one incarnation of the "random" one where depending on the test order (and how CRDS files are downloaded) tests can fail when >1 test downloads the same CRDS file.

PR #393, opened 2025-02-03
Add miri lrs spec wcs model

jemorrison:
Helps Resolve [JP-3848](https://jira.stsci.edu/browse/JP-3848)

This PR replaces #382. There we too many conflicts to resolve easily after the code style changes were made. 

Closes #

Corresponding jwst PR: https://github.com/spacetelescope/jwst/pull/9193

This PR adds a datamodel of the reference file, specwcs, for MIRI LRS data. It can be used for fixed slit or slitless data. The reference file for fixed slit contains the V2,V3 corners of the slit. These values are not in the slitless reference file and default to None when the reference files are read in using the data models

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -2,14 +2,12 @@
 ---
 $schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
 id: "http://stsci.edu/schemas/jwst_datamodel/specwcs_miri_lrs.schema"
-title: MIRI LRS Spec Schema
+title: MIRI LRS Fixed Slit Specwcs Schema
 allOf:
 - $ref: referencefile.schema
 - $ref: keyword_pexptype.schema
 - $ref: keyword_exptype.schema
-- $ref: keyword_readpatt.schema
 - $ref: keyword_filter.schema
-- $ref: keyword_band.schema
 - $ref: subarray.schema
 - type: object
   properties:
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -67,22 +67,22 @@ allOf:
         x_ref:
           type: number
           title: x coord of ref position of MIRIM_SLIT
-          default: pixels
+          units: pixels
           fits_keyword: IMX
         y_ref:
           type: number
           title: y coord of ref position of MIRIM_SLIT
-          default: pixels
+          units: pixels
           fits_keyword: IMY
         x_ref_slitless:
           type: number
           title: x coord of ref position of MIRIM_SLITLESS
-          default: pixels
+          units: pixels
           fits_keyword: IMXSLTL
         y_ref_slitless:
           type: number
           title: y coord of ref position of MIRIM_SLITLESS
-          default: pixels
+          units: pixels
           fits_keyword: IMYSLTL
         v2_vert1:
           type: number
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -17,47 +17,36 @@ allOf:
       datatype:
       - name: x_center
         datatype: float32
-        unit: pixels
         title: "[pixels] x location of the center of a given wavelength element"
       - name: y_center
         datatype: float32
-        unit: pixels
         title: "[pixels] y location of the center of a given wavelength element"
       - name: wavelength
         datatype: float32
-        unit: microns
         title: "[microns] central wavelength of the wavelength element at x_center, y_center"
       - name: x0
         datatype: float32
-        unit: pixels
         title: "[pixels] x location of the upper left corner of a given wavelength element"
       - name: y0
         datatype: float32
-        unit: pixels
         title: "[pixels] y location of the upper left corner of a given wavelength element"
       - name: x1
         datatype: float32
-        unit: pixels
         title: "[pixels] x location of the upper right corner of a given wavelength element"
       - name: y1
         datatype: float32
-        unit: pixels
         title: "[pixels] y location of the upper right corner of a given wavelength element"
       - name: x2
         datatype: float32
-        unit: pixels
         title: "[pixels] x location of the lower right corner of a given wavelength element"
       - name: y2
         datatype: float32
-        unit: pixels
         title: "[pixels] y location of the lower right corner of a given wavelength element"
       - name: x3
         datatype: float32
-        unit: pixels
         title: "[pixels] x location of the lower left corner of a given wavelength element"
       - name: y3
         datatype: float32
-        unit: pixels
         title: "[pixels] y location of the lower left corner of a given wavelength element"
 - type: object
   properties:
@@ -66,7 +55,7 @@ allOf:
       properties:
         x_ref:
           type: number
-          title: "[pixels]  x coord of ref position of MIRIM_SLIT
+          title: "[pixels]  x coord of ref position of MIRIM_SLIT"
           fits_keyword: IMX
         y_ref:
           type: number
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -55,7 +55,7 @@ allOf:
       properties:
         x_ref:
           type: number
-          title: "[pixels]  x coord of ref position of MIRIM_SLIT"
+          title: "[pixels] x coord of ref position of MIRIM_SLIT"
           fits_keyword: IMX
         y_ref:
           type: number
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -1,0 +1,120 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
+id: "http://stsci.edu/schemas/jwst_datamodel/specwcs_miri_lrs.schema"
+title: MIRI LRS Spec Schema
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_pexptype.schema
+- $ref: keyword_exptype.schema
+- $ref: keyword_readpatt.schema
+- $ref: keyword_filter.schema
+- $ref: keyword_band.schema
+- $ref: subarray.schema
+- type: object
+  properties:
+    wavetable:
+      title: Wavelengths and x, y locations of wavelengths
+      fits_hdu: WAVETABLE
+      datatype:
+      - name: x_center
+        datatype: float32
+        unit: pixels
+        title: x location of the center of a given wavelength element
+      - name: y_center
+        datatype: float32
+        unit: pixels
+        title: y location of the center of a given wavelength element
+      - name: wavelength
+        datatype: float32
+        unit: microns
+        title: central wavelength of the wavelength element at x_center, y_center
+      - name: x0
+        datatype: float32
+        unit: pixels
+        title: x location of the upper left corner of a given wavelength element
+      - name: y0
+        datatype: float32
+        unit: pixels
+        title: y location of the upper left corner of a given wavelength element
+      - name: x1
+        datatype: float32
+        unit: pixels
+        title: x location of the upper right corner of a given wavelength element
+      - name: y1
+        datatype: float32
+        unit: pixels
+        title: y location of the upper right corner of a given wavelength element
+      - name: x2
+        datatype: float32
+        unit: pixels
+        title: x location of the lower right corner of a given wavelength element
+      - name: y2
+        datatype: float32
+        unit: pixels
+        title: y location of the lower right corner of a given wavelength element
+      - name: x3
+        datatype: float32
+        unit: pixels
+        title: x location of the lower left corner of a given wavelength element 
+      - name: y3
+        datatype: float32
+        unit: pixels
+        title: y location of the lower left corner of a given wavelength element
+- type: object
+  properties:
+    meta:
+      type: object
+      properties:
+        x_ref:
+          type: number
+          title: x coord of ref position of MIRIM_SLIT
+          default: pixels
+          fits_keyword: IMX
+        y_ref:
+          type: number
+          title: y coord of ref position of MIRIM_SLIT
+          default: pixels
+          fits_keyword: IMY
+        x_ref_slitless:
+          type: number
+          title: x coord of ref position of MIRIM_SLITLESS
+          default: pixels
+          fits_keyword: IMXSLTL
+        y_ref_slitless:
+          type: number
+          title: y coord of ref position of MIRIM_SLITLESS
+          default: pixels
+          fits_keyword: IMYSLTL
+        v2_vert1:
+          type: number
+          title: Slit vertex 1 in V2 frame
+          fits_keyword: V2VERT1
+        v2_vert2:
+          type: number
+          title: Slit vertex 2 in V2 frame
+          fits_keyword: V2VERT2
+        v2_vert3:
+          type: number
+          title: Slit vertex 3 in V2 frame
+          fits_keyword: V2VERT3
+        v2_vert4:
+          type: number
+          title: Slit vertex 4 in V2 frame
+          fits_keyword: V2VERT4
+        v3_vert1:
+          type: number
+          title: Slit vertex 1 in V3 frame
+          fits_keyword: V3VERT1
+        v3_vert2:
+          type: number
+          title: Slit vertex 2 in V3 frame
+          fits_keyword: V3VERT2
+        v3_vert3:
+          type: number
+          title: Slit vertex 3 in V3 frame
+          fits_keyword: V3VERT3
+        v3_vert4:
+          type: number
+          title: Slit vertex 4 in V3 frame
+          fits_keyword: V3VERT4
--- a/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
@@ -18,47 +18,47 @@ allOf:
       - name: x_center
         datatype: float32
         unit: pixels
-        title: x location of the center of a given wavelength element
+        title: "[pixels] x location of the center of a given wavelength element"
       - name: y_center
         datatype: float32
         unit: pixels
-        title: y location of the center of a given wavelength element
+        title: "[pixels] y location of the center of a given wavelength element"
       - name: wavelength
         datatype: float32
         unit: microns
-        title: central wavelength of the wavelength element at x_center, y_center
+        title: "[microns] central wavelength of the wavelength element at x_center, y_center"
       - name: x0
         datatype: float32
         unit: pixels
-        title: x location of the upper left corner of a given wavelength element
+        title: "[pixels] x location of the upper left corner of a given wavelength element"
       - name: y0
         datatype: float32
         unit: pixels
-        title: y location of the upper left corner of a given wavelength element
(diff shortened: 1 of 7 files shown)


jemorrison:
@braingram This PR is ready for another review

jemorrison:
This is ready to be merged so we can get the new reference file in CRDS

braingram:
Would you fix the failing style check?
https://github.com/spacetelescope/stdatamodels/actions/runs/13332329197/job/37239256929?pr=393

I think a few comments/suggestions were lost when this new PR was opened.

Would you also add a unit test (re: https://github.com/spacetelescope/stdatamodels/pull/382#issuecomment-2631633221)?

Also:
> Finally if this is a reference model mapped to a CRDS reference file the test_schema_against_crds file could use an update (to use the new reference model):

Is there an example file yet for this reference model? If so, would you share it so I can test this PR?

braingram:
Also requesting @nden as a reviewer to follow up on comments from #382

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,120 @@
...
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
+id: "http://stsci.edu/schemas/jwst_datamodel/specwcs_miri_lrs.schema"
+title: MIRI LRS Spec Schema
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_pexptype.schema
+- $ref: keyword_exptype.schema
+- $ref: keyword_readpatt.schema
+- $ref: keyword_filter.schema
+- $ref: keyword_band.schema
```
nden, line 12: Are `band` and `readpatt` really necessary?

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,120 @@
...
+        x_ref_slitless:
+          type: number
+          title: x coord of ref position of MIRIM_SLITLESS
+          default: pixels
+          fits_keyword: IMXSLTL
+        y_ref_slitless:
+          type: number
+          title: y coord of ref position of MIRIM_SLITLESS
+          default: pixels
+          fits_keyword: IMYSLTL
```
nden, line 88: Is this schema going to be used for SLITLESS mode? Do we expect a slitless mode reference file?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -398,6 +399,107 @@ def to_fits(self):
         raise NotImplementedError("FITS format is not supported for this file.")
 
 
+class MiriLRSSpecwcsModel(ReferenceFileModel):
+    """
+    A model for a reference file of type "specwcs" for MIRI LRS Slit.
+    The model is for the specwcs for LRS Fixed Slit and LRSSlitless
```
nden, line 405: Do we expect an LRS slitless reference file?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -398,6 +399,107 @@ def to_fits(self):
...
+            self.meta.v3_vert2 = v3_vert2
+        if v3_vert3 is not None:
+            self.meta.v3_vert3 = v3_vert3
+        if v3_vert4 is not None:
+            self.meta.v3_vert4 = v3_vert4
+
+    def populate_meta(self):
+        self.meta.instrument.name = "MIRI"
+        self.meta.instrument.detector = "MIRIMAGE"
+        self.meta.reftype = self.reftype
```
nden, line 488: This needs more keywords to define a LRS fixed slit mode - filter?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
jemorrison: no we will not have an LRS slitless reference file

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
jemorrison: This schema is supposed to work for both FIX slit and slitless.  The slitless data does not need the v2/v3 vertices just the FIXED slit case

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
jemorrison: added  self.meta.instrument.filter = 'P750L'

jemorrison:
@braingram I have an example reference file. How should I share it with you ?

braingram:
> @braingram I have an example reference file. How should I share it with you ?

Would you upload it to the linked jira ticket? The link in the description is broken, would you fix that as well?

jemorrison:
@braingram 
The reference file is already on the JIRI ticket: https://jira.stsci.edu/browse/JP-3848
Click on the first attached file jwst_miri_specwcs_2025jan17.fits
I just downloaded it so you should be able to get it.

jemorrison:
@braingram Sorry I am not tracking.
Looking at the other specwcs reference models: NIRCAMGrismModel, NIRISSGrismModel.
These models are not defined in test_schema_against_cdrs.py
I looked at how they are used in the pipeline:
with NIRCAMGrismModel(reference_files['specwcs'])
with NIRCAMGrismModel(reference_files['specwcs'])

This is how I am using the MIRILRS one:
refmodel = MiriLRSSpecwcsModel(reference_files['specwcs'])

So do I need it in test_schema_against_crds.py ?

braingram:
> @braingram Sorry I am not tracking. Looking at the other specwcs reference models: NIRCAMGrismModel, NIRISSGrismModel. These models are not defined in test_schema_against_cdrs.py I looked at how they are used in the pipeline: with NIRCAMGrismModel(reference_files['specwcs']) with NIRCAMGrismModel(reference_files['specwcs'])
> 
> This is how I am using the MIRILRS one: refmodel = MiriLRSSpecwcsModel(reference_files['specwcs'])
> 
> So do I need it in test_schema_against_crds.py ?

Yeah I would say so. I believe @tapastro wrote the test originally. I think it's a bug (or lack of coverage) that `NIRCAMGrismModel` isn't covered but I'd say let's not extend that bug to the new model added in this PR.

jemorrison:
@braingram  Sorry again
This is obviously wrong because it fails. But what should it be
it seems the format is ref_type: model

"specwcs": dm.SpecwcsModel, dm.MiriLRSSpecwcsModel,
this does not work either:
"specwcs": dm.SpecwcsModel,
"specwcs": dm.MiriLRSSpecwcsModel,

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
jemorrison: @nden Let me clarify as understand it there is a reference file for FIXED slit and a reference file for SLITLESS. But Both of these files are read in with the same datamodel. The existing slitless reference on CRDS will be used.
A new reference file for FIXED SLIT LRS will have the v2,v3 vertices. 
Do you see any problem with this. I tested reading in the new reference file for LRS FIXED slit and the existing slitless reference file using this 1 datamodel model and they are read in correctly

jemorrison:
@braingram Made I need to do something like cubepar_model_map does and then add it to
ref_to_multiples_dict ?
Then I can all the NIRCAM and NIRSSS ones too ?

tapastro:
@jemorrison It looks like your recent updates got you most of the way there - there was a typo in the NRC_WFSS exposure type, and once that is added it uncovered a bug in the NIRCam grism model that this test was designed to catch - so I added the schema reference to fix it. We'll see if tests pass now!

src/stdatamodels/jwst/datamodels/tests/test_specwcs_mirilrs_datamodel.py
```diff
@@ -0,0 +1,59 @@
...
+    # for slitless case assert the v2/v3 vertices are None if not in the file
+    assert model.meta.v2_vert1 is None
+    assert model.meta.v2_vert2 is None
+    assert model.meta.v2_vert3 is None
+    assert model.meta.v2_vert4 is None
+
+    assert model.meta.v3_vert1 is None
+    assert model.meta.v3_vert2 is None
+    assert model.meta.v3_vert3 is None
+    assert model.meta.v3_vert4 is None
```
braingram, line 59:
```suggestion
    assert model.meta.v3_vert4 is None
    model.validate()
```
Adding a call to `validate` here should increase the patch coverage.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -398,6 +399,112 @@ def to_fits(self):
...
+        self.meta.instrument.filter = "P750L"
+
+    def validate(self):
+        super(MiriLRSSpecwcsModel, self).validate()
+        try:
+            assert self.meta.instrument.name == "MIRI"
+            assert self.meta.instrument.detector == "MIRIMAGE"
+            assert self.meta.reftype == self.reftype
```
braingram, line 500: This fails with the example file which has `SPECWCS` instead of `specwcs`.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,118 @@
...
+          fits_keyword: IMY
+        x_ref_slitless:
+          type: number
+          title: x coord of ref position of MIRIM_SLITLESS
+          default: pixels
+          fits_keyword: IMXSLTL
+        y_ref_slitless:
+          type: number
+          title: y coord of ref position of MIRIM_SLITLESS
+          default: pixels
```
braingram, line 85:
```suggestion
          default: pixels
```
Is this supposed to be units? This results in a default value of 'pixels' when creating a reference model:
```python
>> m = dm.MiriLRSSpecwcsModel()
>> m.meta.y_ref_slitless
'pixels'
```
There are a few other instances of this in the schema.

braingram (commented):
Thanks for making the test updates and pointing me to the example file.

I left a few comments about:
- adding a call to validate in the unit test
- clarification of the `default` schema values
- a comment/question about the reftype in the example file causing validation failure

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
jemorrison: so is this really a failure of the reference file not having the correction retype ?
Is it ok to do this to account for that problem 
assert self.meta.reftype == self.reftype.lower()

jemorrison: So I have been using the example reference file and this model to read in the data. So how does it fail for you ?

braingram: What I ran was:
```python
>> import stdatamodels.jwst.datamodels as dm
>> m = dm.MiriLRSSpecwcsModel("jwst_miri_specwcs_2025jan17.fits")
>> m.validate()
<ipython-input-5-6a8bfee86cf8>:1: ValidationWarning: Traceback (most recent call last):
  File "/Users/bgraham/projects/src/stdatamodels/src/stdatamodels/jwst/datamodels/wcs_ref_models.py", line 500, in validate
    assert self.meta.reftype == self.reftype
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError

  m.validate()
```
This might just be an issue with the example file. Presumably the model will be used to generate a new reference model. In that case `reftype` will match (since it's set in `populate_meta`). However if I make a blank reference model it also fails to validate:
```python
>>> m = dm.MiriLRSSpecwcsModel()
>>> m.validate()
/Users/bgraham/projects/src/stdatamodels/src/stdatamodels/jwst/datamodels/reference.py:41: ValidationWarning: Model.meta is missing values for ['description', 'author', 'pedigree', 'useafter']
  self.print_err(f"Model.meta is missing values for {to_fix}")
```
@tapastro is that expected? I'm not that familiar with how these reference models are used.

jemorrison: I am not sure the reference model will be used to create the reference file. It is being created in IDL.
So my test is probably not as complete as yours
 from stdatamodels.jwst import datamodels
  file = 'jwst_miri_specwcs_2025jan17.fits'
  model = datamodels.MiriLRSSpecwcsModel(file)

that seems to work fine
When I added this line to validate MiriLRSSpecwcs function in wcs_ref_models.py assert self.meta.reftype.lower() == self.reftype
It now validates - is that cheating ?

jemorrison:
@tapastro @braingram I am still getting an error when I run pytest test_schema_against_crds.py
I am getting an error related to --no-crds
How am I supposed to run this test. I have been just typing

pytest  test_schema_against_crds.py

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
braingram: I think that might be a crds question. @stscieisenhamer are there any special requirements for reference file `reftype` (are they case sensitive)?

tapastro:
> @tapastro @braingram I am still getting an error when I run pytest test_schema_against_crds.py I am getting an error related to --no-crds How am I supposed to run this test. I have been just typing
> 
> pytest test_schema_against_crds.py

I had the same issue - stdatamodels has a test configuration file in the base test directory that doesn't get picked up if you run these JWST tests in their local directory. If cd to the stdatamodels base directory and run something like `pytest -k selectors`, it should pick up the config option.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
braingram: I missed that these are not in an array datatype block and are instead for properties (datamodel attributes).

`units` isn't an understood schema keyword (nor is `unit`). I believe the convention for keywords is to put the unit before the title in square brackets. Something like:
```yaml
title: "[pixels] y coord of ref position of MIRIM_SLITLESS"
```

src/stdatamodels/jwst/datamodels/tests/test_specwcs_mirilrs_datamodel.py
jemorrison: done

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,114 @@
...
+        unit: pixels
+        title: x location of the center of a given wavelength element
+      - name: y_center
+        datatype: float32
+        unit: pixels
+        title: y location of the center of a given wavelength element
+      - name: wavelength
+        datatype: float32
+        unit: microns
+        title: central wavelength of the wavelength element at x_center, y_center
```
braingram, line 29:
```suggestion
        title: "[microns] central wavelength of the wavelength element at x_center, y_center"
```

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,114 @@
...
+        unit: pixels
+        title: x location of the upper right corner of a given wavelength element
+      - name: y1
+        datatype: float32
+        unit: pixels
+        title: y location of the upper right corner of a given wavelength element
+      - name: x2
+        datatype: float32
+        unit: pixels
+        title: x location of the lower right corner of a given wavelength element
```
braingram, line 49:
```suggestion
        title: "[pixels] x location of the lower right corner of a given wavelength element"
```

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,114 @@
...
+        unit: pixels
+        title: y location of the lower right corner of a given wavelength element
+      - name: x3
+        datatype: float32
+        unit: pixels
+        title: x location of the lower left corner of a given wavelength element 
+      - name: y3
+        datatype: float32
+        unit: pixels
+        title: y location of the lower left corner of a given wavelength element
```
braingram, line 61:
```suggestion
        title: "[pixels] y location of the lower left corner of a given wavelength element"
```

braingram (commented):
These suggestions update the titles to follow the convention of having the unit in square brackets at the start of the title. This convention is followed by many of the existing keywords (see [the core schema for some examples](https://github.com/spacetelescope/stdatamodels/blob/26c9f0f373af8fdfa095b97d1e4b7441e21ef1cd/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml#L1385)) and I suggest that we follow it for new keywords.

jemorrison:
@braingram I made those changes but now I can not get the model to work. 
Maybe I need to set the units off all the arrays ?

This is what I am doing:
wavetable = np.array([(xc[0], yc[0], wave[0], x0[0], y0[0], x1[0], y1[0],
                       x2[0], y2[0], x3[0], y3[0] ),
                      (xc[1], yc[1], wave[1], x0[1], y0[1], x1[1], y1[1],
                       x2[1], y2[1], x3[1], y3[1] ),
                      (xc[2], yc[2], wave[2], x0[2], y0[2], x1[2], y1[2],
                       x2[2], y2[2], x3[2], y3[2] )],dtype=d)

model = datamodels.MiriLRSSpecwcsModel(x_ref=430, y_ref=400, wavetable=wavetable)

braingram:
Thanks. The changes pushed in https://github.com/spacetelescope/stdatamodels/pull/393/commits/823ee6408d9a506b6ab71a7a0b4561b263a48608 don't match the suggestions. They contain yaml syntax errors that are causing the CI failures and the failure you're seeing. There are also still "unit" keywords left in the table datatype that I suggested to remove.

jemorrison:
@braingram Ok thanks I missed the yaml errors and I misunderstood about units. I think I may finally have it. 
Thanks for your help with this.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_lrs.schema.yaml
```diff
@@ -0,0 +1,103 @@
...
+        datatype: float32
+        title: "[pixels] y location of the lower left corner of a given wavelength element"
+- type: object
+  properties:
+    meta:
+      type: object
+      properties:
+        x_ref:
+          type: number
+          title: "[pixels]  x coord of ref position of MIRIM_SLIT"
```
braingram, line 58:
```suggestion
          title: "[pixels] x coord of ref position of MIRIM_SLIT"
```

braingram (commented):
Thanks for all the updates to this. The changes look good to me.

Would you run the regression tests with this PR?

Otherwise I'll leave the approval up to @nden

jemorrison:
@braingram To run the regression tests on datamodels do I just select the JWST Calibration Pipeline workflow and set the git repo the spacetelescope/datamodels. Or is running a regression test on the jwst branch I have made that uses this datamodel model sufficient to test this ? 
I have a first run at https://github.com/spacetelescope/RegressionTests/actions/runs/13533041283 - but I will try a second one because there are failures not related to this PR.

braingram:
> @braingram To run the regression tests on datamodels do I just select the JWST Calibration Pipeline workflow and set the git repo the spacetelescope/datamodels. Or is running a regression test on the jwst branch I have made that uses this datamodel model sufficient to test this ? I have a first run at https://github.com/spacetelescope/RegressionTests/actions/runs/13533041283 - but I will try a second one because there are failures not related to this PR.

I think the most useful run for this PR would be using `spacetelescope/jwst` (jwst main) and providing a dependency override to install the source branch for this PR `git+https://github.com/jemorrison/stdatamodels.git@MIRI_LRS_Specwcs`. If that passes we'd know that merging this PR wouldn't break jwst main. Would you start that run? Let me know if you have any questions or if there's anything I can do to help.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
drlaw1558: Looking at a number of other reference files currently in CRDS it looks like those too have capitalized reftype valuese; e.g., APCORR, CUBEPAR, GAIN, etc.

jemorrison:
@braingram results of regression test:
https://github.com/spacetelescope/RegressionTests/actions/runs/13549832932

braingram (approved):
Thanks. Since the regtests pass I'd say we merge this and any additional changes can be addressed in follow up PRs.

PR #394, opened 2025-02-05
Initial pass at new pictureframe model

drlaw1558:
This is an initial attempt at a new datamodel for the NIRSpec thermal picture frame reference file, see https://jira.stsci.edu/browse/CRDS-925

### DIFF
--- a/src/stdatamodels/jwst/datamodels/pictureframe.py
+++ b/src/stdatamodels/jwst/datamodels/pictureframe.py
@@ -1,7 +1,7 @@
 from stdatamodels.dynamicdq import dynamic_mask
+
 from .dqflags import pixel
 from .reference import ReferenceFileModel
-
 
 __all__ = ["PictureFrameModel"]
 
@@ -11,6 +11,7 @@ class PictureFrameModel(ReferenceFileModel):
     A data model for 2D thermal picture frame reference files.
 
     Parameters
+    ----------
     __________
     data : numpy float32 array
          The science data


drlaw1558:
Put this back to draft since there may still be some iteration on the file format over the next week or so.

melanieclarke:
@drlaw1558 - did you want to get this new datamodel into Build 11.3?

drlaw1558:
@melanieclarke I believe we determined that all we needed for 11.3 was the CRDS rmap for these files, so it should be ok to punt this datamodel to 12.0

melanieclarke:
@drlaw1558 - did you want to add pictureframe for Build 12.0, or are we punting again?

drlaw1558:
@melanieclarke We're punting again as the algorithm isn't finalized, new target is 12.1.

src/stdatamodels/jwst/datamodels/pictureframe.py
```diff
@@ -0,0 +1,38 @@
...
+
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/pictureframe.schema"
+
+    def __init__(self, init=None, **kwargs):
+        super(PictureFrameModel, self).__init__(init=init, **kwargs)
+
+        self.dq = dynamic_mask(self, pixel)
+
+        # Implicitly create arrays
+        self.dq = self.dq
```
emolter, line 37:
```suggestion
```
This isn't needed because dq is already touched by the line above

src/stdatamodels/jwst/datamodels/pictureframe.py
```diff
@@ -0,0 +1,38 @@
...
+        self.err = self.err
```
emolter, line 38: Are we sure that having a zero-filled error array is what we want in all cases where the model is initialized without an error array? That's what this line will enforce.  If the line is removed, the error array will just not be populated until assigned.

drlaw1558:
@melanieclarke Please feel free to take over!

melanieclarke:
Closing in favor of #607

PR #395, opened 2025-02-10
Validate dimension when instantiating datamodel from shape

### LINKED ISSUE #307: datamodels doesn't validate arrays in some cases
`jwst.datamodels` validates array sizes on reading from a FITS file or on assignment.  So `ImageModel` requires the data array to be 2D, and so if I try to read or assign a 3D array, it fails validation:
```
In [22]: m = datamodels.ImageModel()

In [24]: m.data = np.arange(10*10).reshape((1, 10, 10))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-24-d91d60f9cc1c> in <module>
----> 1 m.data = np.arange(10*10).reshape((1, 10, 10))

~/dev/jwst/jwst/datamodels/model_base.py in __setattr__(self, attr, value)
    623             object.__setattr__(self, attr, value)
    624         elif ndmodel.NDModel.my_attribute(self, attr):
--> 625             ndmodel.NDModel.__setattr__(self, attr, value)
    626         else:
    627             properties.ObjectNode.__setattr__(self, attr, value)

~/dev/jwst/jwst/datamodels/ndmodel.py in data(self, value)
    130         if not primary_array_name:
    131             primary_array_name = 'data'
--> 132         properties.ObjectNode.__setattr__(self, primary_array_name, value)
    133 
    134     @property

~/dev/jwst/jwst/datamodels/properties.py in __setattr__(self, attr, val)
    258             if val is None:
    259                 val = _make_default(attr, schema, self._ctx)
--> 260             val = _cast(val, schema)
    261 
    262             node = ObjectNode(attr, val, schema, self._ctx)

~/dev/jwst/jwst/datamodels/properties.py in _cast(val, schema)
     37             raise ValueError(
     38                 "Array has wrong number of dimensions.  Expected {0}, got {1}".format(
---> 39                     schema['ndim'], len(val.shape)))
     40 
     41         if 'max_ndim' in schema and len(val.shape) > schema['max_ndim']:

ValueError: Array has wrong number of dimensions.  Expected 2, got 3
```
That's good.

But if I instantiate an `ImageModel` with a 3D array or from an already existing datamodel with a 3D array, it doesn't do any validation and allows it.
```
In [25]: m = datamodels.ImageModel((1, 10, 10))

In [26]: cube = datamodels.CubeModel((1, 10, 10))

In [27]: m = datamodels.ImageModel(cube)
```
This is incorrect. It should be giving the same validation error as above.

emolter:
Closes #307 

This PR addresses a bug when instantiating datamodels from an array shape.  The problem is discussed in detail on the issue, but in short, the input shape was not checked against the schema `ndim` attribute when creating a default array

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/tests/test_models.py
+++ b/src/stdatamodels/jwst/datamodels/tests/test_models.py
@@ -142,7 +142,7 @@ def test_model_with_nonstandard_primary_array():
         def get_primary_array_name(self):
             return "wavelength"
 
-    m = _NonstandardPrimaryArrayModel((10,))
+    m = _NonstandardPrimaryArrayModel((10, 10))
     assert "wavelength" in list(m.keys())
     assert m.wavelength.sum() == 0
 
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -99,9 +99,11 @@ def test_init_with_array2():
         with BasicModel(array) as dm:
             dm.data  # noqa: B018
 
+
 def test_init_invalid_shape():
     with pytest.raises(ValueError):
         BasicModel((50, 50, 50))
+
 
 def test_set_array():
     with pytest.raises(ValueError):
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -99,6 +99,9 @@ def test_init_with_array2():
         with BasicModel(array) as dm:
             dm.data  # noqa: B018
 
+def test_init_invalid_shape():
+    with pytest.raises(ValueError):
+        BasicModel((50, 50, 50))
 
 def test_set_array():
     with pytest.raises(ValueError):
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -101,8 +101,19 @@ def test_init_with_array2():
 
 
 def test_init_invalid_shape():
+    """Requested some number of dimensions unequal to ndim"""
     with pytest.raises(ValueError):
-        BasicModel((50, 50, 50))
+        BasicModel((50,))
+
+
+def test_init_invalid_shape2():
+    """Requested more dimensions than max_ndim"""
+    with BasicModel() as dm:
+        schema = dm._schema
+    schema["properties"]["data"]["max_ndim"] = 1
+    schema["properties"]["data"]["ndim"] = None
+    with pytest.raises(ValueError):
+        BasicModel((50, 50), schema=schema)
 
 
 def test_set_array():
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -101,7 +101,7 @@ def test_init_with_array2():
 
 
 def test_init_invalid_shape():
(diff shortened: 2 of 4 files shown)


src/stdatamodels/jwst/datamodels/tests/test_models.py
```diff
@@ -142,7 +142,7 @@ class _NonstandardPrimaryArrayModel(JwstDataModel):
         def get_primary_array_name(self):
             return "wavelength"
 
-    m = _NonstandardPrimaryArrayModel((10,))
+    m = _NonstandardPrimaryArrayModel((10, 10))
```
emolter, line 145: This appeared to be an instance of an array passing validation that we would expect to fail, and indeed it does fail as of the bugfix in this PR.  It doesn't look like this change makes much difference for what the test intends to cover, but someone can correct me if that's wrong.

emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/13245788321 are all passing now

emolter:
@jdavies-st Not sure why I can't request you as a reviewer, so pinging you in the comments to take a look

tests/test_models.py
```diff
@@ -100,6 +100,22 @@ def test_init_with_array2():
             dm.data  # noqa: B018
 
 
+def test_init_invalid_shape():
+    """Requested some number of dimensions unequal to ndim"""
```
jdavies-st, line 104: You may want just mention in the docstring here what the expected dimension of `BasicModel.data` should be.

jdavies-st (commented):
Looks good. I would also run the `jwst` regtests against this (or this combined with the other changes) to make sure there's no misuse of this bug in the pipeline code itself.  Same with `romancal`.

emolter:
@jdavies-st I already ran the jwst regtests here: https://github.com/spacetelescope/RegressionTests/actions/runs/13245788321

Romancal no longer uses stdatamodels.

jdavies-st (approved):
Doh!  Sorry, missed that.  I can't see those results as I'm no longer internal to ST.  But everything looks good from my side.

jdavies-st:
> @jdavies-st Not sure why I can't request you as a reviewer, so pinging you in the comments to take a look

It seems this repo only accepts reviews from people with write access to the repo.   I'm no longer internal at STScI, so I don't have write access here.  So my review above looks like a normal comment, not a review.

emolter:
> It seems this repo only accepts reviews from people with write access to the repo. I'm no longer internal at STScI, so I don't have write access here. So my review above looks like a normal comment, not a review.

Gotcha, well thanks for looking anyway.  I can see your approval, but it shows up as a gray checkmark instead of a green checkmark

braingram (approved):
LGTM.

PR #396, opened 2025-02-11
Move find_fits_keyword to fits_support

### LINKED ISSUE #33: Move FITS-related functions from schema.py to fits_support.py
@jdavies-st pointed out in a [PR comment](https://github.com/spacetelescope/stdatamodels/pull/30#discussion_r518947068) that `stdatamodels.schema.find_fits_keyword` probably ought to be moved to `stdatamodels.fits_support`.

emolter:
Closes #33 

This PR makes a small reorganization, moving the FITS-format-specific function `find_fits_keyword` out of `schema/` and into `fits_support/`.  Note that `find_fits_keyword` is also a method of a DataModel, and that method is unchanged except that it internally calls from elsewhere in the codebase. 

That is, the documented way to call this function, e.g., `model.find_fits_keyword('DATE-OBS')`, is unchanged.  

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -929,3 +929,38 @@ def ensure_ascii(s):
     if isinstance(s, bytes):
         s = s.decode("ascii")
     return s
+
+
+def find_fits_keyword(schema, keyword, return_result=False):
+    """
+    Utility function to find a reference to a FITS keyword in a given
+    schema.  This is intended for interactive use, and not for use
+    within library code.
+
+    Parameters
+    ----------
+    schema : JSON schema fragment
+        The schema in which to search.
+
+    keyword : str
+        A FITS keyword name
+
+    Returns
+    -------
+    locations : list of str
+
+    Notes
+    -----
+    return_result is included for backward compatibility and is not used.
+    """
+
+    def find_fits_keyword(subschema, path, combiner, ctx, recurse):
+        if len(path) and path[0] == "extra_fits":
+            return True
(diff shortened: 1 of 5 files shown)


emolter:
This function does not appear to be used at all in JWST, and indeed it's documented that it is not supposed to be used there, but running regression tests anyway just in case https://github.com/spacetelescope/RegressionTests/actions/runs/13254071328

emolter:
@braingram @tapastro Would it be better to deprecate `schemas.find_fits_keyword()`?  On the one hand, it's theoretically possible someone could be calling `from stdatamodels.schemas import find_fits_keyword`.  On the other hand, it seems unlikely to me anyone actually is doing so, since the documented way to call this functionality is as a method on the datamodels, not to call this function.  In `jwst` we implicitly assume anything not in the `__all__` is not "public" and requires no deprecation cycle, is that typically the same in `stdatamodels`?

braingram:
I wouldn't consider `stdatamodels.schema` as part of the public API since it's not in the documentation:
https://stdatamodels.readthedocs.io/en/latest/api.html
Since it's not documented I don't think any deprecation is needed.

I also don't have any objection to keeping it in `schema`. It's a function that walks a schema looking for a particular schema keyword (`fits_keyword`). It is related to FITS since the `fits_keyword` refers to a FITS keyword but it's not needed for writing datamodels to FITS files (which I believe is the goal of `fits_support`).

emolter:
I am also happy just closing the issue, sound like that would be your preference?

PR #399, opened 2025-02-11
increase minimum versions to match jwst

braingram:
This PR updates a few dependencies to match those in jwst:

- asdf-astropy to 0.6.0 (pending https://github.com/spacetelescope/jwst/pull/9177)
- asdf to [3.3](https://github.com/spacetelescope/jwst/blob/d8033b3dd52a15c40fb034eededb4684afdd03f3/pyproject.toml#L21)
- astropy to [6.1](https://github.com/spacetelescope/jwst/blob/d8033b3dd52a15c40fb034eededb4684afdd03f3/pyproject.toml#L22)

## Tasks

</details>

### DIFF
--- a/changes/399.misc.rst
+++ b/changes/399.misc.rst
@@ -1,0 +1,1 @@
+Increase minimum versions of asdf-astropy, asdf and astropy dependencies.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,11 +12,11 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 dependencies = [
-    "asdf>=3.1.0",
+    "asdf>=3.3.0",
     "asdf-transform-schemas>=0.5.0",
     "asdf-astropy>=0.6.0",
     "numpy>=1.25",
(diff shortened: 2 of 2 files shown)


emolter (approved):
LGTM, but a question.  Is there value in having upper pins here, too?  I see an upper pin on asdf in jwst specifying `<5`

braingram:
> LGTM, but a question. Is there value in having upper pins here, too? I see an upper pin on asdf in jwst specifying `<5`

Thanks!

I'd say yes and no. Adding an upper pin if we know there will be incompatible changes that we expect will break the code seems worthwhile so that the old versions on pypi don't suddenly start failing when dependencies are updated. However since this is more a "library" than an "application" (like jwst) having as loose as possible pins is helpful for building environments. If we don't have an expectation that the newer versions of dependencies will break this package I'd say we can leave off the upper pins. For asdf in particular there are no plans for breaking changes in 5 that would impact stdatamodels (although the plan for 5.0 is not finalized).

PR #402, opened 2025-02-12
Deprecate extra_fits

### LINKED ISSUE #295: Deprecate or remove `extra_fits`
As noted in https://jira.stsci.edu/browse/JP-2056 the documentation for `extra_fits` incorrectly refers to it as `_extra_fits`:
https://stdatamodels.readthedocs.io/en/latest/jwst/datamodels/models.html#extra-fits-keywords

There are other references to `extra_fits`: 
https://stdatamodels.readthedocs.io/en/latest/search.html?q=extra_fits&check_keywords=yes&area=default

We should consider deprecating `extra_fits` as part of the "public" API (it appears to only have a single use in jwst):
https://github.com/spacetelescope/jwst/blob/33fd9fa52debf69ccb53aab32ab1073425bc4418/jwst/model_blender/tests/test_blend.py#L127

We should continue to "roundtrip" non-schema fits data but allowing users to arbitrarily update these data while in memory seems dangerous. Any updates can instead use the existing `astropy.io.fits` API.

Since `extra_fits` appears in the documentation I suggest we:
- deprecate it
- update the documentation
- remove it in a later major version (I'm milestoneing this issue for 2.0.0)

emolter:
Closes #295 

Resolves [JP-3886](https://jira.stsci.edu/browse/JP-3886)

This PR deprecates direct reading and setting of the `extra_fits` and `_extra_fits` attributes of datamodels.

## Tasks

</details>

### DIFF
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -293,25 +293,29 @@ use::
 Extra FITS keywords
 -------------------
 
+.. warning::
+
+    This feature is deprecated and will be removed in a future release.
+
 When loading arbitrary FITS files, there may be keywords that are not
 listed in the schema for that data model.  These "extra" FITS keywords
-are put under the model in the `_extra_fits` namespace.
+are put under the model in the `extra_fits` namespace.
 
-Under the `_extra_fits` namespace is a section for each header data
+Under the `extra_fits` namespace is a section for each header data
 unit, and under those are the extra FITS keywords.  For example, if
 the FITS file contains a keyword `FOO` in the primary header, its
 value can be obtained using::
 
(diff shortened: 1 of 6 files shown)


emolter:
@braingram I have to admit I've never had to deprecate access to an attribute before, nor deprecate something that is currently used in so many places.  Leaving this as draft for now - is this anywhere near what you would have envisioned?

braingram:
> @braingram I have to admit I've never had to deprecate access to an attribute before, nor deprecate something that is currently used in so many places. Leaving this as draft for now - is this anywhere near what you would have envisioned?

Thanks. Yeah I think this is on the right track. Since we're awaiting feedback I'll defer a detailed look until we've heard back.

PR #418, opened 2025-02-24
test with latest supported Python version

zacharyburnett:
Resolves [SCSB-195](https://jira.stsci.edu/browse/SCSB-195)

## Tasks

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,11 +63,13 @@ jobs:
       cache-path: ${{ needs.data.outputs.crds_path }}
       cache-key: crds-${{ needs.data.outputs.crds_context }}
       envs: |
-        - macos: py311-xdist
-        - linux: py311-cov-xdist
-          coverage: codecov
+        - linux: py311-xdist
         - linux: py312-xdist
-        - linux: py313-xdist
+        # `tox` does not currently respect `requires-python` versions when creating testing environments;
+        # if this breaks, add an upper pin to `requires-python` and revert this py3 to the latest working version
+        - linux: py3-cov-xdist
+          coverage: codecov
+        - macos: py3-xdist
   test_downstream:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@8c0fde6f7e926df6ed7057255d29afa9c1ad5320  # v1.16.0
     needs: [ data ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,8 +65,13 @@ jobs:
       envs: |
         - linux: py311-xdist
         - linux: py312-xdist
-        # `tox` does not currently respect `requires-python` versions when creating testing environments;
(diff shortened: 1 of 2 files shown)


emolter (commented):
I am a bit confused by this change.  What is the benefit of using `py3` instead of being explicit about `py313`?

What does your inline comment mean - to me, it seems to indicate that as soon as Python 3.14 becomes a thing and `tox` decides to support it, that test is going to start running 3.14 (even if we aren't ready for it yet) and break our stuff.  The comment says that `tox` doesn't respect the `requires-python` pin, but then it says that if this breaks we should add a `requires-python` pin... but what's the point, if it isn't respected?

braingram:
@emolter any objections to this?

.github/workflows/ci.yml
```diff
@@ -63,11 +63,13 @@ jobs:
...
-        - linux: py311-cov-xdist
-          coverage: codecov
+        - linux: py311-xdist
         - linux: py312-xdist
-        - linux: py313-xdist
+        # `tox` does not currently respect `requires-python` versions when creating testing environments;
+        # if this breaks, add an upper pin to `requires-python` and revert this py3 to the latest working version
+        - linux: py3-cov-xdist
+          coverage: codecov
+        - macos: py3-xdist
```
braingram, line 72:
```suggestion
        # `tox` does not currently use the `requires-python` versions 
        # defined in pyproject.toml. This means even if we set an upper limit
        # for python in that file the `py3` below will try to use the newest
        # python even if it's above that limit.
        # If we don't support the latest stable python, update `py3` below
        # to use only the latest supported version (matching what is set
        # for `requires-python` in `pyproject.toml`)
        - linux: py3-cov-xdist
          coverage: codecov
        - macos: py3-xdist
```
@emolter How does this sound?

zacharyburnett:
> If we're not compatible by the time py3 becomes py314 I think it's beneficial to have the CI start breaking to show we're not compatible with 3.14. If we instead used py313 instead of py3 we could miss testing against 3.14 even after it becomes stable.

yes this exactly, sorry for being confusing!

.github/workflows/ci.yml
emolter: That is much better thank you!

zacharyburnett: this is more explanatory than what I had, thanks! Hopefully we'll be able to leverage `tox` and `uv` in the future to auto-limit the Python version based on project metadata, but alas...

braingram (approved):
LGTM. I believe branch protections will need to be updated after merge. @zacharyburnett let me know if you'd like me to do that (I think I have permissions).

zacharyburnett:
> LGTM. I believe branch protections will need to be updated after merge. @zacharyburnett let me know if you'd like me to do that (I think I have permissions).

no worries, I can take care of that thanks!

emolter (approved):

PR #422, opened 2025-02-25
Replace os.path with pathlib in filetype check

### LINKED ISSUE #387: Refactor file_type `check` function to use `Path`
xref: https://github.com/spacetelescope/stdatamodels/pull/383/files/1ce97af164bc6292c34c1a4bcdc35a0e79452181#r1937755328

emolter:
Partially resolves [JP-3884](https://jira.stsci.edu/browse/JP-3884)

Closes #387 

This PR switches `filetype.check()` from os.path to pathlib.Path.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/filetype.py
+++ b/src/stdatamodels/filetype.py
@@ -1,4 +1,4 @@
-import os
+from pathlib import Path
 
 
 def check(init):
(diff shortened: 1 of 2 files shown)


src/stdatamodels/filetype.py
```diff
@@ -18,18 +18,20 @@ def check(init):
...
+        if len(path.suffixes) == 0:
             raise ValueError(f"Input file path does not have an extension: {init}")
 
+        ext = path.suffixes[-1].strip(".")
         if ext not in supported:  # Could be the file is zipped; try splitting again
-            path, ext = os.path.splitext(path)  # noqa: PTH122
-            ext = ext.strip(".")
-
+            err_msg = f"Unrecognized file type for: {init}"
+            try:
```
braingram, line 29: Can we code golf this a bit to remove the try except?
I think something like:
```python
# only consider last 2 suffixes
suffixes = path.suffixes[-2:]
if not suffixes:
    raise ...
for suffix in suffixes[::-1]:
    ext = suffix.strip()
    if ext in ('asdf', 'fits'):
        return ext
    if ext == "json":
        return "asn"
raise ...
```

emolter: sure; https://github.com/spacetelescope/stdatamodels/pull/422/commits/85cd887b532f56434283fc7117754a1499cddd88

braingram (approved):
LGTM.

The more I look at that function the more I want to change it. It doesn't support `Path` objects, it assumes any file object that isn't FITS or ASDF is json, as far as I can tell it's never used with a file object. I'll open a separate issue since that's outside the scope of this PR.

Would you run jwst regtests with this PR before merging?

PR #443, opened 2025-03-13
fix validators run during fits array writing

### LINKED ISSUE #279: `_fits_array_writer` incorrectly checks schema for `dtype` instead of `datatype`
https://github.com/spacetelescope/stdatamodels/blob/8e551d3337ba4b6a2ff1e0d9a2208d987c7defbf/src/stdatamodels/fits_support.py#L283-L284

`dtype` is not used in schemas and `asdf.tags.core.ndarray.validate_dtype` (`ndarray.dtype` above) doesn't exist.

braingram:
Closes #279

This PR fixes the ndim, max_ndim and datatype validation during FITS array writing by yielding from the corresponding validators (these return generators that are not consumed on main).

This PR also fixes the `if` check for `dtype` to instead check for `datatype` (the keyword used in the schema) and uses the custom `_validate_datatype` defined in this package (rather than `validate_dtype` from asdf which doesn't support `ndim` in structured datatypes or `allow_extra_columns`.

An update to one of the `test_multislit.py` unit tests was required where a float64 was being assigned to an array with a float32 dtype defined in the schema.

jwst regression tests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/13841630347/job/38773673047

## Tasks

</details>

### DIFF
--- a/changes/443.bugfix.rst
+++ b/changes/443.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ndim, max_ndim and datatype validation during FITS array writing.
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -289,7 +289,7 @@ def _fits_array_writer(fits_context, validator, _, instance, schema):
     if "max_ndim" in schema:
         yield from ndarray.validate_max_ndim(validator, schema["max_ndim"], instance, schema)
     if "datatype" in schema:
-        yield from ndarray.validate_datatype(validator, schema["datatype"], instance, schema)
+        yield from validate._validate_datatype(validator, schema["datatype"], instance, schema)
 
     hdu_name = _get_hdu_name(schema)
     _assert_non_primary_hdu(hdu_name)
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -285,11 +285,11 @@ def _fits_array_writer(fits_context, validator, _, instance, schema):
         return
 
     if "ndim" in schema:
-        ndarray.validate_ndim(validator, schema["ndim"], instance, schema)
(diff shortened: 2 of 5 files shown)


emolter (commented):
Looks good to me.  I'm a bit confused seeing that there were no regtest failures - didn't you say when we had a Slack huddle that there were going to be many failures in JWST?  What happened to that?

braingram:
> Looks good to me. I'm a bit confused seeing that there were no regtest failures - didn't you say when we had a Slack huddle that there were going to be many failures in JWST? What happened to that?

Magic! :)

It turns out that the assignment validation of array datatypes in stdatamodels differs from what's done in the standard asdf datatype validator. To see all the differences compare:
- https://github.com/spacetelescope/stdatamodels/blob/f504f64263473cb9b61f4bd8496179f0d2e082a0/src/stdatamodels/validate.py#L44
- https://github.com/asdf-format/asdf/blob/d70ac0fa103cb884a7d9fd4276e6a320b68aacb0/asdf/tags/core/ndarray.py#L517

This is already established usage since stdatamodels registers it's own `_validate_datatype` with the jsonschema validator:
https://github.com/spacetelescope/stdatamodels/blob/f504f64263473cb9b61f4bd8496179f0d2e082a0/src/stdatamodels/validate.py#L138
When I fixed the if here:
https://github.com/spacetelescope/stdatamodels/blob/f504f64263473cb9b61f4bd8496179f0d2e082a0/src/stdatamodels/fits_support.py#L291-L292
the standard asdf datatype validator was applied to the datatypes in the schemas (when datamodels were written). This resulted in a bunch of failures (since the stdatamodels datatype definitions don't match the asdf ones). Once I realized this and switched the above line to instead use the stdatamodels `_validate_datatype` the errors went away (except for the one in the unit test fixed in this PR).

emolter (approved):

PR #445, opened 2025-03-13
Define API for lazy-load of metadata attributes from datamodel

### LINKED ISSUE #284: Improve performance of `DataModel` reading for single (or simple) metadata access
There are use cases (like the jwst `resample` step) where loading a single keyword from many `DataModel` containing fits files may be useful. As the number of files might be very large and opening every model might exceed reasonable amounts of RAM it will be important to have a performant way to perform these simple keyword accesses.

Using `meta.wcsinfo.s_region` (contained in the ASDF extension) as an example there are a few ways this keyword can be read:

1) Using `stdatamodels.jwst.datamodels.open`:
```python
with dm.open(fn) as m:
    return m.meta.wcsinfo.s_region
```

2) Using `stdatamodels.asdf_in_fits`:
```python
with asdf_in_fits.open(fn) as af:
    return af['meta']['wcsinfo']['s_region']
```

3) Using `astropy.io.fits` and `asdf.open`:
```python
with fits.open(fn) as ff:
    bs = io.BytesIO(ff['ASDF'].data.tobytes())
    with asdf.open(bs) as af:
        return af['meta']['wcsinfo']['s_region']
```

4) Using `astropy.io.fits` and `asdf.util.load_yaml`:
```python
with fits.open(fn) as ff:
    bs = io.BytesIO(ff['ASDF'].data.tobytes())
    tree = asdf.util.load_yaml(bs)
    return tree['meta']['wcsinfo']['s_region']
```

Of the 4 options above 1-3 are similar in performance (using both a `ImageModel` and a larger `IFUImageModel` as test files). With performance being limited primarily by `asdf.open` (more on that below). 4 is much faster in both cases. See the below table for performance (run with `cProfile` so slightly slower than real).

|  | ImageModel | IFUImageModel |
| ------- | ---------- | ------ |
| dm.open | 0.245 | 3.945 |
| asdf_in_fits | 0.137 | 3.815 |
| asdf.open | 0.143 | 3.820 |
| load_yaml | 0.036 | 0.332 |

The table shows it's ~10x faster to use `load_yaml` as this skips:
- validation
- mapping fits based on the schema
- loading arrays
- conversion from asdf flavored yaml to custom python objects

Although all of the above is public API it seems worthwhile to investigate wrapping 4 as a helper function in stdatamodels (with sufficient documentation about how this is dangerous).

Below is a snakeviz generated graph of the call to `dm.open` for the `IFUImageModel` data file showing the bulk of the time spent in `asdf.open`:
Screen Shot 2024-03-14 at 5 28 29 PM

emolter:
Resolves [JP-3882](https://jira.stsci.edu/browse/JP-nnnn)

Closes #284 

This PR adds infrastructure to allow datamodel metadata to be lazy-loaded, bypassing validation and avoiding loading any arrays into memory.

The script below demonstrates how this new infrastructure can be used to query CRDS.  It also serves as a test that this works as intended.  At some point, this can be modified to become a regression test in JWST:
```
import crds
from stdatamodels.jwst.datamodels import get_metadata
from stcal.testing_helpers import MemoryThreshold

# input file I tested this with takes up 2 GB of disk space
fname = "your_large_input_file.fits" # or ASDF

# MemoryThreshold is a shortcut to testing memory usage.
# It will raise an AssertionError when memory usage exceeded the threshold.
# modify threshold as needed to better test memory usage for your file
with MemoryThreshold("100 MB"):
    tree_flat = read_metadata(fname)
    ref_dict = crds.getreferences(tree_flat) # creates a reasonable dictionary of reference files
```

## Tasks

</details>


src/stdatamodels/jwst/datamodels/tests/test_load_meta.py
```diff
@@ -0,0 +1,52 @@
...
+    # Load the meta attribute from the file
+    meta_attr_foo = load_meta_attribute(path, "meta.foo")
+    meta_attr_foo_list = load_meta_attribute(path, ["meta", "foo"])
+    assert meta_attr_foo == "bar"
+    assert meta_attr_foo_list == "bar"
+
+    # Ensure meta attributes of data are accessible
+    # TODO: these come out differently for ASDF and FITS files
+    # data_shape = load_meta_attribute(path, "data.shape")
+    # assert shp == tuple(data_shape)
```
emolter, line 30: @braingram Out of curiosity, do you know offhand why this is the case?

emolter:
@tapastro @melanieclarke @nden Here is a first draft of an API for loading metadata attributes without loading the entire datamodel.  It would be good to define what this API should look like earlier rather than later.

Not necessarily looking for a detailed review now (although you're welcome to do so), I'm wondering primarily things like

- What do you think of having a separate method instead of somehow altering `datamodels.open` to support lazy loading? (The current proposal is most likely much easier to code).
- What other features would you like to see in this API?
- Should this be able to operate on already-open datamodels, so that we take advantage of the lazy-load aspect when a filename is input but just returns the requested attribute from the model when a JwstDataModel is input?
- Should this support accessing multiple keywords at once, or is it better to just require that the caller iterates over this method to produce their list of desired attributes? The latter requires loading the metadata once per iteration.  The former has a slightly more complicated API, and in practice might not gain much - for example, the `ModelContainer.models_grouped` use-case only requires two attributes.

src/stdatamodels/jwst/datamodels/tests/test_load_meta.py
braingram: Hmm... shape is an attribute of `ndarray`. Looking at the other code in this PR the FITS and ASDF data is handled different. The ASDF data is loaded with `asdf.open` (which fully deserializes the contents, so no "lazy loading" as far as this PR is concerned). The `_traverse_tree` attempts to access shape using __getitem__ rather than getattr (which causes a failure for me). The FITS data is loaded with `load_yaml` (but the other FITS headers and data in the file are ignored). `load_yaml` doesn't deserialize the contents returning the "array" as a dictionary which happens to have a "shape" key allowing this to pass.

emolter: > the FITS and ASDF data is handled different

I'm still very unfamiliar with how ASDF works so I'm not surprised I got this wrong.  Thanks in advance for being patient with me.  `fits.open` doesn't work on asdf data, so they have to be handled at least slightly differently, right?  Looking again it seems I could call `load_yaml` directly on the file, is that the right thing to do here?

>  loaded with asdf.open (which fully deserializes the contents, so no "lazy loading" as far as this PR is concerned).

I should have figured this out from your initial comments on the issue.  But shouldn't that mean that `load_meta_attribute(path, "data")` loads the data array itself, causing `assert getsizeof(data_meta) < getsizeof(data)` to cause an error?  Why doesn't it?

braingram: > Looking again it seems I could call `load_yaml` directly on the file, is that the right thing to do here?

It depends. It would be closer to what's done with the FITS file (where the ASDF content isn't deserialized). For the FITS data, what the plan for handling attributes that are in both the FITS and ASDF portion (like `s_region`)? Calling `load_yaml` on the ASDF extension ignores any possible edits that might have been made to the FITS headers.

> But shouldn't that mean that `load_meta_attribute(path, "data")` loads the data array itself, causing `assert getsizeof(data_meta) < getsizeof(data)` to AssertionError? Why doesn't it?

It looks like for the `asdf` parametrization `data` is a lazy loaded array (see lazy_load [here](https://www.asdf-format.org/projects/asdf/en/latest/api/asdf.AsdfFile.html#asdffile)). Since the `data_meta` isn't accessed in the test it isn't loaded (and `getsizeof` returns the expected smaller value for `data_meta`).

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -365,3 +367,178 @@ def is_association(asn_data):
...
+                hdulist, fits_keyword, schema, ctx.get("hdu_index"), known_keywords, hdu_cache
+            )
+
+            properties.put_value(path, result, tree)
+
+    mschema.walk_schema(schema, callback)
+    return tree
+
+
+def _lazy_load_asdf_from_schema(tree_in, schema):
```
braingram, line 413: I was initially confused why this was needed. I now see that the returned metadata is only the values mapped to FITS keywords in the schema (as described in the docstring for `get_metadata`). Keeping the FITS and ASDF results equivalent seems beneficial and making the FITS match the ASDF would require loading the ASDF extension and merging that tree with the FITS keywords.

If we were to want to do that (to support reading the ASDF only, non-FITS keyword mapped metadata) would the following work?
- For a FITS file, first read the ASDF extension to get the base `tree`, then pass that to `_lazy_load_fits_from_schema` (which will overwrite values with FITS keywords)
- For an ASDF file, just use load_yaml

One issue would be that this would bring through things like `meta.wcs` in it's serialized form (a dict or tagged dict). Would that cause issues for getting parameter reference files?

I'm not suggesting that we do this (as the current solution seems like it would cover all the use cases and is likely simpler than the approach above) but I did want to bring it up in case I'm not thinking of a simpler way to do it that keeps the non-FITS keyword mapped metadata.

braingram (commented):
Overall I think this looks great! The code is simple, documented, tested and effective (with the tests I've done so far) and looks like it would fit the use cases of:
- updating `ModelLibrary` to not use hard-coded FITS keywords
- updating `ModelContainer` to not load the file to compute the group_id (although I don't know if this is worth doing given the non-performance applications of `ModelContainer`)
- reading `s_region` efficiently in resample
- opening up the possibility of implementing a much more efficient mode for `get_crds_parameters` for `DataModel`, `ModelContainer` and `ModelLibrary` (with some stpipe changes to actually use that method)

Are there other use cases in mind? For these how do you propose we schedule those updates with this PR (do we finalize and merge this first, then work on those making follow up PRs as needed or start them in parallel?).

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -365,3 +367,178 @@ def is_association(asn_data):
...
+def _retrieve_schema(model_type):
+    """Load schema for the input model type."""
+    try:
+        schema_url = getattr(dm, model_type).schema_url
+    except AttributeError:
+        raise ValueError(f"Model type {model_type} not found.") from None
+    return asdf.schema.load_schema(schema_url, resolve_references=True)
+
+
+def get_metadata(fname, model_type=None, flatten=True):
```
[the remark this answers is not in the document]
emolter: `read_metadata` sounds good, or `load_metadata`?

I don't personally much like `read_fits_keywords` because it seems to imply that the return dict would be accessible via the fits keywords themselves.  While the difference between FITS-mapped metadata and other metadata is important internally to stdatamodels, I don't think most people (even other jwst devs) know/care about the difference - I certainly would not have understood the difference a few months ago.

src/stdatamodels/jwst/datamodels/util.py
braingram: Thanks. Given the lack of use case I'd say let's consider the naming as a way to make the current behavior (loading only FITS mapped metadata) as obvious as possible.

src/stdatamodels/jwst/datamodels/util.py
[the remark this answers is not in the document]
braingram: Thanks. `read_metadata` or `load_metadata` sound good to me.

src/stdatamodels/jwst/datamodels/util.py
emolter: sounds good.  To make the behavior abundantly clear, I added a line to the extended description of the docstring so it now reads:

```
    The output dictionary will contain every metadata attribute in the schema,
    even if it is not present in the datamodel. If the attribute is not present,
    the value will be None. Likewise, if the attribute is present in the datamodel
    but is not mapped to a FITS keyword by the schema, it will not be included
    in the output.
```

emolter:
a round of regtests starting here: https://github.com/spacetelescope/RegressionTests/actions/runs/13975536110

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
```diff
@@ -0,0 +1,91 @@
...
+    meta = read_metadata(path, flatten=False)
+    assert isinstance(meta, dict)
+    assert isinstance(meta["meta"], dict)
+
+    # Ensure the metadata has the expected attributes
+    assert meta["meta"]["instrument"]["filter"] == FILT
+    assert meta["meta"]["observation"]["date"] == INPUT_TIME
+
+
+def test_read_metadata_bad_inputs(make_and_save_models, tmp_path):
```
braingram, line 72:
```suggestion
def test_read_metadata_bad_inputs(tmp_path):
```

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+TESTFILE_ROOT = "jwst_image."
+FILT = "GR150R"
+INPUT_TIME = "2021-01-01 00:00:00.000"
+DATA = np.zeros((1000, 1000), dtype=np.float32)
+
+
+@pytest.fixture
+def make_and_save_models(tmp_path):
```
braingram, line 18:
```suggestion
@pytest.fixture
def make_and_save_models(tmp_path):
```
This could be a parametrized fixture which when used would paramerized the tests that use it. Something like:
```python
@pytest.fixture(params=["fits", "asdf"])
def saved_model_filename(request, fake_model,tmp_path):
    ....
    extension = request.param
```
Also if it returned the path `tmp_path` wouldn't need to be added to the tests that use this fixture.
Then `test_read_metadata` could be something like:
```python
def test_read_metadata(saved_model_filename):
```
without the need for parametrization or the other fixtures.

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
```diff
@@ -0,0 +1,91 @@
...
+    assert meta["meta.filename"] == root
+
+    # Ensure attributes in schema but not in model are there but give None
+    assert meta["meta.instrument.detector"] is None
+
+    # Ensure the metadata does not have data attributes
+    assert "data" not in meta
+    assert getsizeof(meta) < getsizeof(DATA)
+
+    # test flatten=False mode
```
braingram, line 62:
```suggestion
    # test flatten=False mode
```
Can we split this up into multiple tests. We don't need to go as far as 1 assert per test but fewer would be easier to understand what each assert is testing (based on the test name) and make refactoring easier.

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
```diff
@@ -0,0 +1,91 @@
...
+    assert meta["meta.instrument.filter"] == FILT
+    assert meta["meta.observation.date"] == INPUT_TIME
+    assert meta["meta.filename"] == root
+
+    # Ensure attributes in schema but not in model are there but give None
+    assert meta["meta.instrument.detector"] is None
+
+    # Ensure the metadata does not have data attributes
+    assert "data" not in meta
+    assert getsizeof(meta) < getsizeof(DATA)
```
braingram, line 60: This only measures the size of the `meta` dict. [Not the values](https://docs.python.org/3/library/sys.html#sys.getsizeof):
```python
>> d = {'a': 1}
>> sys.getsizeof(d)
184
>> d['a'] = np.zeros((10000, 10000))
>> sys.getsizeof(d)
184
```

braingram (commented):
Some minor code suggestions, docs suggestions and some questions/suggestions for the tests. If it's helpful I can make a PR for the test changes (they were too spread out to make inline suggestions). I believe the comments describe the general goal/idea but if examples would be helpful let me know.

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -365,3 +370,169 @@ def is_association(asn_data):
...
+    .. warning::
+
+        This function entirely bypasses schema validation. Although validation
+        is done when saving a datamodel to file, if a model is modified and then
+        saved with something other than datamodels.save (e.g. astropy.fits.writeto),
+        the schema will not be validated and invalid data could be loaded here.
+
+    Parameters
+    ----------
+    fname : str or Path, optional
```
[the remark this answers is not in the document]
braingram: Jokes aside this doesn't seem to support `Path`. When I try to provide one it fails `ValueError: Cannot get file type of ...` due to `filetype.check` not supporting `Path`.

Would you update the tests to check this and convert `Path` instances to `str` for `filetype.check` so at least `read_metadata` can support them.

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -365,3 +370,169 @@ def is_association(asn_data):
...
+            if isinstance(result, TaggedString):
+                result = str(result)
+            properties.put_value(path, result, tree)
+
+    mschema.walk_schema(schema, callback)
+    del tree_in
+    return tree
+
+
+def _to_flat_dict(tree):
```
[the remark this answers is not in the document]
emolter: This does indeed raise a RecursionError when it receives a recursive tree.  Do you suggest that I change this so it works with recursion, given that that's not currently encountered?

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -365,3 +370,169 @@ def is_association(asn_data):
...
+    tree = {}
+
+    def callback(schema, path, combiner, ctx, recurse):
+        """Ignore anything that is a data array"""
+        result = None
+
+        if "fits_keyword" in schema:
+            try:
+                # Traverse the tree to get the attribute
+                result = functools.reduce(operator.getitem, path, tree_in)  # noqa: F821
```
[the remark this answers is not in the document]
emolter: `Undefined name: tree_in`, although it goes away if I remove the `del` statement per your other comment

src/stdatamodels/jwst/datamodels/util.py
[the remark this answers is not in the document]
braingram: Thanks! I think it's fine as-is. The only place I recall seeing recursive trees in datamodels is in the wcs which won't make it here since it's not mapped to FITS headers.

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -365,3 +370,169 @@ def is_association(asn_data):
...
+    # hdulist.__getitem__ is surprisingly slow (2 ms per call on my system
+    # for a nirspec mos file with ~500 extensions) so we use a cache
+    # here to handle repeated accesses. A lru_cache around get_hdu
+    # was not used as hdulist is not hashable.
+    hdu_cache = {}
+
+    def callback(schema, path, combiner, ctx, recurse):
+        """Ignore anything that is a data array"""
+        result = None
+
```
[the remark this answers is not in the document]
emolter: Looks like no.  This was a partial copy of the callback function in `fits_support._load_from_schema` where this existed, but yeah it doesn't appear to do anything

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
emolter: whoops.

I wanted to use something like `stcal.testing_helpers.MemoryThreshold` but stcal is not a dependency of stdatamodels.  Does it feel like overkill to copy that entire context manager into here just for this test?

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -365,3 +370,169 @@ def is_association(asn_data):
...
+def _load_asdf_meta_from_schema(tree_in, schema):
+    """Constrain asdf tree to fits-mapped schema keywords only."""
+
+    tree = {}
+
+    def callback(schema, path, combiner, ctx, recurse):
+        """Ignore anything that is a data array"""
+        result = None
+
```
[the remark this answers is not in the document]
emolter: same as above

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
braingram: Probably. What's the goal of the `getsizeof` check? Is it to make sure that data isn't hidden somewhere in the meta dict? Or is the goal to check that `data` was never read during the call to `read_metadata`?

emolter: The plan was the second one, which I realize was not well thought out - even if the `getsizeof` check did check the memory usage of the entire contents of the dict, it wouldn't have ensured that it was never loaded and then discarded.  The desired use-case is indeed exactly what the `MemoryThreshold` context was built to solve.  But on the other hand, is this test necessary to include at all?

src/stdatamodels/jwst/datamodels/util.py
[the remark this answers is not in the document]
emolter: Should be fixed by https://github.com/spacetelescope/stdatamodels/commit/726264709bce7ba67d264d13e82cb55989b00717

braingram:
I tested this with a relatively "empty" ImageModel (`m = dm.ImageModel(shape=(100, 100))`) and as documented the metadata included all FITS mapped keywords as `None`. Is there a use case for this instead of say only returning those that are defined? I think the following diff would accomplish that:
```diff
--- a/src/stdatamodels/jwst/datamodels/util.py
+++ b/src/stdatamodels/jwst/datamodels/util.py
@@ -400,11 +400,17 @@ def _load_fits_meta_from_schema(hdulist, schema):
     def callback(schema, path, combiner, ctx, recurse):
         """Ignore anything that is not mapped to a fits header keyword."""

-        if "fits_keyword" in schema:
-            fits_keyword = schema["fits_keyword"]
-            result = fits_support._fits_keyword_loader(
-                hdulist, fits_keyword, schema, ctx.get("hdu_index"), known_keywords, hdu_cache
-            )
+        if fits_keyword := schema.get("fits_keyword"):
+            hdu_name = fits_support._get_hdu_name(schema)
+            try:
+                hdu = fits_support.get_hdu(hdulist, hdu_name, _cache=hdu_cache)
+            except AttributeError:
+                return
+
+            try:
+                result = hdu.header[fits_keyword]
+            except KeyError:
+                return

             properties.put_value(path, result, tree)

@@ -425,7 +431,7 @@ def _load_asdf_meta_from_schema(tree_in, schema):
                 # Traverse the tree to get the attribute
                 result = functools.reduce(operator.getitem, path, tree_in)
             except KeyError:
-                result = None
+                return
             if isinstance(result, TaggedString):
                 result = str(result)
             properties.put_value(path, result, tree)
```

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
```diff
@@ -0,0 +1,91 @@
...
+    meta = read_metadata(model_path, flatten=False)
+    assert isinstance(meta, dict)
+    assert isinstance(meta["meta"], dict)
+
+    # Ensure the metadata has the expected attributes
+    assert meta["meta"]["instrument"]["filter"] == FILT
+    assert meta["meta"]["observation"]["date"] == INPUT_TIME
+
+
+def test_read_metadata_bad_inputs(model_path):
```
braingram, line 79: Would you split up into 3 tests with the comments as titles and/or docstrings?

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
braingram: It does seem useful to check that `read_metadata` doesn't read the data. I was hoping truncating the files (so they contain the header but impartial data) might work (if `read_metadata` succeeds). However I haven't been able to get that to work.

We could monkeypatch parts of fits/asdf to make sure the data isn't called but that seems brittle. I wonder if we can track filesystem access. I know how to do that for a file object but since the input is a filename/Path I'm not sure how to get that to work. I'll keep poking at this but if we can't come up with a way to test it I think we can go forward with may be a test of memory usage (perhaps using psutil which is already a test dependency).

emolter:
> I tested this with a relatively "empty" ImageModel (m = dm.ImageModel(shape=(100, 100))) and as documented the metadata included all FITS mapped keywords as None. 

I think I'm happy with either behavior.  I had in mind some potential workflow where you want to, say, load all the S_REGIONs where they are defined and not cause errors where they are not defined, something like (pseudocode)
```
for filename in model_list:
    s_region = read_metadata(filename)["meta.wcsinfo.s_region"]
    if s_region is None:
        # load the model and call compute_s_region_imaging
```
There are ways around this though, like checking if the key is in the dict, which is (almost) just as easy

src/stdatamodels/jwst/datamodels/util.py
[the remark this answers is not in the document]
emolter: noticed I forgot to change the doc to remove `optional` but itll be in the next commit

src/stdatamodels/jwst/datamodels/tests/test_read_metadata.py
emolter: sounds good.  Is there anything wrong with having something similar to that stcal context manager?  It uses `tracemalloc` which is part of the standard library

braingram: Nothing wrong with that. I doubt psutil provides anything helpful here.

What do you think about this?
```python
def test_no_data(model_path, monkeypatch):
    def throw_error(self):
        raise Exception()

    if 'asdf' in str(model_path):
        import asdf
        monkeypatch.setattr(asdf, "open", throw_error)
        read_metadata(model_path)
    else:
        from astropy.io import fits
        monkeypatch.setattr(fits.ImageHDU, "data", property(throw_error))

        with fits.open(model_path) as hdulist:
            monkeypatch.setattr(fits, "open", lambda *args, **kwargs: hdulist)
            read_metadata(model_path)
```
It's only patching public API so it hopefully shouldn't be too brittle :-/

braingram:
> > I tested this with a relatively "empty" ImageModel (m = dm.ImageModel(shape=(100, 100))) and as documented the metadata included all FITS mapped keywords as None.
> 
> I think I'm happy with either behavior. I had in mind some potential workflow where you want to, say, load all the S_REGIONs where they are defined and not cause errors where they are not defined, something like (pseudocode)
> 
> ```
> for filename in model_list:
>     s_region = read_metadata(filename)["meta.wcsinfo.s_region"]
>     if s_region is None:
>         # load the model and call compute_s_region_imaging on the wcs
> ```
> 
> There are ways around this though, like checking if the key is in the dict, which is (almost) just as easy

Thanks. I think both are workable. To most closely match the current `get_crds_parameters` I vote for only including defined metadata.

emolter:
Ok I went ahead and implemented that test and the suggested diff, almost exactly as you wrote them.  Thanks for making it easy for me!

emolter:
rerunning JWST regtests after relatively substantial changes: https://github.com/spacetelescope/RegressionTests/actions/runs/13975536110

braingram (approved):
LGTM!

emolter:
Thanks.  I'm going to wait and give other reviewers a chance to take a look before merging.

nden:
Should this work with `MultiSlit` model? I am getting an error
```
File ~/opt/anaconda3/envs/b11.1/lib/python3.11/site-packages/stdatamodels/properties.py:563, in put_value(path, value, tree)
    561             cursor = cursor.setdefault(part, [])
    562         else:
--> 563             cursor = cursor.setdefault(part, {})
    565 if isinstance(path[-1], int):
    566     while len(cursor) <= path[-1]:

AttributeError: 'list' object has no attribute 'setdefault`
```

emolter:
> Should this work with MultiSlit model?

It should, but I didn't test it.  I'm glad you did!  I'll poke around and see if I can figure out how to fix this

emolter:
@nden something like https://github.com/spacetelescope/stdatamodels/pull/445/commits/be66c62724a54d4c1e5971c69bfdbc26ccfaa602 appears to work, although I modified `properties.put_value` to add special handling for lists.  Would you please test it out?  I tried it on the new unit test mock MultiSlitModel as well as a MultiSpecModel from our regtest data (`truth/test_miri_lrs_slit_spec3/jw01530-o005_t004_miri_p750l_x1d.fits`)

Perhaps @braingram can suggest a better way of doing this though.

emolter:
Is the goal to make an identical dictionary to what we get from `get_crds_parameters` though, or to have the actual query to CRDS return identical results?  I'm pretty sure there are many parameters that you get from `get_crds_parameters` but are not needed to query CRDS.  I'm guessing that includes everything under `slits`.

braingram:
> Is the goal to make an identical dictionary to what we get from `get_crds_parameters` though, or to have the actual query to CRDS return identical results? I'm pretty sure there are many parameters that you get from `get_crds_parameters` but are not needed to query CRDS. I'm guessing that includes everything under `slits`.

I think identical would be safest (since as far as I know we're planning to use `read_metadata` and `get_crds_parameters` interchangeably) so likely a good goal. I'm not aware of anything that uses a slit metadata value but I do wonder if a [CRDS wildcard](https://hst-crds.stsci.edu/static/users_guide/rmap_syntax.html#wild-cards) would allow using it (`slits.*.foo`) to say select a reference file if any slit in a file has a given parameter. If we make the decision here to silently drop the per-slit metadata if someone starts using a wildcard they'll see a different reference file if they use an open model vs a filename.

emolter:
I agree, you're probably right it's safer.  I'll look into ways to do this.

emolter:
ok @braingram @nden I made another version of this that returns to using `load_yaml` on the `ASDF` extension of the FITS file.  This makes it easier to read slits from a `MultiSlitModel`.  The code at the most recent commit (https://github.com/spacetelescope/stdatamodels/pull/445/commits/35a73884b06ce8f7656ab517ffa02acc259feed9) behaves as follows:

- For FITS files, read the ASDF tree using `load_yaml`, then walk the schema and update the tree with the attributes found in the fits header (i.e., do a "fits update")
- Do not attempt to do a "fits update" on `model.slits` or any model metadata attribute that is list-like.
- For ASDF files, simply read the tree using `load_yaml` and do nothing except convert tagged strings to strings
- Custom-ignore the WCS because it's recursive (and is the only recursive attribute of model metadata we can foresee existing). This matches `model.get_crds_parameters()`, which also does not have a `meta.wcs` keyword.
- Add a test ensuring every key, val pair returned by `model.get_crds_parameters()` matches a key, val pair from `read_metadata()`.

I tested the performance, and this performs (predictably) very similarly to the `load_yaml` case shown in https://github.com/spacetelescope/stdatamodels/issues/284#issuecomment-2733454863.

Let me know what changes you'd like to see made.

braingram:
Thanks for making these updates.

I did some more testing with this and found a few things.

If I save a MultiSlitModel with 1 slit named "a" then rename it to "b" (using astropy.io.fits) `read_metadata` returns `slits.0.name == 'a'`.

I think this is explained by:
>     * Do not attempt to do a "fits update" on `model.slits` or any model metadata attribute that is list-like.

If I save an ImageModel with a data array as ASDF and read it back in with `read_metadata` it includes:
```
 'data.source': 0,
 'data.datatype': 'float32',
 'data.byteorder': 'big',
```

That's explained by:
>     * For ASDF files, simply read the tree using `load_yaml` and do nothing except convert tagged strings to strings

In the current state of the PR `_load_fits_meta_from_schema` almost matches [_load_from_schema](https://github.com/spacetelescope/stdatamodels/blob/109991c45b3aa7652d2c9878a6d055a63d8c5dbe/src/stdatamodels/fits_support.py#L577) (and `_load_from_schema` handles list-like items). I'd like to test this out a bit more to understand how these relate (and if we can use `_load_from_schema` with some modification).

emolter:
Thanks for doing some testing, I'm glad you're helping me through this one, which is a bit more complicated than I initially thought.

I wasn't sure of the best way to attempt a "fits update" within the slits, so let me know if you have suggestions.

RE the extra data-related meta, I was aware of this but I don't consider it to be a problem for the currently known use-cases.

Good to know RE `_load_from_schema`, I did take code snippets from other various places around the repo but did not realize how similar these had become.  I'm curious to hear what you can figure out.

emolter:
Hi @braingram I'm back from vacation now.  Can you remind me what you feel is still needed in order to get this over the line?

braingram:
> Hi @braingram I'm back from vacation now. Can you remind me what you feel is still needed in order to get this over the line?

Unfortunately I haven't had the time to dig further into `_load_from_schema`. Since the aim of the new API is to replace some uses of `datamodels.open` with something more efficient I think having something that:
- returns the save values as what would be returned with `datamodels.open`
- does so with less resources

would "fit the bill".

If I recall correctly the current PR sometimes returns different values than `datamodels.open`. If this PR was updated to return the same values more efficiently (with hopefully little code duplication) then replacing uses of `datamodels.open` with the new API could be started. At the moment my concerns are:
- the different results could lead to difficult bugs. For public API it would be easiest to explain if the results are identical (but validation was bypassed). 
- understanding the "in practice" performance impact of the new API. For example, for operations most runs will be "in memory" so the new API won't be used for reading "s_region" in resample or "group_id". This leaves the "crds_parameters" use case. I believe this happens once per pipeline run for 1 model and at the start where the "model" is still coming from a file (correct me if I'm wrong). So I think we're weighting the new API/code against saving a few seconds for each pipeline run.

emolter:
Hi Brett,

It was not my understanding that we were planning to replace the majority, or even a large subset, of calls to `datamodels.open()` with this new functionality.  In my view, that would be equivalent to saying that we don't really care if validation happens or not.

I see this additional functionality as a way to extract certain specific metadata keywords from a model without fully loading it. We could give `datamodels.open` a lazy-loading feature, but as I recall, we sort of already had that discussion on the GitHub issue and rejected it because smart memory handling is supposed to be the job of ModelLibrary.

> For public API it would be easiest to explain if the results are identical (but validation was bypassed).

The results are already not going to be identical because we aren't loading any of the data arrays.  Which results need to be identical that currently are not?  I can see the argument that I should find a way to ensure that a fits_update happens to `slits` and other list-like elements that are NOT data arrays, although I am still unsure how to approach that in the best way.  But that's the only "non-identical" result I see.  (I don't really care about e.g. `data.datatype` because it's not part of the metadata via `datamodels.open`).

> understanding the "in practice" performance impact of the new API

I'm not sure we have good data on how many users might be using the `in_memory=False` processing yet.  There has been some more discussion about _a-priori_ guessing the memory usage of calwebb_image3 (https://jira.stsci.edu/browse/JP-3707) and I think that's still a useful thing to support eventually.  We also discussed whether this can be used by ops to categorize data, or internally by CRDS - I don't know the answer but it's a good candidate.  Even with all of those aside, I think the improvement in getting CRDS parameters is worth doing, especially because we already have a prototype that more-or-less works.

But if you're now saying "we shouldn't do this at all" then I'll defer that discussion to @nden, @tapastro, and @melanieclarke.  It might be good to loop them back into this discussion at this stage anyway to get some additional opinions on how this should look.

I can go ahead and look further into the differences between `_load_from_schema` and `_load_fits_meta_from_schema`.

melanieclarke:
I think we definitely want this.  I agree that we are not aiming to replace most datamodels.open calls in the pipeline usage; we are looking for a tool that is useful in some situations for both pipeline and users, like retrieving CRDS reference files or checking appropriate keywords before fully opening a file.

It looks to me like the current solution is very close to what we're looking for, and it's worth pursuing the fits_update behavior for the slits.  Like Ned, I'm unconcerned about the extra keywords in the ASDF case.

emolter:
Ok I made some more changes to this:

- Add options to `_load_from_schema` to get the desired behavior, instead of repeating code in an entirely new helper class
- Do a `fits_update` on list-like attributes that themselves have metadata (e.g. model.slits and model.spec): this was handled already by the original `_load_from_schema`. 
- Add a unit test that ensures the fits_update happens for slit names in a MultiSlitModel

Please test it out and let me know how it looks to you @braingram @melanieclarke 

The output from ASDF files and FITS files is still discrepant for the same reason as before, namely that the tree will contain e.g. `"data.datatype"` for ASDF files.  If there's a simple way to fix this I'm happy to do so, but as discussed before I don't feel as though this discrepancy is a serious issue.

braingram (approved):
LGTM. Would you rerun the regtests before merging.

melanieclarke:
From a couple quick tests, it looks like this works great.  One weird thing - I noticed that if flatten is False, I can access the cal logs with:
`read_metadata(fname, flatten=False)['cal_logs']`
but the log is not present with `read_metadata(fname, flatten=True)`.

emolter:
> From a couple quick tests, it looks like this works great. One weird thing - I noticed that if flatten is False, I can access the cal logs with: `read_metadata(fname, flatten=False)['cal_logs']` but the log is not present with `read_metadata(fname, flatten=True)`.

Thanks for testing this out!  Good catch on the cal_logs, this revealed an error in the logic of `_to_flat_dict()`.  I just fixed it and added a unit test to cover the case.  Now the cal_logs are accessible via e.g. `cal_logs.extract_1d.1` for the first line, `cal_logs.extract_1d.2` for the second line, etc.  Perhaps the line indexing feels a bit silly but it's logically consistent with the rest of the flattened dict and I think it's best to ensure that the information is all there.

emolter:
regression test rerun https://github.com/spacetelescope/RegressionTests/actions/runs/14621976257

melanieclarke:
Yes, I see the logs now.  Thanks!  

> Perhaps the line indexing feels a bit silly but it's logically consistent with the rest of the flattened dict and I think it's best to ensure that the information is all there.

The line indexing makes it really awkward to retrieve and use the log from the flat dictionary, but I guess in that case, you should just use the nested dictionary version.  Maybe we need an additional small utility function to return the log as printable string?

emolter:
> Maybe we need an additional small utility function to return the log as printable string?

What is your proposed use-case?  Does that need to be done now?  And would you want to add custom handling for `cal_logs` within `read_metadata` so that it is returned as a formatted string?

melanieclarke:
> > Maybe we need an additional small utility function to return the log as printable string?
> 
> What is your proposed use-case? Does that need to be done now? And would you want to add custom handling for `cal_logs` within `read_metadata` so that it is returned as a formatted string?

It doesn't need to be done now.  I was just thinking it would be useful to do something like `my_log = read_log(fname, pipeline="calwebb_detector1")`, and then be able to print it or check it for some substring I'm interested in.  With the current implementation, I can do the equivalent with 
`my_log = "\n".join(read_metadata(fname,flatten=False)['cal_logs']['calwebb_detector1'])`
or, much more awkwardly with the default flat dictionary:
```
tree = read_metadata(fname)
log_strings = []
for key, value in tree.items():
    if key.startswith('cal_logs.calwebb_detector1'):
        log_strings.append(value)
my_log = '\n'.join(log_strings)
```

Custom handling within read_metadata would work too.

emolter:
restarted the regtest run https://github.com/spacetelescope/RegressionTests/actions/runs/14621976257

melanieclarke:
Thank you - `read_metadata(fname)['cal_logs']` now looks perfect.  I think everything looks good now, pending clean regtests.  Thanks for adding all the unit tests!

melanieclarke (approved):

emolter:
Tyler said on Slack that this looks fine, going to go ahead and merge

PR #451, opened 2025-03-17
TST: remove SlitDataModel

braingram:

Closes #

This PR addresses ...

## Tasks

</details>


emolter:
Paging @melanieclarke @jemorrison the spectroscopy experts on Chartreuse.  Do you think a deprecation period is warranted for this unused datamodel type, or do you think it's safe to just remove?  (Melanie is on vacation so this may need to sit for a week.)

braingram:
Thanks! No rush on my end for this one. I was just looking at schemas for other reasons and was confused by `SlitModel` and `SlitDataModel`.

jemorrison:
I believe that the SlitModel is used in  the MultiSlitModel.
I do not remember why we have both SlitModel and SlitDataModel 
But a MultiSlit mode is a set of SlitModels. 
I think Howard might know why we have both
@hbushouse

braingram:
> I believe that the SlitModel is used in the MultiSlitModel. I do not remember why we have both SlitModel and SlitDataModel But a MultiSlit mode is a set of SlitModels. I think Howard might know why we have both @hbushouse

Thanks. Yeah `SlitModel` is used throughout jwst. `MultiSlitModel` is weird. `jwst-pipeline/dev/nirspec/fs/jw01309022001_04102_00001_nrs1_cal.fits` open as a `MultiSlitModel` and has 4 slits:
```python
>> len(m.slits)
4
```
Indexing a slit via `slits` returns an `ObjectNode`
```python
>> m.slits[0]
<stdatamodels.properties.ObjectNode at 0x10a575880>
```
Indexing a slit via `__getitem__` returns a `SlitModel`:
```python
>> m[0]
<SlitModel(45, 1222) from jw01309022001_04102_00001_nrs1_cal.fits>
```
Internally the `slits` are `AsdfObject` instances:
```python
>> m._instance['slits'][0]
asdf.tags.core.AsdfObject
```
`SlitDataModel` is used is temporarily (on main)in the `MultiSlitModel` constructor if created with a `SlitModel`:
https://github.com/spacetelescope/stdatamodels/blob/438e4af208ea3d28f9fb95653814d2216f5b209f/src/stdatamodels/jwst/datamodels/multislit.py#L75-L83
However since `slits` is an attribute of `MultiSlitModel` the append casts the created `SlitDataModel` to the `MultiSlitModel` schema defined value (so the `SlitDataModel` only exists for lines 79 and 80). This PR switches that usage to `SlitModel` which is almost identical (`SlitModel` includes `int_times`, `SlitDataModel` does not but neither creates a default if one doesn't exist).

melanieclarke:
This looks fine to me.  I don't see any reason for a deprecation period, since there are no data products that correspond to SlitDataModel specifically and there's no reason a user would directly instantiate it.

nden:
The reason for `SlitDataMOdel` was to generate an array of slit data without actually validating the MultiSlit model. In the past (this may have changed now) adding a Slit model triggered validation of the whole Multislit model and so it took more than 24 hours to run calspec2. I haven't followed develpment but removing this needs to be tested on a largish MultiSlit Molde, i.e. ~ 200  slits

braingram:
Converting to draft until the large MultiSlitModel tests can be run.

braingram:
On stdatamodels main running the following code and tracking calls to `asdf.schema.validate`
```python
>> m = dm.MultiSlitModel()  # A
>> m.slits.append(m.slits.item())  # B
>> m.slits.append(m.slits.item())  # C
```
results in:
- A: 2x validation
  - Validation of `meta.model_type` [assignment](https://github.com/spacetelescope/stdatamodels/blob/4c3a3d0ca854697fcc219be3387f83f99c4dacd8/src/stdatamodels/model_base.py#L497).
  - Validation of `meta.date` [assignment](https://github.com/spacetelescope/stdatamodels/blob/4c3a3d0ca854697fcc219be3387f83f99c4dacd8/src/stdatamodels/jwst/datamodels/model_base.py#L45).
- B: 2x validation
  - Validation of the `ObjectNode` created in [item()](https://github.com/spacetelescope/stdatamodels/blob/4c3a3d0ca854697fcc219be3387f83f99c4dacd8/src/stdatamodels/properties.py#L501)
  - Validation of node added to `slits` during the [append](https://github.com/spacetelescope/stdatamodels/blob/4c3a3d0ca854697fcc219be3387f83f99c4dacd8/src/stdatamodels/properties.py#L457)
- C: 2x validation (same as B)

The same is true for this PR. I'll look at generating a 200 slit model next.

PR #409, opened 2025-02-21
remove unused dqflag functions

braingram:
Closes #408 

Requires: https://github.com/spacetelescope/jwst/pull/9212

This PR cleans up `stdatamodels.dqflags` removing the submodule as all functions are unused. Removing `dqflags` removes the last use of `basic_utils` and `basic_utils` also removed in this PR. Finally the removed functions are also removed from `stdatamodels.jwst.dqflags.__all__`.

Regtests with https://github.com/spacetelescope/jwst/pull/9212
https://github.com/spacetelescope/RegressionTests/actions/runs/13465093241

## Tasks
- [x] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [x] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [x] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
  - [x] [run `jwst` regression tests](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) with this branch installed (`"git+https://github.com/<fork>/stdatamodels@<branch>"`)

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details>

PR #459, opened 2025-03-28
Revert "remove unused dqflag functions"

braingram:
Reverts spacetelescope/stdatamodels#409

### DIFF
--- a/tests/test_dq.py
+++ b/tests/test_dq.py
@@ -1,29 +1,6 @@
-"""JWST Data Quality Flags
+from stdatamodels import dqflags
 
-The definitions are documented in the JWST RTD:
-
-https://jwst-pipeline.readthedocs.io/en/latest/jwst/references_general/references_general.html#data-quality-flags
-
-
-Implementation
--------------
-
-The flags are implemented as "bit flags": Each flag is assigned a bit position
-in a byte, or multi-byte word, of memory. If that bit is set, the flag assigned
-to that bit is interpreted as being set or active.
-
-The data structure that stores bit flags is just the standard Python `int`,
-which provides 32 bits. Bits of an integer are most easily referred to using
-the formula `2**bit_number` where `bit_number` is the 0-index bit of interest.
-"""
-
-# These imports are here for backwards compatibility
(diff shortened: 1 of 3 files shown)


tests/test_dq.py
```diff
@@ -0,0 +1,43 @@
...
+    "UNRELIABLE_DARK": 2**23,  # Dark variance large
+    "UNRELIABLE_SLOPE": 2**24,  # Slope variance large (i.e., noisy pixel)
+    "UNRELIABLE_FLAT": 2**25,  # Flat variance large
+    "OPEN": 2**26,  # Open pixel (counts move to adjacent pixels)
+    "ADJ_OPEN": 2**27,  # Adjacent to open pixel
+    "UNRELIABLE_RESET": 2**28,  # Sensitive to reset anomaly
+    "MSA_FAILED_OPEN": 2**29,  # Pixel sees light from failed-open shutter
+    "OTHER_BAD_PIXEL": 2**30,  # A catch-all flag
+    "REFERENCE_PIXEL": 2**31,  # Pixel is a reference pixel
+}
```
tapastro, line 37: This should probably be deleted, it's redundant (and outdated) to the listing provided in dqflags.

emolter: should be fixed by https://github.com/spacetelescope/stdatamodels/pull/459/commits/874b0986ddeb9817c812b2d67cbb033b0325d7ee

emolter:
What should be done about the changelog entry in this case?  As it stands, the patch release would have an entry both for removing these and restoring them.  Should the changelog fragment from the original (reverted) change be removed instead of adding one for this PR?

tapastro (approved):
LGTM

emolter:
> I think it would be fine to remove the old fragment and add a fragment for this PR detailing the remaining changes.

Can do, but the only meaningful change (c.f. prior to the original removal) would be to add the API for these DQ flags to the docs, is that what I should put in the changelog entry?

tapastro:
Ah, good point, the build process consumed all of the fragments... I guess the simple path is just to list the revert; I'm fine with that.

emolter:
sounds good, in that case I will leave as-is and we can go ahead and merge this

emolter (approved):

PR #465, opened 2025-04-22
JP-1028: Propagate slit name through NIRSpec WCS

melanieclarke:
Resolves [JP-1028](https://jira.stsci.edu/browse/JP-1028)

Requires https://github.com/spacetelescope/jwst/pull/9404

Update NIRSpec WCS transforms to propagate the slit name throughout.

Changes include:
- Inverse transforms for Slit2Msa and Gwa2Slit
- Change outputs for Slit2Msa to include slit name
- Add a legacy model to handle old output structure for Slit2Msa
- Add a slit ID to the Slit tuple, to allow NIRSpec FS slits to be uniquely identified with a numerical value (slit.name is a string, which can't be passed through the WCS transforms).

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/transforms/models.py
+++ b/src/stdatamodels/jwst/transforms/models.py
@@ -9,7 +9,9 @@ registered with ASDF through entry points.
 """
 
 import math
+import warnings
 from collections import namedtuple
+
 import numpy as np
 from astropy.modeling.core import Model
 from astropy.modeling.parameters import Parameter, InputParameterError
@@ -21,6 +23,7 @@ from ...properties import ListNode
 __all__ = [
     "Gwa2Slit",
     "Slit2Msa",
+    "Slit2MsaLegacy",
     "Logical",
     "NirissSOSSModel",
     "Slit",
@@ -441,9 +444,10 @@ class Gwa2Slit(Model):
             Wavelength.
         """
         index = self.slit_ids.index(name)
-        return (name, ) + self.models[index](x, y, z)
+        return (name,) + self.models[index](x, y, z)
 
     def inverse(self):
+        """Create an inverse model."""
         inv_models = [m.inverse for m in self.models]
         return Slit2Gwa(self.slits, inv_models)
 
@@ -465,6 +469,7 @@ class Slit2Gwa(Model):
         List of models (`~astropy.modeling.core.Model`) corresponding to the
         list of slits.
     """
+
     _separable = False
 
     n_inputs = 4
@@ -480,10 +485,10 @@ class Slit2Gwa(Model):
 
         self.models = models
         super(Slit2Gwa, self).__init__()
-        self.outputs = ('name', 'angle1', 'angle2', 'angle3')
-        """ Name of the slit and the three angle coordinates at the GWA going from detector to sky."""
-        self.inputs = ('name', 'x_slit', 'y_slit', 'lam')
-        """ Name of the slit, x and y coordinates within the virtual slit and wavelength."""
+        self.outputs = ("name", "angle1", "angle2", "angle3")
+        """Slit name and the three angle coordinates at the GWA going from detector to sky."""
+        self.inputs = ("name", "x_slit", "y_slit", "lam")
+        """Slit name, x and y coordinates within the virtual slit and wavelength."""
 
     @property
     def slits(self):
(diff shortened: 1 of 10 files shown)


melanieclarke:
JWST regtests with this branch + the branch that uses the new transforms: https://github.com/spacetelescope/RegressionTests/actions/runs/14778390226

JWST regtests with this branch + main:
https://github.com/spacetelescope/RegressionTests/actions/runs/14778426018

emolter (commented):
This is a part of the repository I'm still quite unfamiliar with, so I'll defer more detailed review to someone else like Nadia or Brett.

Is it possible to add some unit tests?  It seems to me that with the addition of these reverse transforms, some round-trip tests could be tractable? e.g. https://github.com/spacetelescope/stdatamodels/blob/cabdf76b06e4a92d229d8b3c1ee0d02105021cd0/src/stdatamodels/jwst/transforms/tests/tests.py#L37

src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-1.1.0.yaml
```diff
@@ -32,12 +32,26 @@ tags:
     It maps slit to the transform from the Grating Wheel Assembly (GWA)
     to the coordinate frame of the slit, where (0, 0) is the center of
     the slit.
+- tag_uri: "tag:stsci.edu:jwst_pipeline/slit_to_gwa-1.0.0"
```
braingram, line 35: Would you put these changes in a new manifest file (1.2.0) that is:
- contains all the 1.1.0 tag definition (started as a copy of 1.1.0)
- has these new tag definitions added
- has the extension_uri, id and title updated to 1.2.0

This is needed for backwards compatibility. Old software with the old 1.1.0 manifest will not be aware of these new tags. If they attempt to open a file generated with these tags (if they're added to the 1.1.0 manifest) the software will think it knows all the tags for 1.1.0 (the 1.1.0 version uri is stored in the file header) but then it will encounter unknown tags (slit_to_gwa-1.0.0) and not load correct (and produce less helpful warnings).

melanieclarke:
> Is it possible to add some unit tests? It seems to me that with the addition of these reverse transforms, some round-trip tests could be tractable?

Should be doable.  I'll look into it.

src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-1.1.0.yaml
melanieclarke: Will do, thanks for catching this.

melanieclarke:
@emolter - I added some unit tests to verify the changes to inverses and input/output arrangement.  The transforms are trivial, but I think that's okay for this use case - the real ones are tested in the jwst unit and regression tests.  Let me know if there are more you think I should add.

nden (approved):
LGTM. I tested with old and new files and results are the same.

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -443,9 +455,75 @@ def evaluate(self, name, x, y, z):
...
+        else:
+            self._slits = list(slits)
+            self.slit_ids = self._slits
+
+        self.models = models
+        super(Slit2Gwa, self).__init__()
+        self.outputs = ("name", "angle1", "angle2", "angle3")
+        """Slit name and the three angle coordinates at the GWA going from detector to sky."""
+        self.inputs = ("name", "x_slit", "y_slit", "lam")
+        """Slit name, x and y coordinates within the virtual slit and wavelength."""
```
emolter, line 507: very minor comment but shouldn't inline comments like these be demarcated by `#` instead of triple-quoted?

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -534,6 +619,169 @@ def evaluate(self, name, x, y):
...
+            "quadrant", "source_id", "shutter_state", "source_name",
+            "source_alias", "stellarity", "source_xpos", "source_ypos",
+            "source_ra", "source_dec")
+        models : list
+            List of models (`~astropy.modeling.core.Model`) corresponding to the
+            list of slits.
+        """
+        super(Slit2Msa, self).__init__()
+        self.inputs = ("name", "x_slit", "y_slit")
+        """ Name of the slit, x and y coordinates within the virtual slit."""
```
emolter, line 649: same here

emolter (approved):
As I said before I'm not the expert here but just skimming the code the changes make sense to me

src/stdatamodels/jwst/transforms/models.py
melanieclarke: I think this was intended to be a docstring for the attribute, in this style:
https://peps.python.org/pep-0258/#attribute-docstrings

We don't actually build the documentation for the models in this file, but testing it out locally, it looks like Sphinx doesn't render the docstring for the inputs and outputs attributes formatted in this way.  If we do get around to building docs for these models, we should move these doctrings to an Attributes section. For now, I think they can stay where they are, but let me know if you disagree.

emolter: I'm fine either way thanks

emolter: and I should have said, thanks for linking the PEP I was not aware of it!

braingram:
@melanieclarke is this waiting on anything? There is some other transform work that will conflict with this that I'd like to get started on.

melanieclarke:
@braingram - this is waiting on approval for the JWST PR, but I think we are close to being able to close it out.

braingram (commented):
LGTM. If you'd like to increase the coverage for the converters the unit tests could include something like:
```
asdf.AsdfFile({"model": gwa2slot}).write_to(tmp_file)
with asdf.open(tmp_file) as af:
    af["model"]  # ... tests that it round tripped
```

melanieclarke:
> For `Slit2Gwa` it looks like that isn't serialized with the current unit tests (and perhaps never serialized) since the inverse is not serialized for `Gwa2Slit` (and is instead re-computed when the inverse of `Gwa2Slit` is accessed). Is there a use case where `Slit2Gwa` will be serialized?

None that I know of.  I suppose it's possible that someone doing deep diagnostic work on the NIRSpec WCS might want to try it.

PR #466, opened 2025-04-25
JP-3931: New data format for WFSS extracted spectra

emolter:
Partially resolves [JP-3931](https://jira.stsci.edu/browse/JP-3931)
Partially resolves [JP-3751](https://jira.stsci.edu/browse/JP-3751)

This PR provides datamodels to represent all 1-D spectra extracted from a WFSS observation in a single file.  Specifically, the following models are added:
- WFSSMultiSpecModel: contains a single binary table that can hold spectra from multiple sources in a single exposure
- WFSSMultiExposureSpecModel: contains one WFSSMultiSpecModel-like binary table per exposure. This is the new `_x1d` file format for WFSS.
- WFSSMultiCombinedSpecModel: contains a single binary table that holds combined spectra from multiple sources.  The format of this model is similar to WFSSMultiSpecModel but with different columns because the spectra have been exposure-averaged.  This is the new `_c1d` file format for WFSS.

The WFSSMultiExposureSpecModel also contains the two new metadata columns requested by JP-3751.

Corresponding JWST PR is here: https://github.com/spacetelescope/jwst/pull/9402 

## Tasks

</details>


melanieclarke (commented):
Looks good! I noticed a couple typos.

src/stdatamodels/jwst/datamodels/multispec.py
```diff
@@ -107,3 +107,29 @@ def __init__(self, init=None, **kwargs):
             return
 
         super(TSOMultiSpecModel, self).__init__(init=init, **kwargs)
+
+
+class WFSSMultiSpecModel(JwstDataModel):
```
melanieclarke, line 112: I'm trying to understand this structure, now that it's expected that there will be multiple spectral tables output from extract1d for WFSS, one per spectral order.

I think it would be more analogous to the spectral structure for other modes if this model was a WFSSSpecModel, like a SpecModel.  That model is expected to contain only a spec_table and some metadata.  Then, what you currently have as WFSSMultiExposureSpecModel could be called WFSSMultiSpecModel, like a MultiSpecModel.  That model is already expected to contain a list of spec_table + metadata.

emolter: The logic behind it was that there are, well, multiple spectra included in it.  But I see your point that it has a `spec_table` accessible in the same place as the other SpecModels.  I will change it

emolter: see https://github.com/spacetelescope/stdatamodels/pull/466/commits/978c8e112313326a90a58c6245dddc1b5e8c8b24

src/stdatamodels/jwst/datamodels/spec.py
```diff
@@ -59,3 +59,29 @@ class TSOSpecModel(JwstDataModel):
 
     def get_primary_array_name(self):  # noqa: D102
         return "spec_table"
+
+
+class WFSSSpecModel(JwstDataModel):
+    """
+    A data model for NIRCam andNIRISS WFSS 1D spectra.
```
melanieclarke, line 66:
```suggestion
    A data model for NIRCam and NIRISS WFSS 1D spectra.
```

src/stdatamodels/jwst/datamodels/spec.py
```diff
@@ -59,3 +59,29 @@ class TSOSpecModel(JwstDataModel):
...
+
+    This model differs from the other MultiSpecModel classes in that
+    it is designed to hold all the spectra from a single WFSS exposure
+    in a "flat" table format. Therefore, it does not have the `spec` attribute
+    that is present in the other MultiSpecModel classes. Instead, it has
+    a `spec_table` attribute that contains the spectral data and metadata
+    for all sources in the exposure.
```
melanieclarke, line 73: Docs need updating, since this moved out of the multispec module.

melanieclarke (commented):
Thanks for the updates!

Just to note - the change to the TSO model introduces an error in the jwst downstream, so merging this will need to coordinate with merging the jwst PR.

emolter:
Regression tests using JWST are here: https://github.com/spacetelescope/RegressionTests/actions/runs/15124647542

I just kicked off a re-run but the previous run should be mostly representative, too.

There are several failures expected:

- A single keyword difference for x1dints files from spectroscopic TSO data, due to the change from `NELEMENTS` to `N_ALONGDISP`
- Failures for all NIRCam and NIRISS WFSS x1d and c1d files
- Failure to find any cal files coming out of spec3 for NIRCam and NIRISS WFSS data (spec2 cal files should still be created)
- Failures in asn generation for NIRCam and NIRISS WFSS spec3 associations (the product name has changed, although I'm not sure about the current state of how that type of failure will show up in the report)

emolter:
@braingram @tapastro This is ready now.  Could one of you skim it?  I don't think it's necessary (unless you're keen) to do a super detailed review, since we've been iterating a lot with the INS teams already and Melanie has seen it more than once.  However it's a big enough PR that I'd feel more comfortable if one more set of eyes were on it.

src/stdatamodels/jwst/datamodels/schemas/combinedspec.schema.yaml
```diff
@@ -15,6 +15,12 @@ allOf:
       default: 1
       fits_keyword: SPORDER
       fits_hdu: COMBINE1D
+    dispersion_direction:
+      title: Dispersion direction
+      type: integer
+      fits_keyword: DISPAXIS
+      fits_hdu: COMBINE1D
+      default: -999
```
braingram, line 23: What does "-999" do here and why not leave off the default (if it's a nonsense value)? Also, is the reasoning for this "-999" the same as for the other "-999" in the other schemas for this PR?

braingram (commented):
I gave it a skim. Overall LGTM. One question about the "-999" defaults.

src/stdatamodels/jwst/datamodels/schemas/combinedspec.schema.yaml
emolter: I think you are right that these are never used and should be removed where they have been added for metadata.  In fact, it may be worth considering whether to remove the default for exposure times and things too.

The defaults are used nontrivially when they exist in the spec_table attributes, because they populate the table where there is no data.  But here, there's either a global value for the whole dataset, or there isn't.

emolter: not special I just missed it somehow, glad you checked again

braingram:
I'll leave the final approval up to @melanieclarke if that works for all involved. If not let me know and I'll look at the regtest differences.

tapastro (approved):
No notes from me 👍

melanieclarke (approved):
All ready to go, I think.  Downstream tests will fail until the jwst PR is merged, but hopefully that should be soon.

PR #469, opened 2025-04-29
Add support for multi-valued defaults in recarray-like schemas

### LINKED ISSUE #468: Support multi-valued defaults for table-like data
Related to https://github.com/spacetelescope/stdatamodels/pull/466, there was a request to change the default values of some of the columns in a table to NaN.

Imagine the schema for the data table looks like
```
%YAML 1.1
---
$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
id: "http://stsci.edu/schemas/jwst_datamodel/wfss_spectable.schema"
title: Extracted spectral data table
fits_hdu: EXTRACT1D
default: 0.0
datatype:
- name: WAVELENGTH
  datatype: float64
- name: DQ
  datatype: uint32
- name: NAME
  datatype: [ascii, 20]
```
which is a slimmed-down version of what we need for https://github.com/spacetelescope/jwst/issues/9282.  Now call
```
model = dm.WFSSMultiSpecModel((50,))
print(model.spec_table["WAVELENGTH"]) # returns array of zero-valued float64
print(model.spec_table["DQ"]) # returns array of zero-valued int
print(model.spec_table["NAME"]) # returns array of byte strings b'0.0'
```

The desired behavior would be that the wavelength column could default to an array of NaNs, the DQ column could default to zero, and the name column could default to an empty string.  However, it appears there's no way to make this happen at present, because `stdatamodels.properties._make_default_array()` cannot handle a multi-valued data type, and (AFAIK) there's no way to define a default for each column.

Possible solutions would be:

- Handle parsing multi-valued datatypes in `_make_default_array`.  It appears there are no issues loading an arbitrarily complex datatype value from the schema, it just ends up as a string.  So we could e.g. check if the datatype field looks like a dict (or list) and handle it from there.
- Make it easier for individual datamodel classes to override `_make_default_array()`.  Right now it's a helper function to `getattr` and not directly accessible via e.g. `super()._make_default_array()`

Maybe there are also easier ways to handle this that I'm missing.

emolter:
Relates to [JP-3931](https://jira.stsci.edu/browse/JP-3931)

Closes #468 

This PR allows multi-valued defaults to be passed in for table-like schemas.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/properties.py
+++ b/src/stdatamodels/properties.py
@@ -243,7 +243,11 @@ def _make_default_array(attr, schema, ctx):
     if default is not None:
         if isinstance(default, list):
             # support multi-valued defaults for when array is recarray
-            default = np.array(tuple(default), dtype=dtype)
+            try:
+                default = np.array(tuple(default), dtype=dtype)
+            except ValueError:
+                msg = f"Invalid default value {default} for recarray dtype {dtype}"
+                raise ValueError(msg) from None
         array[...] = default
     return array
 
--- a/src/stdatamodels/properties.py
+++ b/src/stdatamodels/properties.py
@@ -241,6 +241,9 @@ def _make_default_array(attr, schema, ctx):
 
     array = np.empty(shape, dtype=dtype)
     if default is not None:
+        if isinstance(default, list):
+            # support multi-valued defaults for when array is recarray
+            default = np.array(tuple(default), dtype=dtype)
         array[...] = default
     return array
 
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -167,6 +167,23 @@ def test_default_value_anyof_schema():
         assert dm.meta.foo is None
 
 
+def test_multivalued_default_table_schema():
+    """Test setting list-like default values in a table schema"""
+    with TableModel((10,)) as dm:
+        defaults = dm.schema["properties"]["table"]["default"]
+        columns = [col["name"] for col in dm.schema["properties"]["table"]["datatype"]]
+        for default, name in zip(defaults, columns):
+            print(dm.table[name])
+            if default == "NaN":
+                default = np.nan
+            elif isinstance(default, str):
+                # allclose has trouble with string comparisons
+                for elem in dm.table[name]:
+                    assert elem.decode("utf-8") == default
+                continue
+            assert np.allclose(dm.table[name], default, equal_nan=True)
+
+
 def test_secondary_shapes():
     """
     Confirm that a non-primary array takes on the shape
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -6,7 +6,7 @@ import numpy as np
 
 from stdatamodels import DataModel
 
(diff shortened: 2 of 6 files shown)


tests/test_models.py
```diff
@@ -167,6 +167,23 @@ def test_default_value_anyof_schema():
         assert dm.meta.foo is None
 
 
+def test_multivalued_default_table_schema():
+    """Test setting list-like default values in a table schema"""
+    with TableModel((10,)) as dm:
+        defaults = dm.schema["properties"]["table"]["default"]
+        columns = [col["name"] for col in dm.schema["properties"]["table"]["datatype"]]
+        for default, name in zip(defaults, columns):
+            print(dm.table[name])
```
melanieclarke, line 176: Debug print

src/stdatamodels/properties.py
```diff
@@ -241,6 +241,9 @@ def _make_default_array(attr, schema, ctx):
 
     array = np.empty(shape, dtype=dtype)
     if default is not None:
+        if isinstance(default, list):
```
melanieclarke, line 244: Should we add some checking here to make sure the default list matches the table columns?

melanieclarke (commented):
Looks good to me, but we might want to add some error checking for the default list definition.  If there isn't another test that already does this, it might also be nice to write a test that checks that a single default value is still supported.

tests/test_models.py
emolter: whoops thanks

src/stdatamodels/properties.py
emolter: Sounds good.  I wrote a test for this, and see that the default Python error handling raises errors for this already. 
 But I agree that the messages could be more descriptive.

emolter: Let me know how https://github.com/spacetelescope/stdatamodels/pull/469/commits/822b5fe65d5aa443897266e33862d2249ff08a23 looks to you

emolter: Let me know how https://github.com/spacetelescope/stdatamodels/pull/469/commits/822b5fe65d5aa443897266e33862d2249ff08a23 looks to you

melanieclarke: Looks clear enough to me!

melanieclarke (commented):
I tested these changes with my draft models for TSO support and was able to get a sensible empty spec_table with:
empty_model = datamodels.TSOSpecModel((10,))

For my information, are the default values used in any other context?

emolter:
> For my information, are the default values used in any other context?

Not AFAIK

melanieclarke:
@braingram - did you want to review this change?  I'd like to get this in soon if possible so I can send in PRs for JP-3841 that rely on this change.

braingram (approved):

braingram:
LGTM. I saw your review and assumed it was ready to go. Thanks for pinging me and please do so if I haven't chimed in on something where I could help.

PR #476, opened 2025-05-01
Refactor to_flat_dict to handle nested lists

### LINKED ISSUE #475: `read_metadata` fails with option `flatten=True` on uncal files
```
from stdatamodels.jwst.datamodels import read_metadata
fname = "~/test_bigdata/jwst-pipeline/dev/miri/image/jw01024001001_04101_00001_mirimage_uncal.fits"
read_metadata(fname, flatten=False) # works
read_metadata(fname, flatten=True) # fails on _to_flat_dict()
```

```
Traceback (most recent call last):
    tree = read_metadata(fname, flatten=True) # fails on _to_flat_dict
  File "stdatamodels/src/stdatamodels/jwst/datamodels/util.py", line 573, in read_metadata
    return _to_flat_dict(tree)
  File "stdatamodels/src/stdatamodels/jwst/datamodels/util.py", line 474, in _to_flat_dict
    recurse(tree, [])
    ~~~~~~~^^^^^^^^^^
  File "stdatamodels/src/stdatamodels/jwst/datamodels/util.py", line 463, in recurse
    recurse(val, current_path)
    ~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "stdatamodels/src/stdatamodels/jwst/datamodels/util.py", line 463, in recurse
    recurse(val, current_path)
    ~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "stdatamodels/src/stdatamodels/jwst/datamodels/util.py", line 468, in recurse
    recurse(item, path + [indexed_key])
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "stdatamodels/src/stdatamodels/jwst/datamodels/util.py", line 457, in recurse
    for key, val in subtree.items():
                    ^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'items'
```

The failure happens with `subtree = ['NEXTEND', 4, 'Number of file extensions'], path = ['extra_fits', 'PRIMARY', 'header.0']`.  Apparently a dict is expected but it's receiving a list instead.

See also regtest failures here https://github.com/spacetelescope/RegressionTests/actions/runs/14627996568

emolter:
Closes #475 

This PR fixes a bug that the initial implementation of `_to_flat_dict` failed to handle nested lists.

That was causing the failures discussed on the issue thanks to our favorite piece of stdatamodels, `extra_fits`, which apparently represents headers as nested lists.  The fact that this only occurred inside `extra_fits` was one reason we didn't catch this initially.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/util.py
+++ b/src/stdatamodels/jwst/datamodels/util.py
@@ -461,7 +461,7 @@ def _to_flat_dict(tree):
                     continue
                 current_path = path + [key]
                 recurse(val, current_path)
-        elif isinstance(subtree, list):
+        elif isinstance(subtree, (list, tuple)):
             for i, item in enumerate(subtree):
                 indexed_key = f"{path[-1]}.{i}" if path else str(i)
                 recurse(item, path[:-1] + [indexed_key])
--- a/src/stdatamodels/jwst/datamodels/util.py
+++ b/src/stdatamodels/jwst/datamodels/util.py
@@ -454,22 +454,19 @@ def _to_flat_dict(tree):
     flat_dict = {}
 
     def recurse(subtree, path):
-        for key, val in subtree.items():
-            if key == "wcs":
-                # Skip the WCS object because it is recursive
-                continue
-            current_path = path + [key]
-            if isinstance(val, dict):
+        if isinstance(subtree, dict):
+            for key, val in subtree.items():
+                if key == "wcs":
+                    # Skip the WCS object because it is recursive
+                    continue
+                current_path = path + [key]
                 recurse(val, current_path)
-            elif isinstance(val, (list, tuple)):
-                for i, item in enumerate(val):
-                    indexed_key = f"{key}.{i}"
-                    if isinstance(item, (dict, list, tuple)):
-                        recurse(item, path + [indexed_key])
-                    else:
-                        flat_dict[".".join(path + [indexed_key])] = item
-            else:
-                flat_dict[".".join(current_path)] = val
+        elif isinstance(subtree, list):
+            for i, item in enumerate(subtree):
+                indexed_key = f"{path[-1]}.{i}" if path else str(i)
+                recurse(item, path[:-1] + [indexed_key])
+        else:
+            flat_dict[".".join(path)] = subtree
 
     recurse(tree, [])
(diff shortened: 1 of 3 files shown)


emolter:
Regression tests with jwst main: https://github.com/spacetelescope/RegressionTests/actions/runs/14782538805

Regression tests with https://github.com/emolter/jwst/tree/subclass-crds and https://github.com/braingram/stpipe/tree/subclass_get_crds_parameters: https://github.com/spacetelescope/RegressionTests/actions/runs/14782507395
The latter tests call `read_metadata` for nearly every input data file in our regression test suite - this was how the bug was discovered initially.

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -454,22 +454,19 @@ def _to_flat_dict(tree):
...
-            current_path = path + [key]
-            if isinstance(val, dict):
+        if isinstance(subtree, dict):
+            for key, val in subtree.items():
+                if key == "wcs":
+                    # Skip the WCS object because it is recursive
+                    continue
+                current_path = path + [key]
                 recurse(val, current_path)
-            elif isinstance(val, (list, tuple)):
```
melanieclarke, line 464: It looks to me like this is dropping recursion handling for a tuple val - is that not needed?

melanieclarke (commented):
This looks good to me -- testing on an uncal file locally, it now reports metadata instead of crashing.  One quick question, below.

src/stdatamodels/jwst/datamodels/util.py
emolter: Just an oversight on my end.  It turns out not to matter for real datamodels, but I don't see any harm in handling that case.  https://github.com/spacetelescope/stdatamodels/pull/476/commits/f11f21ca2130eb51479badd7a3b07dd5b35f3bdd adds it back and adds a unit test.  I will restart the regtest runs

melanieclarke: Excellent, thanks.

emolter:
Both regression test reruns are passing, see links above

melanieclarke (approved):
LGTM

PR #477, opened 2025-05-01
JP-3841: Flat spectral table for TSO

melanieclarke:
Resolves [JP-3841](https://jira.stsci.edu/browse/JP-3841)

Add a new spectral model for TSO data, to hold a flat table containing all integrations in a single FITS extension.

New models are used in https://github.com/spacetelescope/jwst/pull/9430

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/mrs_spec.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/mrs_spec.schema.yaml
@@ -5,6 +5,7 @@ id: "http://stsci.edu/schemas/jwst_datamodel/mrs_spec.schema"
 title: Spectrum data model with residual fringe correction for MIRI MRS
 allOf:
 - $ref: core.schema
+- $ref: specmeta.schema
 - type: object
   properties:
     spec_table:
--- a/src/stdatamodels/jwst/datamodels/schemas/mrs_spectable.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/mrs_spectable.schema.yaml
@@ -4,6 +4,7 @@ $schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
 id: "http://stsci.edu/schemas/jwst_datamodel/mrs_spectable.schema"
 title: Extracted spectral data table with residual fringe correction for MIRI MRS
 fits_hdu: EXTRACT1D
+default: [NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, 0, NaN, NaN, NaN, NaN, NaN, 0, NaN, NaN, NaN]
 datatype:
 - name: WAVELENGTH
   datatype: float64
--- a/src/stdatamodels/jwst/datamodels/schemas/mrs_spectable.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/mrs_spectable.schema.yaml
@@ -4,7 +4,29 @@ $schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
 id: "http://stsci.edu/schemas/jwst_datamodel/mrs_spectable.schema"
 title: Extracted spectral data table with residual fringe correction for MIRI MRS
 fits_hdu: EXTRACT1D
-default: [NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, 0, NaN, NaN, NaN, NaN, NaN, 0, NaN, NaN, NaN]
+default: [
+  NaN,  # WAVELENGTH
+  NaN,  # FLUX
+  NaN,  # FLUX_ERROR
+  NaN,  # FLUX_VAR_POISSON
+  NaN,  # FLUX_VAR_RNOISE
+  NaN,  # FLUX_VAR_FLAT
+  NaN,  # SURF_BRIGHT
+  NaN,  # SB_ERROR
+  NaN,  # SB_VAR_POISSON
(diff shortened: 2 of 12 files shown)


melanieclarke:
Updated regression tests:
https://github.com/spacetelescope/RegressionTests/actions/runs/14935241014

src/stdatamodels/jwst/datamodels/schemas/mrs_spectable.schema.yaml
```diff
@@ -4,6 +4,7 @@ $schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
 id: "http://stsci.edu/schemas/jwst_datamodel/mrs_spectable.schema"
 title: Extracted spectral data table with residual fringe correction for MIRI MRS
 fits_hdu: EXTRACT1D
+default: [NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, 0, NaN, NaN, NaN, NaN, NaN, 0, NaN, NaN, NaN]
```
emolter, line 7: small thing, but would it be possible to add comments labeling each of these defaults with a comment so it's easier to see what corresponds to what?

src/stdatamodels/jwst/datamodels/schemas/mrs_spec.schema.yaml
emolter, line 1: Is the addition of specmeta here and in spec.schema.yaml necessary for the PR, or is it a "while we're at it" kind of change?

src/stdatamodels/jwst/datamodels/schemas/specmeta.schema.yaml
```diff
@@ -154,36 +159,6 @@ properties:
     default: 's'
     fits_keyword: TIMEUNIT
     fits_hdu: EXTRACT1D
-  start_time_mjd:
```
emolter, line 157: why are all of these being removed from specmeta?  I haven't read your white light step refactor yet, but isn't this going to cause failures on https://github.com/spacetelescope/jwst/blob/b38915d73031c771332b543bf991ca64bd0e332e/jwst/white_light/white_light.py#L57 ?  Even if not, why are all of these pieces of information deemed no longer helpful?

emolter (commented):
Looks great Melanie, I just had a few questions

src/stdatamodels/jwst/datamodels/schemas/mrs_spectable.schema.yaml
melanieclarke: Yes, I noticed you did that in your PR and liked it a lot.  I'll add it here too.

src/stdatamodels/jwst/datamodels/schemas/mrs_spec.schema.yaml
melanieclarke: Not necessary, but nice to have. We never really use SpecModels by themselves, but we do add these metadata keys to spec models before they are appended to the spec list in the multi-spec models.  This works to get the additions in the final model, but it's not ideal to expect it to store attributes it doesn't know about.

src/stdatamodels/jwst/datamodels/schemas/specmeta.schema.yaml
melanieclarke: These are time keys that are only relevant or populated for multi-integration spectra, which will always be TSO spectral models instead, and will need the time metadata in the flat table instead.  I've revised the white light step to expect them in the table instead of the metadata.

src/stdatamodels/jwst/datamodels/schemas/mrs_spec.schema.yaml
emolter: Yeah I agree it's strange that it didn't exist in the SpecModel itself.

With it being added there, does it still need to be explicitly specified in MultiSpecModel?

melanieclarke: I think it does, yes, because the MultiSpecModel does not really contain a list of SpecModel, it just contains a list of objects with the same metadata and table structure as a SpecModel.

emolter: so confusing

but ok thanks

melanieclarke:
Regtests with jwst on main:
https://github.com/spacetelescope/RegressionTests/actions/runs/15192548995

emolter:
I assume the failure in the test `test_nircam_frame_averaged_darks: jwst.regtest.test_nircam_image` was caused by Tyler's PR and not this one?

melanieclarke:
> I assume the failure in the test `test_nircam_frame_averaged_darks: jwst.regtest.test_nircam_image` was caused by Tyler's PR and not this one?

Thanks, I overlooked that one - I was in the process of merging and okifying Tyler's PR while this test was running.  I can run again to get a cleaner result to okify from.

emolter (approved):
This looks great Melanie, I just made a few non-crucial comments.

src/stdatamodels/jwst/datamodels/schemas/tso_spectable.schema.yaml
```diff
@@ -0,0 +1,90 @@
...
+  NaN,  # MJD-BEG
+  NaN,  # MJD-AVG
+  NaN,  # MJD-END
+  NaN,  # TDB-BEG
+  NaN,  # TDB-MID
+  NaN   # TDB-END
+]
+datatype:
+- name: INT_NUM
+  datatype: int32
```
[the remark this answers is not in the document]
emolter: ok if it's reproducing what is in the core metadata then I agree that what you've done is preferred

PR #480, opened 2025-05-15
JP-3906: Remove err array initialization from dark models

tapastro:
Related to [JP-3906](https://jira.stsci.edu/browse/JP-3906)

~~This PR removes the error arrays from the dark schemas~~ 
This PR removes the forced initialization of error arrays from the dark datamodel `__init__` methods. They are no longer referenced in pipeline processing. This change should not exclude their being provided in the dark files, but darks that do not contain them will not have defaults auto-populated, unnecessarily inflating their file size.

## Tasks

</details>

### DIFF
--- a/changes/480.removal.rst
+++ b/changes/480.removal.rst
@@ -1,0 +1,1 @@
+Remove initialization of err array from dark datamodel init methods.
--- a/changes/482.removal.rst
+++ b/changes/482.removal.rst
@@ -1,0 +1,1 @@
+Deprecate DataModel.get_fits_wcs and DataModel.set_fits_wcs
--- a/src/stdatamodels/jwst/datamodels/dark.py
+++ b/src/stdatamodels/jwst/datamodels/dark.py
@@ -16,8 +16,6 @@ class DarkModel(ReferenceFileModel):
          Dark current array
     dq : numpy uint32 array
          2-D data quality array for all planes
-    err : numpy float32 array
-         Error array
     dq_def : numpy table
          DQ flag definitions
     """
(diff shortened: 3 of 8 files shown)


melanieclarke:
@tapastro - did you check with NIRSpec on this one?  They may use the ERR array for diagnostic purposes when generating and validating dark files.

tapastro:
This change shouldn't preclude darks from holding an error array, it just drops the requirement that such an array must exist. The primary change in behavior would be that any error array would be found in the datamodel as `model.extra_fits.ERR.data` rather than `model.err`. I'll ask Christian to take a look.

emolter:
The array doesn't need to be populated just because it exists in the schema.  While it's true that attempting to touch `model.err` in any way will cause them to be populated with a default value, if not touched the array shouldn't exist in memory.  However, I see that the `__init__` of `DarkModel` and `DarkMIRIModel` both currently include the line `self.err = self.err` which would cause the error array to populate with zeros.  I would think the easiest thing to do here would be simply to remove that line, such that that ERR remains in the schema, but the `err` array doesn't exist unless requested by the pipeline.

emolter:
Code looks good to me.  Can you run the jwst regression tests?

melanieclarke:
jwst regtests with the associated changes for stcal and jwst here:
 https://github.com/spacetelescope/RegressionTests/actions/runs/15053943380

I'm not sure if there was a run with this branch + main for jwst, so I'll start one here:
https://github.com/spacetelescope/RegressionTests/actions/runs/15170987515

melanieclarke (approved):
Regtests on main are passing, so I think we can move ahead with this.  @emolter - are you happy with the current state?

melanieclarke:
Unknown error on the readthedocs side -- it happens sometimes.  I don't have permissions to restart the docs build for this repo, but it's unlikely to be related to this PR.

PR #485, opened 2025-05-27
JP-3904 New Nirspec Dark model and added p_grating to superbias

jemorrison:
Resolves [JP-3004(https://jira.stsci.edu/browse/JP-3904)

Closes #

This PR creates a new NIRSpec dark reference datamodel. In this new model, two extensions are added to the data model, DRK and UNC. These extensions are used by the NIRSpec team for their testing. 
In addition, the P_GRATTING was added to both the NIRSpec dark datamodel and super bias.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/__init__.py
+++ b/src/stdatamodels/jwst/datamodels/__init__.py
@@ -17,6 +17,7 @@ from .contrast import ContrastModel
 from .cube import CubeModel
 from .dark import DarkModel
 from .darkMIRI import DarkMIRIModel
+from .darkNIRSpec import DarkNIRSpecModel
 from .emi import EmiModel
 from .extract1d_spec import Extract1dIFUModel
 from .flat import FlatModel
@@ -145,6 +146,7 @@ __all__ = [
     "CubeModel",
     "DarkModel",
     "DarkMIRIModel",
+    "DarkNIRSpecModel",
     "DisperserModel",
     "DistortionModel",
     "DistortionMRSModel",
--- a/src/stdatamodels/jwst/datamodels/darkNIRSpec.py
+++ b/src/stdatamodels/jwst/datamodels/darkNIRSpec.py
@@ -1,0 +1,37 @@
+from stdatamodels.dynamicdq import dynamic_mask
+from .dqflags import pixel
+from .reference import ReferenceFileModel
+
+
+__all__ = ["DarkNIRSpecModel"]
+
+
+class DarkNIRSpecModel(ReferenceFileModel):
+    """
+    A data model for NIRSpec dark reference files.
+
+    Attributes
+    ----------
+    data : numpy float32 array
+         Dark current array
+    dq : numpy uint32 array
+         2-D data quality array for all planes
+    dq_def : numpy table
+         DQ flag definitions
+    average_dark_current: numpy float32 array
+         Average dark current for each pixel
+    dark_rate : numpy float32 array
+         Dark rate used by NIRSpec team
+    dark_rate_unc : numpy float32 array
+         Dark rate uncertainties
+    """
+
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/darkNIRSpec.schema"
+
+    def __init__(self, init=None, **kwargs):
+        super(DarkNIRSpecModel, self).__init__(init=init, **kwargs)
+
+        self.dq = dynamic_mask(self, pixel)
+
+        # Implicitly create arrays
+        self.dq = self.dq
--- a/src/stdatamodels/jwst/datamodels/schemas/darkNIRSpec.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/darkNIRSpec.schema.yaml
@@ -1,0 +1,59 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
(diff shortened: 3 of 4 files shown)


emolter:
@jemorrison RE your question, the module name (`.py` file) cannot have capital letters in it.  Instead of making it lowercase though, I suggest putting the new NIRSpec dark model inside the existing `dark.py`

jemorrison:
@tapastro @braingram @melanieclarke 
With @braingram help he pointed out that in the reference files the P_GRATIN = 'N/A' does not work. That keyword needs to be  set to  P_GRATIN='N/A |'
Brett pointed me at: https://hst-crds.stsci.edu/static/users_guide/reference_conventions.html#matching-keyword-patterns
Note the line: The trailing | is required to satisfy the calibration code data model schema checks of allowed patterns.

I took the NIRSPEC super bias and dark reference file that gave me the errors. Both of these datamodels now have 
- $ref: keyword_pgrating.schema
I edited these reference files and changed P_GRATIN = 'N/A ' to P_GRATIN='N/A |' and now they work.

So those reference files need to be updated. I made a nirspec specific dark datamodel - that now includes - $ref: keyword_pgrating.schema so the just the NIRSPEC darks need to be fixed.

At this time I edited the general super bias datamodel and added keyword_pgrating.schema. That might be a problem. I will re-run the regression tests and see if it is.

tapastro:
> @tapastro @braingram @melanieclarke With @braingram help he pointed out that in the reference files the P_GRATIN = 'N/A' does not work. That keyword needs to be set to P_GRATIN='N/A |' Brett pointed me at: https://hst-crds.stsci.edu/static/users_guide/reference_conventions.html#matching-keyword-patterns Note the line: The trailing | is required to satisfy the calibration code data model schema checks of allowed patterns.
> 
> I took the NIRSPEC super bias and dark reference file that gave me the errors. Both of these datamodels now have
> 
> * $ref: keyword_pgrating.schema
>   I edited these reference files and changed P_GRATIN = 'N/A ' to P_GRATIN='N/A |' and now they work.
> 
> So those reference files need to be updated. I made a nirspec specific dark datamodel - that now includes - $ref: keyword_pgrating.schema so the just the NIRSPEC darks need to be fixed.
> 
> At this time I edited the general super bias datamodel and added keyword_pgrating.schema. That might be a problem. I will re-run the regression tests and see if it is.

You may want to alert Christian on the ticket that this will require new delivery of all old files, to allow use of the grating pattern schema. I don't think that was assumed to be necessary.

It kind of seems like buggy/undesirable behavior - I would have imagined the purpose of the pattern was to allow for "normal" entries - if we require a trailing pipe for comparison, should the comparator be investigated? Perhaps this usage is already widespread and I'm just out of the loop.

jemorrison:
The NIRSpec Superbias reference file in CRDS has an incorrect value for P_GRATIN = 'N/A' it needs to be 'N/A|'
Until this is updated an error message is logged to the screen - but the step runs and the reference file is applied.

changes/485.feature.rst
```diff
@@ -0,0 +1 @@
+Added new nirspec_dark model with dark_rate and dark_rate_unc and added p_gratin to dark and superbias models
```
emolter, line 1: I think this could be split onto two lines, something like
```
Added new DarkNirspecModel that includes `dark_rate` and `dark_rate_unc` extensions.
Added `p_gratin` header keyword to dark and superbias models
```
and if it's easy to explain in half a sentence, you could add the reason why `p_gratin` is being added.  Isn't it about CRDS queries?

src/stdatamodels/jwst/datamodels/schemas/dark_nirspec.schema.yaml
```diff
@@ -0,0 +1,59 @@
...
+      datatype: float32
+    average_dark_current:
+      title: Average dark current
+      fits_hdu: AVDRKCUR
+      default: 0.0
+      ndim: 2
+      datatype: float32
+    dark_rate:
+      title: Dark rate array
+      fits_hdu: DRK
```
emolter, line 49: If this is what the instrument team wants I'm ok with it, but the name of this HDU seems confusing to me. The whole model is a dark model so at a glance it leaves me wondering what the difference is between this and the SCI array.

emolter (commented):
Thanks for the updates.

I just started a rerun of the regression tests, it looked like there were one or more unrelated failures there so I'm mostly just making sure those go away.

Do we need to wait to coordinate with the JWST PR?

src/stdatamodels/jwst/datamodels/schemas/dark_nirspec.schema.yaml
jemorrison: Yes this is what the instrument team wants. They have existing software that reads these extensions. There is a long conversation on this in the original ticket:https://jira.stsci.edu/browse/JP-3904.
There are subtle differences (apparently)

emolter: there are definitely differences because one of them is 2-D and one is 3-D.  Maybe it would help to update the title field in some way?  Although I don't know what to suggest because I don't really understand this model very well

melanieclarke: The difference is that the SCI array contains raw ramp counts, the DRK array contains a rate.  The NIRSpec ref file software generates the SCI ramp counts from a projected rate value, so it's helpful to store the rate image in the dark for diagnostic purposes.

PR #486, opened 2025-05-28
Deprecate interpret_bit_flags

### LINKED ISSUE #460: Consider deprecation of `interpret_bit_flags` in favor of the astropy version
https://github.com/spacetelescope/stdatamodels/pull/409 attempted removal of `interpret_bit_flags` and `dqflags_to_mnemonics`.  It was discovered that these are in active use and the change was reverted by https://github.com/spacetelescope/stdatamodels/pull/459.  However, some cleanup could still be done with the functions that handle DQ flags; Astropy provides a (nearly?) 1:1 replacement for `interpret_bit_flags`:  https://docs.astropy.org/en/stable/_modules/astropy/nddata/bitmask.html#interpret_bit_flags.  It's likely we could deprecate the stdatamodels version in favor of the Astropy version.

emolter:
Closes #460 

This PR deprecates the stdatamodels version of `interpret_bit_flags` in favor of directly calling the Astropy version.

## Tasks

</details>


src/stdatamodels/dqflags.py
```diff
@@ -21,6 +22,9 @@ def interpret_bit_flags(bit_flags, flip_bits=None, mnemonic_map=None):
     Wraps `astropy.nddata.bitmask.interpret_bit_flags`, allowing the
     bit mnemonics to be used in place of integers.
 
+    This method is deprecated and will be removed in a future version.
```
pllim, line 25: Use this? https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

pllim (commented):
I don't see any API change in this PR. If there is no API change, there is no need for transition guide. Thanks!

You cannot add me as reviewer because I do not have access to this repo. I would need to be in a team with at least a "read" access to be added as reviewer. It doesn't make sense because this repo is public but that is how GH assignment work. 🤷‍♀️

tests/test_dq.py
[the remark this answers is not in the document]
emolter: good suggestion. see https://github.com/astropy/astropy/issues/18190

tests/test_dq.py
```diff
@@ -1,8 +1,12 @@
...
 from stdatamodels.jwst.datamodels.dqflags import pixel
 
 
 def test_dqflags():
     assert dqflags.dqflags_to_mnemonics(1, pixel) == {"DO_NOT_USE"}
     assert dqflags.dqflags_to_mnemonics(7, pixel) == {"JUMP_DET", "DO_NOT_USE", "SATURATED"}
-    assert dqflags.interpret_bit_flags("DO_NOT_USE + WARM", mnemonic_map=pixel) == 4097
+    with pytest.warns(DeprecationWarning):
+        assert dqflags.interpret_bit_flags("DO_NOT_USE + WARM", mnemonic_map=pixel) == 4097
+    assert ap_interpret_bit_flags("DO_NOT_USE+WARM", flag_name_map=pixel) == 4097
```
pllim, line 12: Maybe not something as serious as conversion guide but this whitespace business should be documented somewhere for users looking to switch over? Is string usage with operator common?

src/stdatamodels/dqflags.py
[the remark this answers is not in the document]
emolter: should be fixed in https://github.com/spacetelescope/stdatamodels/pull/486/commits/3655c3bf89365c67edd66defb152b3f0a0d8ddaa

tests/test_dq.py
emolter: let me know if https://github.com/spacetelescope/stdatamodels/pull/486/commits/3655c3bf89365c67edd66defb152b3f0a0d8ddaa looks ok

docs/source/api.rst
```diff
@@ -3,6 +3,7 @@ stdatamodels API
 
 .. automodapi:: stdatamodels
 .. automodapi:: stdatamodels.exceptions
+.. automodapi:: stdatamodels.dqflags
```
braingram, line 6: Minor comment but this adds the docstring of `dqflags` to the docs. That docstring (not modified in this PR so I can't add a suggestion there) is incorrect:

> The data structure that stores bit flags is just the standard Python `int`,
which provides 32 bits.

To be pedantic I'd argue the array of values are "flags" and each element is a [bit field](https://en.wikipedia.org/wiki/Bit_field). Perhaps we modify the docstring to:

```
Interpret JWST data quality flags.

The flags are binary-packed structures representing information about a given element (a bit field). This information can be mapped to and from plain text mnemonics using `astropy.nddata.bitmask.interpret_bit_flags` and `stdatamodels.dqflags.dqflags_to_mnemonics`.
```

braingram (approved):
Just one comment about the `dqlfags` module docstring. Otherwise LGTM.

PR #487, opened 2025-05-29
Remove code that attempts to handle opening datamodels from buffer

### LINKED ISSUE #423: Simplify and improve filetype.check
Follow up from https://github.com/spacetelescope/stdatamodels/pull/422

filetype.check has 1 use in jwst:
https://github.com/spacetelescope/jwst/blob/7545a60d7e6c20ab584c135c2e7ecf1052e4deb5/jwst/resample/resample_step.py#L61
where a string is provided.

There are 2 uses in stdatamodels:
https://github.com/spacetelescope/stdatamodels/blob/26c9f0f373af8fdfa095b97d1e4b7441e21ef1cd/src/stdatamodels/jwst/datamodels/util.py#L114-L115
https://github.com/spacetelescope/stdatamodels/blob/26c9f0f373af8fdfa095b97d1e4b7441e21ef1cd/src/stdatamodels/model_base.py#L229-L233
Both appear to provide a string (although the first looks like it's within a block which might contain a file object that would trigger an exception on the conversion to a `Path`).
Both of those uses look like they'd be simplified if `filetype.check` supported `Path` inputs (which are not currently supported).

If we don't drop the file object input we may consider:
- using the file type checking provided by asdf for determining if a file is ASDF or FITS: https://www.asdf-format.org/projects/asdf/en/latest/api/asdf.util.FileType.html#asdf.util.FileType
- fixing the assumption that any non-ASDF/FITS file is a json file

emolter:
Closes #423 

This PR addresses the above-linked issue by making `filetype.check()` accept `Path` as input, and removing handling of `BytesIO` type.

It was discovered that reading datamodels from a `BytesIO` or `bytes`  representation already does not work on main.  This PR removes all code that attempted to handle that case.

This PR also changes an instance of `ValueError` to `TypeError` in `DataModel.__init__()`.  This is probably a slightly risky change and it's important to be careful with it.  (I could also be convinced to revert this if reviewers dislike it.)

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/filetype.py
+++ b/src/stdatamodels/filetype.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+import sys
 
 
 def check(init):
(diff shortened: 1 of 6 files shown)


emolter:
JWST regression tests started here https://github.com/spacetelescope/RegressionTests/actions/runs/15327200023

src/stdatamodels/filetype.py
```diff
@@ -7,43 +8,30 @@ def check(init):
 
     Parameters
     ----------
-    init : str or file-like object
-        The file path or file object
+    init : str, Path, or bytes
```
braingram, line 11: What use case requires `bytes` support here?

src/stdatamodels/model_base.py
```diff
@@ -228,10 +228,6 @@ def __init__(
             asdffile = fits_support.from_fits(init, self._schema, self._ctx, **kwargs)
 
         elif isinstance(init, (str, bytes, PurePath)):
-            if isinstance(init, PurePath):
-                init = str(init)
-            if isinstance(init, bytes):
-                init = init.decode(sys.getfilesystemencoding())
```
braingram, line 234: I think removing this results in passing a `bytes` instance to `asdf.open` (on line 244). This isn't supported. This might be "scope creep" but why are we supporting `bytes` here?

braingram (changes requested):
I think the `model_base` change results in passing an unsupported `bytes` object to `asdf.open`.

melanieclarke:
> LGTM. The rerun of the regtests show 3 errors that look unrelated.
> 
> They all contain `Extra keyword 'S_CLNFNS' in a: 'SKIPPED'`. Was there a CRDS or truth file update?

Those are mine, related to artifactory updates for https://github.com/spacetelescope/jwst/pull/9404.

PR #493, opened 2025-06-06
New datamodel for MIRI WFSS specwcs reference file 

jemorrison:

Closes #

This PR addresses ...

## Tasks

</details>


jemorrison:
running regression tests now. This is a new datamodel so I do not expect any failures

tapastro:
This looks like a good start! 

I don't know if the plan with this PR is just to support delivery of reference files, or if the aim is to support some amount of processing through spec2 - if the latter, here is some additional work required.

The entries in wcs_ref_models.py will allow `jwst` to open the specwcs file as a MIRIWFSSSpecwcsModel, but the next step in WCS generation is to use the polynomials provided in the specwcs datamodel to build the grism dispersion models, which we store in `stdatamodels/jwst/transforms/models.py`.

The NIRCam examples are here (as well as others): https://github.com/spacetelescope/stdatamodels/blob/533aeefc75d374e294b84a03441f7f2f13600aa3/src/stdatamodels/jwst/transforms/models.py#L1251

You can see how the pipeline uses the dispersion models to build the WCS here:
https://github.com/spacetelescope/jwst/blob/becaef5dfd2f0226017a68245d235328f391b570/jwst/assign_wcs/nircam.py#L393-L433

emolter (commented):
Thanks for putting this together Jane. It makes me vaguely uncomfortable to merge this without any example usages.  I understand that you want it to become available sooner rather than later.  Is there any way to show us, on the code review side, that this does indeed load all the extensions and relevant metadata from the example reference file they presumably have delivered to you?

changes/493.feature.rst
```diff
@@ -0,0 +1 @@
+added datamodel for MIRI WFSS specwcs reference file
```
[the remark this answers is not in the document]
jemorrison: I am the only one working on this for the next weeks and the team did load the files in CRDS test. So I could broaden this pr and add reading and using the datamodel.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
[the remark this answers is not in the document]
emolter: What you've done, separating them, is probably the safest thing to do.  I do wonder though whether adding those refs to the nircam model (and all the other specwcs models for that matter) would break anything or not.

I guess overall, and this probably doesn't concern you too much Jane, I'm worried about the ever-increasing number of nearly-but-not-quite-identical datamodels we support.

Do you have any thoughts here @braingram ?

changes/493.feature.rst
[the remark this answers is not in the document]
emolter: Do you mean, add a companion JWST PR where this is used?  If so, I think yes, that would be helpful on my end in order to understand this PR better.  But I'm also not sure I have the "authority" to make that call

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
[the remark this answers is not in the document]
braingram: At the moment I don't know enough about the uses and data structure here to have an informed opinion about consolidating these. I can echo your point that f they're not identical (and different data models are needed) then keeping them separate seems like the safest thing to do.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -458,6 +459,88 @@ def to_fits(self):
...
+            MIRI WFSS row dispersion model
+        dispy : `~astropy.modeling.Model`
+            MIRI WFSS column dispersion model
+        invdispl : `~astropy.modeling.Model`
+            MIRI WFSS inverse wavelength dispersion model
+        orders : `~astropy.modeling.Model`
+            MIRI WFSS orders, matched to the array locations of the
+            dispersion models
+        fwcpos_ref : float
+            The reference value for the filter wheel position
```
braingram, line 498: I don't see this in the schema or in https://jwst-crds-test.stsci.edu/browse/jwst_miri_specwcs_0137.asdf
Should it be in those or removed from here? Thanks for helping me understand this.

changes/493.feature.rst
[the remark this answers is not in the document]
jemorrison: Look at my Jupyter notebook example. I can add to this PR the transforms for MIRI WFSS in 
git/stdatamodels/src/stdatamodels/jwst/transforms
That is next thing I have to do anyway.

braingram:
@jemorrison anything I can do to help with this PR? Is it on hold until the next build?

jemorrison:
I have been trying to follow what NIRISS did (I might have missed something).
I have defined the "MIRIWFSSBackwardDispersion" class iin
/Users/morrison/git/stdatamodels/src/stdatamodels/jwst/transforms

I now see there is stuff in /stdatamodels/src/stdatamodels/jwst/transforms/converters that I need to add.  So once I add MIRIWFSSBackwardDispersion to  jwst_models.py in converters  should that be everything or do I need to do something else. 

Thanks I have never added a transform to datamodels so I do not know all the steps to get this working

nden:
@jemorrison You will also need a schema for this model, similar to the `niriss_grism_dispersion` schema.

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -2059,6 +2060,77 @@ def evaluate(self, x, y, x0, y0, order):
...
+        """
+        self.xmodels = xmodels
+        self.ymodels = ymodels
+        self.lmodels = lmodels
+
+        if name is None:
+            name = "miriwfss_backward_dispersion"
+        super(MIRIWFSSBackwardDispersion, self).__init__(name=name)
+        self.inputs = ("x", "y")
+        self.outputs = ("x", "y", "x0", "y0")
```
emolter, line 2099: How do these models avoid taking wavelength as an input?  Surely the wavelength is what allows you to go between the direct image x,y and the grism x,y?

jemorrison: There is only 1 order. I can add order to the model but it is always = 1. 
The lmodels (lambda models) is how to determine the wavelength. We do not provide a wavelength when going to direct to dispersed because we have not information on the wavelength in the direct image. But we can calculate the wavelength in the dispersed using the lmodels.

emolter: But you would still need a wavelength to get a _specific_ dispersed value.  If I have a direct image, and I want to model a source with a given spectrum, I need to know which part of the trace corresponds to what wavelength.

Right now the `evaluate` method you wrote has `t` hard-coded to `np.linspace(0, 1, 10)`, and just gives me the x,y values at all 10 of those trace positions.  In the other models, NIRISS and NIRCam, `t` is determined from the input wavelength and the lmodel, and then used to get x,y.  For those models, if I pass in one wavelength, I get out one x,y position. If I pass in 100 wavelengths, I get out 100 x,y positions.

nden:
@jemorrison The comments in this PR indicate this is supposed to add only a new reference file. However, it now includes work for the actual model. Please, either split the PR or update the comments  to avoid confusion.

jemorrison:
@braingram -When you get a chance could you see if I set up the new  manifest correctly.

braingram:
You're missing the update to [TRANSFORM_EXTENSIONS](https://github.com/spacetelescope/stdatamodels/blob/e77fc6a2facb58f0d374f576190ad32b15115232/src/stdatamodels/jwst/transforms/extensions.py#L41) to register the manifest.

I'll take a look at the manifest as well.

To make this PR easier to review would you add a unit test that uses the new transform (and saves it to an asdf file)?

src/stdatamodels/jwst/transforms/converters/jwst_models.py
```diff
@@ -124,6 +125,36 @@ def to_yaml_tree_transform(self, model, tag, ctx):
         return node
 
 
+class MIRIWFSSDispersionConverter(TransformConverterBase):
+    tags = ["tag:stsci.edu:jwst_pipeline/miriwfss_dispersion-*"]
```
braingram, line 129:
```suggestion
    tags = ["tag:stsci.edu:jwst_pipeline/miri_wfss_dispersion-*"]
```

braingram:
> @braingram Making progress. I can now run assign_wcs and not get an error. Now I need to make sure the transforms are giving the correct results. You mentioned making a unit test that uses the transform to write the results to an asdf file. What would the test be ? Comparing it to the asdf reference file to the input asdf file or something else ?

Perhaps https://github.com/spacetelescope/stdatamodels/blob/b9a49c2355507cccd07b7f18f60c0518d3070283/src/stdatamodels/jwst/transforms/tests/test_transforms.py#L265 is a relevant example test.

Would you also add a unit test for the new model?

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,182 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+            Output spectral order, same as input
+        """
+        if (wavelength < 0).any():
+            raise ValueError("Wavelength should be greater than zero")
+        try:
+            iorder = self._order_mapping[int(order.flatten()[0])]
+        except KeyError as err:
+            raise ValueError("Specified order is not available") from err
+        # t is trace normalization parameters (it was values of 0 to 1)
+        t = self.tmodels[iorder](wavelength)
```
nden, line 2064: `self.tmodels` is not defined anywhere

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,182 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+        xmodel = self.xmodels[iorder]
+        ymodel = self.ymodels[iorder]
+
+        dx = xmodel(t)
```
nden, line 2068: I don't think this will work with the current reference file and schema for it. `self.xmodels[iorder]` is a list f two models and so it can't be evaluated like this. OTH, I'm not sure why it's a list of two models and what it is supposed to be. I would have expected either one model (for no spatial dependence) or three (for spatial dependence). Is there a description of the calibration? In any case, this model and possibly the schema and model for the reference file needs to be fixed.

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,182 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+    def evaluate(self, x, y, x0, y0, order):
+        """
+        Transform from the dispersed plane into the direct image plane.
+
+        Parameters
+        ----------
+        x, y :  int, float, list
+            Input x, y location
+        x0, y0 : int, float, list
+            Source object x-center, y-center
```
nden, line 2130:
```suggestion
            Source object x-center, y-center in the direct image.
```

src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-1.3.0.yaml
```diff
@@ -0,0 +1,117 @@
...
+asdf_standard_requirement:
+  gte: 1.6.0
+tags:
+- tag_uri: "tag:stsci.edu:jwst_pipeline/miri_wfss_dispersion-1.0.0"
+  schema_uri: http://stsci.edu/schemas/jwst_pipeline/miri_wfss_dispersion-1.0.0
+  title: MIRI WFSS dispersion model
+  description: |-
+    Supports two models:
+    Given x, y, wave, order in effective direct image returns
+    x,y,wave,order in dispersed
```
nden, line 15:
```suggestion
    x, y, wave, order in dispersed
```

src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-1.3.0.yaml
```diff
@@ -0,0 +1,117 @@
...
+    Given x, y, x0, y0, order in dispersed image returns x, y, wave, order
+    in effective direct
+- tag_uri: "tag:stsci.edu:jwst_pipeline/nircam_grism_dispersion-1.2.0"
+  schema_uri: http://stsci.edu/schemas/jwst_pipeline/nircam_grism_dispersion-1.2.0
+  title: NIRCAM Grism dispersion model
+  description: |-
+    Supports two models:
+    Given x, y, wave, order in effective direct image returns
+    x,y,wave,order in dispersed
```
nden, line 24:
```suggestion
    x, y, wave, order in dispersed
```

src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/miri_wfss_dispersion-1.0.0.yaml
```diff
@@ -0,0 +1,48 @@
...
+anyOf:
+  - $ref: "http://stsci.edu/schemas/asdf/transform/transform-1.4.0"
+  - type: object
+    properties:
+      xmodels:
+        description: |
+          MIRI WFSS row dispersion model
+        type: array
+        items:
+          $ref: "http://stsci.edu/schemas/asdf/transform/transform-1.4.0"
```
nden, line 20: See my comment above but I think this does not match the reference file, where `xdisp` is a list of a list of 2 models.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
```diff
@@ -0,0 +1,60 @@
...
+            Input units for wavelength
+          anyOf:
+            - type: string
+            - $ref: http://stsci.edu/schemas/asdf/unit/unit-1.0.0
+        output_units:
+          description: |
+            Output units for wavelength
+          anyOf:
+            - type: string
+            - $ref: http://stsci.edu/schemas/asdf/unit/unit-1.0.0
```
nden, line 60: Should some of these be made `required`?

nden (changes requested):
I believe, if tests were written, they would have caught the major issues in this PR.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
```diff
@@ -0,0 +1,60 @@
...
+          MIRI WFSS inverse wavelength dispersion model, instance of astropy.modeling.Model
+        type: array
+        items:
+          type: object
+    orders:
+        description: |
+          MIRI WFSS orders, matched to the array locations of the dispersion models
+        type: array
+        items:
+          type: number
```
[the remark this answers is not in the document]
jemorrison: It probably does not matter but I just copied what NIRISS did - so if you think having 'number' is a problem when the specwcs_niriss_grism.schema.yaml might need to be updated as well (but not in the PR). 
As well as specwcs_nircam_grism.schema.yaml

src/stdatamodels/jwst/transforms/converters/jwst_models.py
```diff
@@ -124,6 +125,35 @@ def to_yaml_tree_transform(self, model, tag, ctx):
...
+        )
+
+    def to_yaml_tree_transform(self, model, tag, ctx):
+        yll = [list(m) for m in model.ymodels]
+        node = {
+            "xmodels": list(model.xmodels),
+            "ymodels": yll,
+            "lmodels": list(model.lmodels),
+            "model_type": type(model).name,
+        }
```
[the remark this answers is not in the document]
jemorrison: Thanks. I added orders later and forgot I added it. There is only 1 order for MIRI but Ned suggested I should have order to the schema so MIRI acts like the other grism type modes.

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,182 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+        meta : dict
+            Unused
+        """
+        self._order_mapping = {int(k): v for v, k in enumerate(orders)}
+        self.xmodels = xmodels
+        self.ymodels = ymodels
+        self.lmodels = lmodels
+        meta = {"orders": orders}
+        if name is None:
+            name = "miri_wfss_backward_dispersion"
```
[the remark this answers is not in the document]
jemorrison: Actually I just copied what NIRISS and NIRCAM did . 
  In src/stdatamodels/jwst/tranforms.models - when a model is initialized the name is set to None. This is how all the models are set up.  I assumed what they are doing is the preferred way. Maybe not - but I think it could be confusing to set it up differently just for miri.

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,182 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+    def evaluate(self, x, y, wavelength, order):
+        """
+        Transform from the direct image plane to the dispersed plane.
+
+        Parameters
+        ----------
+        x, y : float
+            Input x, y location
+        wavelength : float
+            Wavelength in angstroms
```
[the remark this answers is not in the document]
jemorrison: Whoops - it is microns - good catch

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,182 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+            The list of orders which are available to the model
+        lmodels : list
+            The list of models for the polynomial model in l
+        xmodels : list[tuples]
+            The list of tuple(models) for the polynomial model in x
+        ymodels : list[tuples]
+            The list of tuple(models) for the polynomial model in y
+            Name of the model
+        meta : dict
+            Unused
```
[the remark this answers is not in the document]
jemorrison: Ok that is probably a bad comment. I do set meta to the orders. So I will remove unused.

src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/miri_wfss_dispersion-1.0.0.yaml
```diff
@@ -0,0 +1,48 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/jwst_pipeline/miri_wfss_dispersion-1.0.0"
+title: >
+  MIRI WFSS dispersion model
+description: |
+  Supports one model:
+   - Given x,y,wave,order in effective direct image return x,y,wave,order in dispersed
```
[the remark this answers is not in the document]
jemorrison: removed.

src/stdatamodels/jwst/transforms/models.py
jemorrison: I have created a new reference file the old was incorrect. 
The xmodel is just 1D model. I was waiting to deliver a new reference file to make sure we did not need any other changes to it. Right now the xmodel has no spatial dependency. - But that might change once MIRI gets data for this mode.

src/stdatamodels/jwst/transforms/models.py
jemorrison: Fixed it - it should be lmodels.

src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/miri_wfss_dispersion-1.0.0.yaml
jemorrison: We have changed what xmodels is - it is now a 1D model

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
jemorrison: Honestly I do not know. None of the other grism models have required -but maybe they should be. What makes something 'required' in a schema. We need them all  - so are they all required ?

jemorrison:
@braingram I am not finished with this PR I am working on some tests.

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,158 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+            iorder = self._order_mapping[int(order.flatten()[0])]
+        except KeyError as err:
+            raise ValueError("Specified order is not available") from err
+
+        # The next two lines are to get around the fact that
+        # modeling.standard_broadcasting=False does not work.
+        x00 = x0.flatten()[0]
+        y00 = y0.flatten()[0]
+
+        t = np.linspace(0, 1, 10)  # sample t
```
emolter, line 2132:
```suggestion
        t = np.linspace(0, 1, self.sampling)  # sample t
```
Instead of hard-coding this to have 10 values, can you use the `self.sampling` attribute instead?  (`_GrismDispersionBase` already has this attribute defined and set to a default value of 40, which I assume would also work fine here.  But if you want to change the default you can always override it in this class)

src/stdatamodels/jwst/transforms/tests/test_transforms.py
```diff
@@ -639,6 +639,58 @@ def test_niriss_grism_roundtrip(direction):
     np.testing.assert_allclose(ordersi, orders)
 
 
+
+def test_miri_wfss_roundtrip():
```
emolter, line 643: This looks great Jane!

src/stdatamodels/jwst/transforms/tests/test_transforms.py
emolter, line 1: Can you also please add a test that checks that arrays of x, y, and wl can be passed into the transforms, i.e. that the transforms are vectorized?

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -2043,20 +2044,37 @@ def evaluate(self, x, y, wavelength, order):
...
         except KeyError as err:
             raise ValueError("Specified order is not available") from err
         # t is trace normalization parameters (it was values of 0 to 1)
+        
         t = self.lmodels[iorder](wavelength)
+        print('t', t)
+        print(type(t))
+        
+        print(self.lmodels[iorder])
+        print('input x, y', x,y)
```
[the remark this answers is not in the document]
emolter: @braingram I don't understand why our ruff check didn't flag this.  We have rule "T20" enabled, which is supposed to disallow print statements in code, and I don't see any ignores for this file.  Any ideas?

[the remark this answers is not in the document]
jemorrison: Oh it did I left in the print statements for my testing. I am still in testing mode.  
I did a --no-verify option to keep them. Sorry I should have flagged that at the top of the ticket. 
I still need to add a few more tests for this PR

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,158 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+            iorder = self._order_mapping[int(order.flatten()[0])]
+        except KeyError as err:
+            raise ValueError("Specified order is not available") from err
+        # t is trace normalization parameters (it was values of 0 to 1)
+
+        t = self.lmodels[iorder](wavelength)
+        xmodel = self.xmodels[iorder]
+        ymodel = self.ymodels[iorder]
+        dx = xmodel(t)
+        dy = ymodel[0](x, y) + dx * ymodel[1](x, y) + dx**2 * ymodel[2](x, y)
```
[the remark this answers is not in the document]
jemorrison: maybe. I really wanted to keep this simple for now. I liked to have to code right there - rather than having to look it up. Let me finished with some tests and then I could change it.

src/stdatamodels/jwst/transforms/models.py
jemorrison: Ok yes I can do that.

src/stdatamodels/jwst/transforms/models.py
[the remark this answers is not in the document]
emolter: that's fair enough.  The benefit of the helper function is that it handles arbitrary polynomial orders.  But I'm fine either way.

src/stdatamodels/jwst/transforms/models.py
jemorrison: @nden How is the best way to share it ? I tried to attach it but it would not let me attach an asdf file.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -458,6 +459,83 @@ def to_fits(self):
...
+        if displ is not None:
+            self.displ = displ
+        if dispx is not None:
+            self.dispx = dispx
+        if dispy is not None:
+            self.dispy = dispy
+        if invdispl is not None:
+            self.invdispl = invdispl
+        if orders is not None:
+            self.orders = orders
```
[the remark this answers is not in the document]
jemorrison: @emolter Clarification needed.  I hate to keep coming back to - I just copied what other instruments did. I assumed there was a good reason they all initialized them the way they did.
But maybe it is code that is not needed. 
In the _init- all the 'dispy...' are set to None. It seems like that is not needed if we are just going to
force all the self.dispy = None.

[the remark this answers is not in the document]
nden: That allows some perhaps unneeded flexibility, i.e. generating the file programmatically. I thought it can be useful when working on generating the ref file.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -458,6 +459,83 @@ def to_fits(self):
...
+            assert isinstance(self.meta.output_units, (str, u.NamedUnit))
+            assert self.meta.instrument.name == "MIRI"
+            assert self.meta.exposure.type == "MIR_WFSS"
+            assert self.meta.instrument.detector == "MIRIMAGE"
+            assert self.meta.reftype == self.reftype
+        except AssertionError:
+            if self._strict_validation:
+                raise
+            else:
+                warnings.warn(traceback.format_exc(), ValidationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
jemorrison: Hum what do you mean by the default schema handling this. I guess to me it makes sense to do it right off the bat - but I might not be understanding your suggestion correctly

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
nden: I don't think it's a good idea to touch the other instruments in this PR. Also I would leave the `input_units` and `output_units` out of `required` because they are not essential for the transform, only informational.

src/stdatamodels/jwst/transforms/tests/test_transforms.py
emolter: this one has array inputs https://github.com/spacetelescope/stdatamodels/blob/c73243314dfafcd3655cd3666207cef60c84d371/src/stdatamodels/jwst/transforms/tests/test_transforms.py#L349

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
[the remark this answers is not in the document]
emolter: Every time a datamodel gets opened, it goes through validation against the schema.  That's the "whole point" of schemas - to define expected attributes of a model and their acceptable values.

So my question is, if the schema is the standard way that we enforce that e.g. the [detector is a real JWST detector and not some random value](https://github.com/spacetelescope/stdatamodels/blob/c73243314dfafcd3655cd3666207cef60c84d371/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml#L419), then what is stopping us from updating the schema to enforce that the detector for this specific model has to be `"MIRIMAGE"`, rather than handling that validation in its own separate custom code.

Maybe this is not possible / complicated @braingram ?  I must admit I don't know what happens when you attempt to override a schema entry with another schema entry.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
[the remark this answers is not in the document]
emolter: @nden I'm not sure what additional flexibility it allows.  It would make it so these attributes are not assigned if they are set to `None`, but what specifically is having them set to `None` here preventing you from doing?

@jemorrison I know that, and it's not really your fault that Nadia and I have come in here and started questioning various aspects of the entire setup.  This review just opened up the opportunity to revisit some of the design decisions.  Overall I would be fine with you copying everything that NIRISS did and having a follow-up PR that applies changes like this one to all the models at once.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
```diff
@@ -0,0 +1,76 @@
...
+    required: [displ]
+  - type: object
+    required: [dispx]
+  - type: object
+    required: [dispy]
+  - type: object
+    required: [invdispl]
+  - type: integer
+    required: [orders]
+
```
nden, line 76: @jemorrison I think you need to remove lines 65-76. 
Then add a line
`required: [displ, dispx, dispy, invdispl, orders]`
It should be aligned with `properties`, i.e. leave 2 spaces.

jemorrison: @nden 
I figure it out and put a test in datamodels to catch for required values not given. 
Adding more things to this test and then I will push it up

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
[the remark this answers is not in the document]
jemorrison: I took out assert self.meta.instrument.detector == "MIRIMAGE" but left the other asserts.
That seems to be the standard for the other instruments

jemorrison:
@nden @emolter I think this is ready for another review. This has been a learning curve for me - making a new transform. I have gone through all the comments a few times and I think I have addressed them - but if I missed one sorry about that and let me know. I will look at this again tomorrow.

src/stdatamodels/jwst/datamodels/schemas/specwcs_miri_wfss.schema.yaml
jemorrison: @nden By the way as we look at better way to do things in the future. In the datamodels there is hardly any required values in the datamodels and I would think that many of the values would be required.  We might want to take another look at that issue.

src/stdatamodels/jwst/transforms/tests/test_transforms.py
```diff
@@ -639,8 +640,110 @@ def test_niriss_grism_roundtrip(direction):
...
+    y0 = 140
+    wl = np.linspace(5.5e-6, 6.5e-6, 21)  # 
+    model = models.MIRIWFSSBackwardDispersion(orders, lmodels, xmodels, ymodels)
+
+    xi, yi, x, y, order = model.evaluate(x0, y0, wl, orders)
+    assert xi.size == wl.size
+    assert yi.size == wl.size
+    assert x == x0
+    assert y == y0
+    assert_allclose(order, orders[0])
```
emolter, line 732: This is kind of a silly usage of `assert_allclose`; won't `order` be single-valued and equal to `1`?

emolter (commented):
Thanks very much for your hard work on this Jane.! I like the new tests a lot.

I had a few relatively minor suggestions - I think the biggest ones are some suggested renamings for consistency with NIRCam and NIRISS.

I didn't actually try to use the transforms or the specwcs file myself.  Do you yet have a draft JWST PR that uses the relevant pieces in `assign_wcs`?  If not I guess it would be ok, we can always go back and patch this PR if we discover anything wrong with it before these transforms are put into pipeline use.  But obviously in a perfect world we can get it right the first time.

Also, I opened https://github.com/spacetelescope/stdatamodels/issues/542 to track the suggestions from Nadia and I that are also applicable to the NIRCam and NIRISS models.  Feel free to add anything I missed into the comments on that issue.

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -36,6 +36,8 @@
     "NIRISSForwardRowGrismDispersion",
     "NIRISSForwardColumnGrismDispersion",
     "NIRISSBackwardGrismDispersion",
+    "MIRIWFSSBackwardDispersion",
+    "MIRIWFSSForwardDispersion",
```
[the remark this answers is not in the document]
jemorrison: I do not like Grism for MIRI because we do not have a Grism but a Prism.

changes/493.feature.rst
```diff
@@ -0,0 +1 @@
+Added datamodel for MIRI WFSS specwcs reference file
```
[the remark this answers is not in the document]
jemorrison: Yes I have a  draft PR that uses these transforms in assign wcs. https://github.com/spacetelescope/jwst/pull/9689
That is how I check that it works

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -458,6 +459,83 @@ def to_fits(self):
         raise NotImplementedError("FITS format is not supported for this file.")
 
 
+class MiriWFSSSpecwcsModel(ReferenceFileModel):
```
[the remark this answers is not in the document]
jemorrison: MIRI does not have a grism but a double prism. I realize this a minor point but I think we should not mix terms to be very clear.

src/stdatamodels/jwst/transforms/tests/test_transforms.py
jemorrison: Right did not mean all close. Will fix that

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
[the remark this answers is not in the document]
emolter: I'm fine with it since this seems like a pretty good reason

src/stdatamodels/jwst/transforms/models.py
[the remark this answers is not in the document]
emolter: Sounds good that's reasonable

nden (approved):
LGTM. Thanks!

braingram:
> @braingram I would like to move to get this PR merged. Should I run the regression tests using JWST main and this PR ? I have a branch of JWST where I use the transforms but I have to override a number of reference files for this mode and edit some test data to make it MIRI WFSS. We do not have this type of data in the regression suite because it is not an official mode yet and there is still so much work to do on the JWST pipeline side of things.

I have not followed all the comments on this PR. There are quite a few that are unresolved. I attempted to resolve/hide some but it's unclear to me if they've all been addressed.

Let me see if I follow what this PR adds:
- `MiriWFSSSpecwcsModel` reference file datamodel (and associated schema: `specwcs_miri_wfss.schema`)
- `MIRIWFSSForwardDispersion` and `MIRIWFSSBackwardDispersion` transforms (and associated schemas, converters, extensions)
- a few unit tests that:
  - `test_miri_wfss_specwcs` that constructs a `MiriWFSSSpecwcsModel` and checks that attributes were assigned
  - `test_miri_wfss_roundtrip` that runs values through `MIRIWFSSForwardDispersion` and `MIRIWFSSBackwardDispersion` and checks those values are close enough
  - `test_miriwfss_backward_dispersion_single` that passes both multiple xys at a single wavelength and a single xy at multiple wavelengths through `MIRIWFSSBackwardDispersion`

I do not see any tests that write the transforms out to an ASDF file to test the converters. If I'm not overlooking these would you add a test(s) that write the transforms to an ASDF file then read them back in to confirm that the converters/schema/etc are working as expected?

As far as how the jwst work is handled. I can think of a few options:
1) Run the jwst regression tests (after adding the above requested unit test(s)) and confirm that nothing breaks. Once they pass, merge this PR and start the jwst work (which may reveal a need for further stdatamodels work that can be handled in follow-up PRs).
2) Start the jwst work now using the stdatamodels branch for this PR. Once the jwst work is ready, the merging of the 2 PRs (the jwst PR and this one) can be coordinated.

I vote for approach 1 if no one has any other opinions.

For the jwst work you might consider making a new regtest data directory, modifying it to contain the required files and then using that directory for testing.

src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/miri_wfss_dispersion-1.0.0.yaml
```diff
@@ -0,0 +1,48 @@
...
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/jwst_pipeline/miri_wfss_dispersion-1.0.0"
+title: >
+  MIRI WFSS dispersion model
+description: |
+  Supports one model:
+   - Given x, y, wave, order in direct image return x,y,wave,order in dispersed
+   - Given x, y, x0, y0, order in dispersed image returns x, y, wave, order in the direct image.
+anyOf:
```
braingram, line 11:
```suggestion
allOf:
```
Is there a reason for using `anyOf` instead of `allOf` here? Using `anyOf` would mean that if the instance passed validation again `transform-1.4.0` the reset of the schema contents would be ignored. `allOf` considers both subschemas (`transform-1.4.0` and the additional schema contents).

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,161 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+        ----------
+        orders : list
+            The list of orders which are available to the model. For MIRIWFSS orders = 1.
+        lmodels : list
+            The list inverse dispersion 1D polynomial models
+        xmodels : list
+            The list of 1D polynomial model in x.
+        ymodels : list[tuple]
+            The list of 2D polynomial models in y.
+            There are 3 2D polynomials to correct for  spatial-dependence.
```
braingram, line 2011:
```suggestion
        ymodels : list[list]
            The list of 2D polynomial models in y. The length of ymodels must be
            equal to the length of orders. Each element in ymodels is a sublist of
            3 2D polynomials to correct for spatial-dependence.
```
Am I understanding the code here that?
- orders should be a list of integers with length 1
- lmodels and xmodels are lists of 1D polynomials with lengths equal to the length of orders (N=1)
- ymodels is a list with length equal to the length of orders (N=1) and containing elements that are 3 item long lists containing 2D polynomials

Also, I'm confused by `correct for spatial-dependence`. Correct for spatial-dependence of what? Are there some more details you can add here to help me understand what these are used for?

emolter: I can help here.  This all looks correct to me.

Orders being a list makes sense for the other instruments where you can have multiple spectral orders, each of which has their own trace function.  I think it's better for MIRI to keep orders as a list for consistency in call structure between the instruments (but I could be convinced otherwise)

lmodels, xmodels, and ymodels are all functions of the trace parameter `t`, such that `lmodels` defines wavelength(t), `xmodels` defines `x(t)`, and `ymodels` defines `y(t)`.  These functions can have arbitrary complexity, where more complex would be a better representation of the true spectral trace.  It turns out that the three functions need to have different amounts of complexity in order to approximate the "true" trace.

Specifically, the `lmodel` and `xmodel` can be approximated by a simple 1st order polynomial in `t` for MIRI, but the `ymodel` is more complicated.  The `ymodel` depends not only on `t` but also on the `x0, y0` pixel location in the direct image that you're dispersing into the grism frame.  So each of the coefficients `C0`, `C1`, and `C2` of the second-order polynomial in `t` themselves have dependence on the direct-image position, so the full functional form looks something like `y_dispersed = C0(x0, y0) + C1(x0, y0)*t + C2(x0, y0)*t**2`

src/stdatamodels/jwst/transforms/tests/test_transforms.py
```diff
@@ -639,8 +640,117 @@ def test_niriss_grism_roundtrip(direction):
...
+    assert yi.size == y0.size
+
+    # test roundtrip to file- test specific values
+    tmp_file = tmp_path / "miri_wfss.asdf"
+    asdf.AsdfFile({"model": model}).write_to(tmp_file)
+    with asdf.open(tmp_file) as af:
+        assert af["model"].xmodels[0].c0 == xmodels[0].c0
+        assert af["model"].xmodels[0].c1 == xmodels[0].c1
+        # now test the full model
+        assert_model_equal(model, af["model"])
```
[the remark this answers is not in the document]
jemorrison: done

src/stdatamodels/jwst/transforms/models.py
jemorrison: @emolter Thank you for explaining that. It is a bit complex. Let me see if I can add more details

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -459,6 +460,83 @@ def to_fits(self):
...
+
+        Parameters
+        ----------
+        init : str, None, optional
+            File name for input file in ASDF format from which to initialize the model.
+        displ : list of `~astropy.modeling.Model`
+            MIRI WFSS wavelength dispersion model
+        dispx : list of `~astropy.modeling.Model`
+            MIRI WFSS row dispersion model
+        dispy : list of `~astropy.modeling.Model`
```
braingram, line 490:
```suggestion
        dispy : list of list of `~astropy.modeling.Model`
```
Isn't this a list of list of models?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -459,6 +460,83 @@ def to_fits(self):
...
+
+    def validate(self):
+        super(MiriWFSSSpecwcsModel, self).validate()
+        try:
+            assert isinstance(self.meta.input_units, (str, u.NamedUnit))
+            assert isinstance(self.meta.output_units, (str, u.NamedUnit))
+            assert self.meta.instrument.name == "MIRI"
+            assert self.meta.exposure.type == "MIR_WFSS"
+            assert self.meta.reftype == self.reftype
+            assert len(self.orders) == 1
```
braingram, line 529:
```suggestion
            n_orders = len(self.orders)
            assert n_order == 1
            assert len(self.displ) == n_orders
            assert len(self.dispx) == n_orders
            assert len(self.dispy) == n_orders
            assert all((len(sub_y) == 3 for sub_y in self.dispy))
```
Since the code has some expectations that aren't in the schema, how about adding these additional checks here?

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,164 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+
+        Parameters
+        ----------
+        orders : list
+            The list of orders which are available to the model. For MIRIWFSS orders = 1.
+        lmodels : list[1D models]
+            The list inverse dispersion 1D polynomial models. Use to determine the trace parameter.
+        xmodels : list[1D models]
+            The list of 1D polynomial models in x. It depends on the trace parameter
+        ymodels : list[2D models]
```
braingram, line 2009:
```suggestion
        ymodels : list[2D models]
```
Isn't this a list of lists? If so would you update this docstring?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
```diff
@@ -459,6 +460,83 @@ def to_fits(self):
...
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/specwcs_miri_wfss.schema"
+    reftype = "specwcs"
+
+    def __init__(
+        self,
+        init=None,
+        displ=None,
+        dispx=None,
+        dispy=None,
+        invdispl=None,
```
braingram, line 475: Where (if anywhere) is `invdispl` used?

braingram (commented):
Thanks for all your work on this. Just a few (possibly final comments/suggestions).

Would you ping previous commenters to resolve any comments/questions that have been addressed?

src/stdatamodels/jwst/transforms/tests/test_transforms.py
```diff
@@ -639,8 +640,157 @@ def test_niriss_grism_roundtrip(direction):
...
+    # These coefficients are taken from a real reference file.
+    mock_l = Polynomial1D(degree=1, c0=3.125, c1=10.893)
+    mock_invl = Polynomial1D(degree=1, c0=-0.28688148, c1=0.09180207)
+    mock_x = Polynomial1D(degree=1, c0=0.193314, c1=-1.993914)
+    # mock y models here parametrize a slightly curved trace
+    # causing enough of a shift in the x and y coordinates
+    # to ensure that the forward and backward models are not just
+    # identity models.
+
+    mock_y = Polynomial2D(degree=2, c0_0=0.1, c1_0=0.01)
```
emolter, line 657:
```suggestion
    mock_x = Polynomial1D(degree=1, c0=0.193314, c1=-1.993914)
    
    # mock y models here parametrize a slightly curved trace
    # causing enough of a shift in the x and y coordinates
    # to ensure that the forward and backward models are not just
    # identity models.
    mock_y = Polynomial2D(degree=2, c0_0=0.1, c1_0=0.01)
```
tiny spacing update to clarify what line the comment refers to

src/stdatamodels/jwst/transforms/models.py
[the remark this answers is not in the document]
emolter: This issue is now tracked by https://github.com/spacetelescope/stdatamodels/issues/542https://github.com/spacetelescope/stdatamodels/issues/542 and I think we can defer any changes until later

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,164 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+    """Calculate the dispersion extent of MIRI WFSS pixels."""
+
+    def __init__(self, orders, lmodels=None, xmodels=None, ymodels=None, name=None):
+        """
+        Initialize the model.
+
+        Parameters
+        ----------
+        orders : list
+            The list of orders which are available to the model. For MIRIWFSS orders = 1.
```
emolter, line 2004:
```suggestion
            The list of orders which are available to the model. For MIRIWFSS only order 1 is supported,
            so orders is expected to equal [1,].
```

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,164 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+
+        if (wavelength < 0).any():
+            raise ValueError("Wavelength should be greater than zero")
+        try:
+            iorder = self._order_mapping[int(order.flatten()[0])]
+        except KeyError as err:
+            raise ValueError("Specified order is not available") from err
+        # t is trace normalization parameters (it has values of 0 to 1)
+
+        t = self.lmodels[iorder](wavelength)
```
emolter, line 2060:
```suggestion

        # t is trace normalization parameters (it has values of 0 to 1)
        t = self.lmodels[iorder](wavelength)
```
spacing change to clarify which line the comment refers to

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1989,6 +1991,158 @@ def _evaluate_transform_guess_form(model, x=0, y=0, t=0):
...
+        xmodel = self.xmodels[iorder]
+        ymodel = self.ymodels[iorder]
+        dx = xmodel(t)
+        dy = ymodel[0](x, y) + dx * ymodel[1](x, y) + dx**2 * ymodel[2](x, y)
+        return x + dx, dy, x, y, order
+
+
+class MIRIWFSSForwardDispersion(_ForwardGrismDispersionBase):
+    """
+    Calculate the wavelengths of vertically dispersed MIRIWFSS data.
```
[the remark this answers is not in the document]
emolter: @jemorrison I assume that you've ensured this works as expected already?  If so please mark the conversation as resolved and/or comment here

emolter (commented):
This looks good to me now.  I added a few more very tiny suggestions.

I'm holding off on hitting the Approve button just because there are a few remaining unresolved conversations.  Once those are resolved this looks ready to go in.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
emolter: see https://github.com/spacetelescope/jwst/blob/f5e42aa105aa41b16faa4a3015afa4b6989fb517/jwst/assign_wcs/niriss.py#L456 where `lmodel` gets assigned to be `invdispl` for the backward transform

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
emolter: @braingram do these look like a good idea to also add for NIRISS and NIRCam?  If so, would you mind adding a comment to https://github.com/spacetelescope/stdatamodels/issues/542 ?

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
jemorrison: If we had more than 1 order it would be a list of lists. Since we just have 1 order it is just 1 model

src/stdatamodels/jwst/transforms/tests/test_transforms.py
emolter: so it's obvious this relates to the ymodel

src/stdatamodels/jwst/transforms/tests/test_transforms.py
```diff
@@ -639,8 +640,110 @@ def test_niriss_grism_roundtrip(direction):
...
+    forward = models.MIRIWFSSForwardDispersion(
+        orders,
+        lmodels=lmodels,
+        xmodels=xmodels,
+        ymodels=ymodels
+    )
+
+    backward = models.MIRIWFSSBackwardDispersion(
+        orders,
+        lmodels=invlmodels,
```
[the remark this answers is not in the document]
jemorrison: yes I do.

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
jemorrison: Nadia has approved it and I think I have addressed all of @emolter comments. I have just going through it again. I tried to resolve all the issues I made the changes to the code.

src/stdatamodels/jwst/transforms/models.py
braingram: Even with 1 order isn't it a list of lists?
https://github.com/spacetelescope/stdatamodels/blob/1577c46042d600f581117b19710f460cbebdd77c/src/stdatamodels/jwst/transforms/models.py#L2063-L2065
first indexes into to `self.ymodels` using the `iorder` to get `ymodel` and then indexes using 0, 1, 2 as in your example.

braingram:
I went through all my prior comments. The only outstanding ones are related to `dispy` docstrings claiming list of models when the tests and reference file have lists of lists of models. Once those are resolved would you run another set of regtests to confirm everything is still compatible with jwst main?

@emolter any outstanding comments?

jemorrison:
@braingram Is there a lingering issue about the list of lists that I have missed ?
As the new specwcs reference file we have lists of models not list of lists.

jemorrison:
@emolter I can the regression tests and I am getting failures in the wfss_contam_step. Are these expected  ? 
I do not think they are related to this PR.

braingram:
> @braingram Is there a lingering issue about the list of lists that I have missed ? As the new specwcs reference file we have lists of models not list of lists.

Thanks. Yes there are the lingering issues about the list of list of models. The file you shared `MIRI_WFSS_specwcs_July2025.asdf` contains:
```yaml
dispy:
- - !transform/polynomial-1.2.0
    coefficients: !core/ndarray-1.1.0
      byteorder: little
      datatype: float64
      shape: [3, 3]
      source: 1
    domain:
    - *id001
    - *id001
    inputs: [x, y]
    outputs: [z]
    window:
    - *id001
    - *id001
  - !transform/polynomial-1.2.0
    coefficients: !core/ndarray-1.1.0
      byteorder: little
      datatype: float64
      shape: [3, 3]
      source: 2
    domain:
    - *id001
    - *id001
    inputs: [x, y]
    outputs: [z]
    window:
    - *id001
    - *id001
  - !transform/polynomial-1.2.0
    coefficients: !core/ndarray-1.1.0
      byteorder: little
      datatype: float64
      shape: [3, 3]
      source: 3
    domain:
    - *id001
    - *id001
    inputs: [x, y]
    outputs: [z]
    window:
    - *id001
    - *id001
```
This is a list of list of models for `dispy`. This is consistent with the code that [provides a list of tuple of models](https://github.com/spacetelescope/stdatamodels/blob/c300c599258a1a97a196226369fbb486fe208ab3/src/stdatamodels/jwst/datamodels/tests/test_specwcs_miriwfss_datamodel.py#L51) in the tests and the validate method which check for a [list of list of models](https://github.com/spacetelescope/stdatamodels/blob/c300c599258a1a97a196226369fbb486fe208ab3/src/stdatamodels/jwst/datamodels/wcs_ref_models.py#L535). However these are inconsistent with the [docstrings](https://github.com/spacetelescope/stdatamodels/blob/c300c599258a1a97a196226369fbb486fe208ab3/src/stdatamodels/jwst/datamodels/wcs_ref_models.py#L490) which claim a list of models.

braingram:
> @emolter I can the regression tests and I am getting failures in the wfss_contam_step. Are these expected ? I do not think they are related to this PR. @braingram regression test results at: https://github.com/spacetelescope/RegressionTests/actions/runs/16757097875

Those failures look like they're from out-of-date tags on your stdatamodels fork. The stdatamodels version is dynamically assigned based on the newest tag. Your fork [currently has no tags](https://github.com/jemorrison/stdatamodels/tags) so stdatamodels defaults to a 0.1... version which is less than the version used from some regtest files (1.10.1 for the failing wfss tests). This results in a warning from asdf which jwst then turns into an error. One way to fix this is to push tags up to your fork. The following assumes that:
- you are in your local clone
- you have a remote called `upstream` that points to https://github.com/jemorrison/stdatamodels/tags

If the above assumptions hold run the following to get tags from upstream and push them to your fork:
```
git fetch upstream --tags
git push --tags
```

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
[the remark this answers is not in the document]
jemorrison: Putting a hold on this idea for the global wfss updates

src/stdatamodels/jwst/transforms/models.py
braingram: Sounds great!

src/stdatamodels/jwst/datamodels/wcs_ref_models.py
braingram: list of tuples of models or list of list of models sounds good to me.

braingram (approved):
Thanks! Let's get this merged and put it first in the queue to avoid further rebases/updates.

PR #497, opened 2025-06-11
Vectorize evaluation of NIRCAMBackwardGrismDispersion

### LINKED ISSUE #496: NIRCAMBackwardGrismDispersion invdisp_interp doesn't work on 2-D array input
https://github.com/spacetelescope/jwst/pull/9220 is refactoring the `wfss_contam` step to minimize the number of looped calls to WCSs and model transforms.  Part of this refactor includes attempting to pass 2-D arrays as input to the `imgxy_to_grismxy` transform; that transform is primarily composed of the `BackwardGrismDispersion` model.  For NIRISS, 2-D arrays are handled fine, and it appears that the transform just evaluates the input `lmodel`, `xmodel`, and `ymodel` directly.  For NIRCam, 2-D arrays fail to pass through `invdisp_interp`, which looks to be an attempt to derive an inverse transform from the forward transform `lmodel`, `xmodel` and `ymodel`.  The offending line is https://github.com/spacetelescope/stdatamodels/blob/493babacd365f8bb8df1fb276bdd0e939bc1d42e/src/stdatamodels/jwst/transforms/models.py#L1584 where `np.interp` requires 1-D inputs but they are now 2-D.

It's not clear to me whether we should just simply modify the lines around that one to handle 2-D input, perhaps by choosing a smarter interpolation function or adding a `for` loop, or if `NIRCAMBackwardGrismDispersion` can somehow be made to look more like `NIRISSBackwardGrismDispersion` (which I guess requires us to derive an empirical backward transform or something?)

emolter:
Partially Resolves [JP-4023](https://jira.stsci.edu/browse/JP-nnnn)

Closes #496 

This PR vectorizes the computation of the inverse dispersion.  In the old version of the code, the forward transform was evaluated at several points `t0`, then an interpolation function was used to find the 

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/transforms/models.py
+++ b/src/stdatamodels/jwst/transforms/models.py
@@ -1588,23 +1588,7 @@ class NIRCAMBackwardGrismDispersion(Model):
             return f
 
         def _optimize_function(t, x0, y0, wavelength):
-            """
-            Find the value of the trace function that matches the wavelength.
-
-            Parameters
-            ----------
-            t : float
-                The pixel offset value to optimize
-            x0, y0 : float
-                Source object x-center, y-center
-            wavelength : float
-                The target wavelength to match
-
-            Returns
-            -------
-            float
-                The squared difference between the trace function output and the wavelength
-            """
+            """Find value of trace function that matches the wavelength."""  # numpydoc ignore=RT01
             xr = trace_function(t, x0, y0)
             return (xr - wavelength) ** 2
 
--- a/src/stdatamodels/jwst/transforms/models.py
+++ b/src/stdatamodels/jwst/transforms/models.py
@@ -1576,7 +1576,7 @@ class NIRCAMBackwardGrismDispersion(Model):
             trace_function = partial(_trace_quadratic, model=model)
 
         else:
-            # Handle legacy versions of the trace mode
+            # Handle legacy versions of the trace model
             t0 = np.linspace(0.0, 1.0, 40)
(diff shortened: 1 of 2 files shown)


emolter:
regression tests here https://github.com/spacetelescope/RegressionTests/actions/runs/15620801101

downstream test failures are unrelated

braingram:
> regression tests here https://github.com/spacetelescope/RegressionTests/actions/runs/15620801101
> 
> downstream test failures are unrelated.. @braingram can you help me understand those downstream failures? I'm the one who made the PR in JWST that broke this... was I supposed to put a `__init__.py` into that test folder or something?

Not quite sure. Would you open a separate issue? It looks like a jwst (not stdatamodels) issue. I just skimmed it (about all I have time for right now) and I'm confused why the fixtures are imported. A __init__.py in `jwst/datamodels/utils/tests` might fix it but why are there wfss specific fixtures in there?

emolter:
Tagging @Russell-Ryan as another potential reviewer (if you have time)

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1519,6 +1526,13 @@ def evaluate(self, x, y, wavelength, order):
         order : int
             Output spectral order, same as input
         """
+        if isinstance(x, (int, float)):
```
tapastro, line 1529: Could this be `np.atleast_1d`?

PR #499, opened 2025-06-16
JP-1866: Use stpipe parent log

melanieclarke:
Partially resolves [JP-1866](https://jira.stsci.edu/browse/JP-1866)

Replace all local logs with names that start with "stpipe", to allow stpipe to manage log configuration.

## Tasks

</details>

### DIFF
--- a/changes/499.misc.rst
+++ b/changes/499.misc.rst
@@ -1,0 +1,1 @@
+Use "stpipe" as the root logger in all modules.
--- a/src/stdatamodels/dynamicdq.py
+++ b/src/stdatamodels/dynamicdq.py
@@ -2,9 +2,7 @@ import numpy as np
 
 import logging
 
-log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
-log.addHandler(logging.NullHandler())
+log = logging.getLogger("stpipe.stdatamodels")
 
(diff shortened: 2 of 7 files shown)


melanieclarke:
Regtests with jwst on main:
https://github.com/spacetelescope/RegressionTests/actions/runs/15688924445

braingram:
Is there a way with this new approach to "allow-list" a set of loggers? I'm thinking instead of prefixing everything with "stpipe" if we inform stpipe (perhaps via an entry point) that all "stdatamodels" loggers are to be logged we don't need to change the logger names here.

melanieclarke:
> Is there a way with this new approach to "allow-list" a set of loggers?

We could do that, but it would require a change to stpipe, which I think we don't want to do before the build deadline on Friday.  We would like to get some changes in before then, though, because logging is currently broken in Python 3.13 (https://github.com/spacetelescope/jwst/issues/9544).

Maybe we can discuss more in the follow up refactor for stpipe (https://github.com/spacetelescope/stpipe/pull/238).  If we decide to go that way, we could revert the changes here.

melanieclarke:
Setting this back to draft: log refactoring for stpipe-downstream packages will be deferred to next build.

PR #500, opened 2025-06-17
JP-4032: Add requested FGS metadata to primary header

emolter:
Resolves [JP-4032](https://jira.stsci.edu/browse/JP-4032)

Closes [spacetelescope/jwst#9541](https://github.com/spacetelescope/jwst/issues/9541)

This PR adds a few new keywords related to FGS navigation to the core schema, making them available in the primary header.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
@@ -1726,6 +1726,41 @@ properties:
             title: Observatory UTC time at guider function end
             type: string
             fits_keyword: GF_END
+          fgs_aperture:
+            title: Aperture used for guide star acquisition
+            type: string
+            fits_keyword: FGS_APER
+            blend_table: True
+          fgs_ideal_x:
+            title: 
+            type: float
+            fits_keyword: GSIDL_X
+            blend_table: True
+          fgs_ideal_y:
+            title:
+            type: float
+            fits_keyword: GSIDL_Y
+            blend_table: True
+          fgs_v2_ref:
+            title: "[arcsec] Telescope V2 coord of reference point"
+            type: number
+            fits_keyword: FGSV2REF
+            blend_table: True
+          fgs_v3_ref:
+            title: "[arcsec] Telescope V3 coord of reference point"
+            type: number
+            fits_keyword: FGSV3REF
+            blend_table: True
+          fgs_v3yangle:
+              title: "[deg] Angle from V3 axis to Ideal y axis"
+              type: number
+              fits_keyword: FGSV3YAN
+              blend_table: True
+          fgs_vparity:
+            title: Relative sense of rotation between Ideal xy and V2V3
+            type: integer
+            fits_keyword: FGS_VPAR
+            blend_table: True
           data_start:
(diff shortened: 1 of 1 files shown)


src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1726,6 +1726,41 @@ properties:
...
+          fgs_v3_ref:
+            title: "[arcsec] Telescope V3 coord of reference point"
+            type: number
+            fits_keyword: FGSV3REF
+            blend_table: True
+          fgs_v3yangle:
+              title: "[deg] Angle from V3 axis to Ideal y axis"
+              type: number
+              fits_keyword: FGSV3YAN
+              blend_table: True
```
tapastro, line 1758: Slightly off indentation here.

src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1726,6 +1726,41 @@ properties:
...
+          fgs_vparity:
+            title: Relative sense of rotation between Ideal xy and V2V3
+            type: integer
+            fits_keyword: FGS_VPAR
+            blend_table: True
```
tapastro, line 1763: These four entries don't match the keywords that SDP will populate - they are using V2_REF, V3_REF, V3I_YANG and VPARITY. Because these match the keyword entries in `wcsinfo` for SCI extensions, it may be useful to specify the fits_hdu for these entries as PRIMARY. I don't have great intuition on if this will cause issues, but this would match the update made in the grit merge request filed by Linda against JWSTKD-436. We can see if RTs find problems.

emolter: They match what's in the Excel spreadsheet for FITS keyword... for future reference how do I find what is _actually_ needed myself?  It sounds like you looked at the grit merge request?  When I try to follow the link I get a 404 page not found

src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
emolter: thanks for catching that

src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
emolter: ok yeah I agree it's a slightly different case and probably the linked issue should not be relevant.
how does https://github.com/spacetelescope/stdatamodels/pull/500/commits/519373c717b7886254d538caed8bc86f9ac78d86 look to you?

emolter:
from @rcooper295 :

> Hi Ned, the original justification for creating these effectively duplicate keywords was because the wcsinfo versions of those keywords were tied to the SCIENCE extension, which isn't part of the level 3 AMI products, so it created an extra empty "SCI" extension just for storing those keywords. If moving meta.ami.v3yangle and meta.ami.vparity to the meta.guidestar part of the tree wouldn't cause the creation of an empty extension, that should be fine. I think it would just require minor changes to the AMI3 code where the output datamodel is populated, though I haven't checked exactly where.

tapastro (approved):
LGTM

PR #507, opened 2025-06-25
Remove memmap option

### LINKED ISSUE #506: Opening models that contain unsigned data does not work with `memmap=True`
Attempting to call `datamodels.open(filename, memmap=True)` on any model with unsigned data raises
```
ValueError: ERROR loading embedded ASDF: Cannot load a memory-mapped image: BZERO/BSCALE/BLANK header keywords present. Set memmap=False.
```

Apparently `memmap=None` works differently from both `memmap=True` and `memmap=False` in astropy fits, such that if using the Astropy default value of `memmap=None`, the `dq` array will not be memory-mapped but arrays of other dtypes will.  See https://github.com/astropy/astropy/issues/7717#issuecomment-411461267 

I see a few ways of fixing this in stdatamodels:
- Make it so `memmap=True` in stdatamodels actually sets `memmap=None` in `fits.open`
- Document very clearly that `memmap=None` is typically the best-practice way to achieve memory mapping.  This seems weird but is apparently in-line with the Astropy API
- Do not expose the `memmap` option at all, and keep it off all the time, which is already the default.  I actually somewhat like this option because the pipeline never actually uses `memmap` in a non-default way, and I have my doubts that users do it either, especially given that it's broken for any model with a DQ array

emolter:
Closes #506 

This PR hard-codes `memmap` to `False` when reading FITS and ASDF, and emits a warning when a user attempts to use memory mapping.

## Tasks

</details>

### DIFF
--- a/changes/507.removal.rst
+++ b/changes/507.removal.rst
@@ -1,0 +1,1 @@
+Drop support for memory mapping when opening datamodels


drlaw1558:
LG2M

emolter:
This PR needs to wait until https://github.com/spacetelescope/jwst/pull/9593 is merged; otherwise the warnings are emitted whenever a ModelContainer is initialized

melanieclarke:
Looks good, pending clean regtests.

Did this one have an associated JP ticket?  I know we asked INS if it was okay to remove the argument, did we hear a final word on that?

emolter:
regtests are clean

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -107,7 +114,7 @@ def open(init=None, guess=True, memmap=False, **kwargs):  # noqa: A001
             return ModelContainer(init, **kwargs)
 
         elif file_type == "asdf":
-            asdffile = asdf.open(init, memmap=memmap)
+            asdffile = asdf.open(init, memmap=False)
```
melanieclarke, line 117: I just remembered about the ASDF weirdness with memmap and lazy_load.  Do we need to also set lazy_load = False here and in model_base to make sure the data is really read in?

melanieclarke: That's fair.  I can file a separate ticket to consider whether lazy_load should be False in datamodel.open for ASDF files.

PR #509, opened 2025-06-27
Map extrafits data

### LINKED ISSUE #490: extra_fits should use mappings to data arrays instead of copying it
Now that we have decided not to deprecate or remove `extra_fits` (#295 ) we should improve the way `extra_fits` works.

While loading from fits, when an extension is encountered that is not in the datamodel but contains a data array, currently that data array gets copied to the ASDF tree (as a data block) and is also retained in the hdulist.  It is therefore duplicated both in memory and in the output file on save.

Instead, `extra_fits` should be loaded in the same way as regular hdus: header information should be duplicated into the asdf extension (relatively small memory cost), while data arrays should be represented in the asdf extension as a mapping to the real data living in the hdulist.

emolter:
Closes #490 

This PR fixes a bug where array-like data that was not mapped to the schema (i.e. "extra fits" data) was duplicated in the ASDF extension.  This PR updates the methods for saving `extra_fits` to treat data arrays in the same way that schema-mapped data are already handled; that is, by replacing arrays with links to those same arrays in the main hdulist.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -509,6 +509,17 @@ def _save_extra_fits(hdulist, tree):
                     continue
                 hdu.header.append((key, val, comment), end=True)
 
+    def callback(node):
+        if isinstance(node, (np.ndarray, NDArrayType)):
+            # replace extra_fits arrays with special source values
+            # that represent links to that array in the FITS extension
+            for hdu_index, hdu in enumerate(hdulist):
+                if hdu.data is not None and node is hdu.data:
+                    return _create_tagged_dict_for_fits_array(hdu, hdu_index)
+        return node
+
+    return treeutil.walk_and_modify(tree, callback)
+
 
 def _save_history(hdulist, tree):
     if "history" not in tree:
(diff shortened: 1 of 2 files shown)


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/15928774518

src/stdatamodels/fits_support.py
```diff
@@ -509,6 +509,17 @@ def _save_extra_fits(hdulist, tree):
...
+    def callback(node):
+        if isinstance(node, (np.ndarray, NDArrayType)):
+            # replace extra_fits arrays with special source values
+            # that represent links to that array in the FITS extension
+            for hdu_index, hdu in enumerate(hdulist):
+                if hdu.data is not None and node is hdu.data:
+                    return _create_tagged_dict_for_fits_array(hdu, hdu_index)
+        return node
+
+    return treeutil.walk_and_modify(tree, callback)
```
braingram, line 521: Can we avoid this extra `walk_and_modify`? It deep copies the tree:
https://www.asdf-format.org/projects/asdf/en/latest/api/asdf.treeutil.walk_and_modify.html#asdf.treeutil.walk_and_modify

emolter: What's the equivalent function for not deep copying and modifying in-place?  `iter_tree`?

emolter:
started a regression test rerun just in case https://github.com/spacetelescope/RegressionTests/actions/runs/15928774518

src/stdatamodels/fits_support.py
braingram: That's habit forming... Looks good to me!

PR #511, opened 2025-06-27
Fix SlitModel init bugs

### LINKED ISSUE #296: Calling `datamodels.open` on a `SlitModel` clears `wht`
Minimal example:

```python
import numpy as np

m0 = dm.SlitModel((27, 1299))
m0.wht[:] = 1

m1 = dm.open(m0)
np.testing.assert_equal(m1.wht, 0)
```

emolter:
Resolves [JP-3709](https://jira.stsci.edu/browse/JP-3709)

Closes https://github.com/spacetelescope/jwst/issues/8712
Closes #296 

This PR simplifies the `__init__` of `SlitModel` to fix two bugs caused by the unnecessary custom handling there.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/slit.py
+++ b/src/stdatamodels/jwst/datamodels/slit.py
@@ -1,5 +1,4 @@
 from .model_base import JwstDataModel
-from .image import ImageModel
 
 
 __all__ = ["SlitModel", "SlitDataModel"]
@@ -46,33 +45,10 @@ class SlitDataModel(JwstDataModel):
     schema_url = "http://stsci.edu/schemas/jwst_datamodel/slitdata.schema"
 
     def __init__(self, init=None, **kwargs):
-        if isinstance(init, (SlitModel, ImageModel)):
-            super(SlitDataModel, self).__init__(init=None, **kwargs)
-            self.data = init.data
-            self.dq = init.dq
-            self.err = init.err
-            if init.hasattr("area"):
-                self.area = init.area
-            if init.hasattr("var_poisson"):
-                self.var_poisson = init.var_poisson
-            if init.hasattr("var_rnoise"):
-                self.var_rnoise = init.var_rnoise
-            if init.hasattr("var_flat"):
-                self.var_flat = init.var_flat
-            if init.hasattr("wavelength"):
-                self.wavelength = init.wavelength
+        super(SlitDataModel, self).__init__(init=init, **kwargs)
+        if kwargs:
             for key in kwargs:
                 setattr(self, key, kwargs[key])
-
-            if init.meta.hasattr("wcs"):
-                self.meta.wcs = init.meta.wcs
-            else:
-                self.meta.wcs = None
-        else:
-            super(SlitDataModel, self).__init__(init=init, **kwargs)
-            if kwargs:
-                for key in kwargs:
-                    setattr(self, key, kwargs[key])
 
 
 class SlitModel(JwstDataModel):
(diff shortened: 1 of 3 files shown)


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/15933522296 

All the regtest failures are due to zero-populated area extensions no longer being present, which is the expected behavior.

src/stdatamodels/jwst/datamodels/slit.py
```diff
@@ -117,29 +94,7 @@ class SlitModel(JwstDataModel):
...
-                self.meta.wcs = None
-        else:
-            super(SlitModel, self).__init__(init=init, **kwargs)
-            if kwargs:
-                for key in kwargs:
-                    setattr(self, key, kwargs[key])
+        super(SlitModel, self).__init__(init=init, **kwargs)
+        if kwargs:
+            for key in kwargs:
+                setattr(self, key, kwargs[key])
```
melanieclarke, line 100: Is this kwarg handling necessary, or could we just remove the init function entirely for this model?  Looking at the base init implementation, it looks like it expects many kwargs that should not necessarily be set as attributes in the model.  Looking through the init functions for other models here, it does not like like this is common usage.

melanieclarke (commented):
Glancing through init functions, it looks like IFUImageModel might need the same changes.

src/stdatamodels/jwst/datamodels/ifuimage.py
```diff
@@ -36,24 +35,6 @@ class IFUImageModel(JwstDataModel):
     schema_url = "http://stsci.edu/schemas/jwst_datamodel/ifuimage.schema"
 
     def __init__(self, init=None, **kwargs):
```
melanieclarke, line 37: I think we can now just remove this init function too - it's just calling the parent init.

melanieclarke (commented):
Thanks for the updates - can we add the IFUImageModel changes to the change log, and add a test for it too?

src/stdatamodels/jwst/datamodels/ifuimage.py
emolter: I don't think that's strictly true.  It's also doing
```
        # Implicitly create arrays
        self.dq = self.dq
        self.err = self.err
```

The usefulness of that is up for debate, but without those lines zero-filled DQ and err arrays won't be created on init.

melanieclarke (dismissed):
Looks great!  Thanks!

emolter:
The failure was happening because the change to `SlitModel` init was affecting `MultiSlitModel.__getitem__`.  For example when you do
```
model = MultiSlitModel()
model.extend([list_of_slits])
model[0]
```
the expected behavior is that `model[0]` is an `ObjectNode` internally but returns a `SlitModel`, and that functionality was taking advantage of the custom keyword handling in `SlitModel.__init__`.  I updated `MultiSlitModel.__getitem__` to hopefully handle this, but I'm not sure it's the best solution. 

Restarting regtests https://github.com/spacetelescope/RegressionTests/actions/runs/16147175783 to see if this causes any failures.

melanieclarke:
> the expected behavior is that `model[0]` is an `ObjectNode` internally but returns a `SlitModel`, and that functionality was taking advantage of the custom keyword handling in `SlitModel.__init__`. I updated `MultiSlitModel.__getitem__` to hopefully handle this, but I'm not sure it's the best solution.

That is obscure, and unexpected!  We could put back the kwargs handling, since it turns out it does have a purpose after all.

emolter:
hopefully final regtest run after restoring custom kwarg handling to slitmodel
https://github.com/spacetelescope/RegressionTests/actions/runs/16150220251

src/stdatamodels/jwst/datamodels/tests/test_open.py
```diff
@@ -292,14 +294,27 @@ def test_open_kwargs_fits(tmp_path):
             model.validate()
 
 
-def test_memmap_deprecation(tmp_path):
```
melanieclarke, line 295: What's going on with this test?  Was it accidentally removed here?

emolter:
Thanks for your attention to detail!  It would have been a shame to drop the other test

src/stdatamodels/jwst/datamodels/tests/test_models.py
```diff
@@ -258,16 +258,17 @@ def test_slit_from_image():
...
     slit = SlitModel(im)
+    assert isinstance(slit, SlitModel)
     assert_allclose(im.data, slit.data)
     assert_allclose(im.err, slit.err)
     assert hasattr(slit, "wavelength")
     assert slit.meta.instrument.name == "MIRI"
 
     im = ImageModel(slit)
-    assert type(im) is ImageModel
+    assert isinstance(im, ImageModel)
```
braingram, line 268: Did something require changing `type(im) is ImageModel` to this `isinstance(im, ImageModel)`. The former is more strict (doesn't allow `ImageModel` subclasses).

emolter: no there was no reason, my IDE just didn't like it. will revert

emolter: https://github.com/spacetelescope/stdatamodels/pull/511/commits/14d4073152489323a9f157ca39f669ea1d2d24a0

Issue #390, opened 2025-01-31
Increase test coverage to remove some added `noqa`s


braingram:
https://github.com/spacetelescope/stdatamodels/pull/383 introduced a large number of new style checks. Some checks fail on uncovered code. Rather than update this code with a risk of introducing a bug (for unsafe and manual fixes) that would be uncaught `noqa` comments were added.

Increase test coverage for `noqa` lines of code to allow their safe removal.

Related to https://github.com/spacetelescope/stdatamodels/issues/388

Issue #389, opened 2025-01-31
Use custom exception in `src/stdatamodels/jwst/transforms/models.py`


braingram:
Bare exceptions are raised in `models.py`:
https://github.com/spacetelescope/stdatamodels/blob/b49f5f437e49322c5fcc48bd47052779c6a066de/src/stdatamodels/jwst/transforms/models.py#L702
https://github.com/spacetelescope/stdatamodels/blob/b49f5f437e49322c5fcc48bd47052779c6a066de/src/stdatamodels/jwst/transforms/models.py#L840

Use a custom exception here with an informative name and docstring.

PR #519, opened 2025-07-01
Unit tests and small bugfixes for transform models

### LINKED ISSUE #388: Increase test coverage for `src/stdatamodels/jwst/transforms/models.py`
This appears to be uncovered by unit tests making it impossible to check for bugs introduced by refactoring and style changes.

Investigate adding unit tests for this model to allow the style to be updated.

xref: https://github.com/spacetelescope/stdatamodels/pull/383/files/1ce97af164bc6292c34c1a4bcdc35a0e79452181#r1937738621

### LINKED ISSUE #517: NIRCam forward row/column grism dispersion models make trace functions inappropriately
On [this line](https://github.com/spacetelescope/stdatamodels/blob/cef08ead136ad73f8034ac51d52f451870c86ea3/src/stdatamodels/jwst/transforms/models.py#L1398) (and the corresponding line for NIRCAMForwardRowGrismDispersion) the number of interpolation points in `t` is set according to the size of the input `y` array.  This doesn't make sense and causes errors.

For example, if all you need to do is transform a single x,y data point, then the interpolation will only know about the trace function at a single point: the line `t0 = np.linspace(0.0, 1.0, t_len)` returns `np.array(0.0)`.  Since the "whole point" of the `invdisp_interp` method is to return an interpolation function, this is not very helpful.  A funny side-effect is that the accuracy of the solution depends on the number of input `x` and `y` data points.

This bug is causing the forward and inverse transforms to fail to round-trip except if `x` and `y` are large arrays.

emolter:
Closes #388 
Closes #517 

This PR:

- Adds unit test coverage for the transforms in `jwst/transforms/models.py`.
- Improves the way that sampling the trace function is handled by WFSS transforms, including a bugfix for #517 
- Drops partial support for NIRCam models that have trace functions x(t), y(t) whose coefficients are 1st order polynomials in (x0, y0).  These are supported in the forward transform but fail to round-trip on main due to being unsupported by the backward transform.
- Adds scipy as a testing dependency, which is required for the use of certain Astropy models.

This PR is a prerequisite to handling #389 #390 and #520 

## Tasks

</details>

### DIFF
--- a/changes/519.bugfix.rst
+++ b/changes/519.bugfix.rst
@@ -1,0 +1,1 @@
+Fix incorrect trace function sampling in NIRCam forward grism dispersion models
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ test = [
     "pytest-doctestplus",
     "crds>=11.17.1",
     "requests",
+    "scipy>=1.9.2", # same as astropy
 ]
 docs = [
     "sphinx",
--- a/src/stdatamodels/jwst/transforms/models.py
+++ b/src/stdatamodels/jwst/transforms/models.py
@@ -1457,6 +1457,7 @@ class NIRCAMBackwardGrismDispersion(Model):
         inv_ymodels=None,
         name=None,
         meta=None,
+        sampling=40,
     ):
         """
         Initialize the model.
@@ -1484,6 +1485,8 @@ class NIRCAMBackwardGrismDispersion(Model):
             Name of the model
         meta : dict
             Unused
+        sampling : int, optional
+            Number of sampling points in t to use; these will be linearly interpolated.
         """
         self._order_mapping = {int(k): v for v, k in enumerate(orders)}
         self.orders = orders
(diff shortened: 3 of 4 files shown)


emolter:
Unit tests are failing because apparently the Astropy `Tabular1D` model, which is used by the NIRISS transforms, requires scipy: it's giving `ModuleNotFoundError: Tabular model requires scipy.`.

What do you think is the best solution here?  I think it would be feasible to refactor this to use `np.interp1d` instead of this tabular model, although there's some concern that, just like NIRCam, this will make it difficult to ensure this is vectorized.  Or, we can make `scipy` a dependency of stdatamodels, or a testing dependency.  Or we can just xfail the test...

emolter:
jwst regtests https://github.com/spacetelescope/RegressionTests/actions/runs/16008468504

emolter:
@jemorrison it looks like you're not taggable as a reviewer of stdatamodels PRs, so just pinging you here - feel free to review if you want

PR #528, opened 2025-07-17
Refactor grism dispersion

### LINKED ISSUE #520: Refactor grism dispersion transform models to reduce repeated code
The six grism dispersion transforms, `NIR[ISS/CAM][Forward/Backward][Row/Column]GrismDispersion` contain a lot of repeated code that could be easily refactored.  Currently the repeats are an annoyance for development, since many changes have to be applied several times across multiple model types.

Specifically:

- All six of these have the same inputs to `__init__` and could share a base class.
- The two forward transforms, row and column, for each instrument have extremely similar `evaluate` methods, with the only difference being that `x`, `x0`, and `xmodel` are swapped with `y`, `y0`, and `ymodel`.
- There are various attempts to handle multiple input model types, e.g. different orders of polynomial, spread across these models.  They could be unified into one helper function.  This has sort of been attempted for `NIRCAMBackwardGrismDispersion` already with the `assess_model` function, but this could be generalized and then reused elsewhere

This refactor would be much safer once https://github.com/spacetelescope/stdatamodels/issues/388 is addressed

emolter:
Closes #520 

This PR refactors the grism transforms to make the following improvements:

- Makes all the grism dispersion classes share a base class.
- Unifies the `evaluate()` methods for the two forward transforms, row and column, for each instrument, such that they call the same underlying code.
- Makes all evaluation of polynomial models (in the forward direction) call the same underlying code
- Drops support for unused and partially-implemented types of input models.  Specifically, attempted handling of polynomial models whose coefficients depended on `x` but not `y` was removed (there was never attempted handling for the `y` but not `x` case)

## Tasks

</details>

### DIFF
--- a/changes/529.bugfix.rst
+++ b/changes/529.bugfix.rst
@@ -1,0 +1,1 @@
+Updated units in WFSS tables to adhere to FITS standard.
--- a/changes/530.bugfix.rst
+++ b/changes/530.bugfix.rst
@@ -1,0 +1,1 @@
+Update units to match FITS standard.
--- a/src/stdatamodels/jwst/_tests/test_schemas.py
+++ b/src/stdatamodels/jwst/_tests/test_schemas.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 
 import asdf
+import astropy.units
 import pytest
 import yaml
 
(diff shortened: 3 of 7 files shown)


braingram:
> I can't decide if a changelog entry would be helpful here or not.

My vote is for one. Not that I think you're introducing bugs but if one crawls in it will be easier to back track possible causes with a changelog.

emolter:
@jemorrison feel free to review this one too.  I think it should make your job somewhat easier as you look at adding the MIRI transforms

emolter:
> Not that I think you're introducing bugs but if one crawls in it will be easier to back track possible causes with a changelog.

I wouldn't be so sure... :)

at least these models have some test coverage now though

src/stdatamodels/jwst/transforms/models.py
```diff
@@ -1659,7 +1527,12 @@ def _trace_quadratic(t, x0, y0, model):
             y0 = y0[0].flatten()
             wavelength = wavelength[:, 0].flatten()
 
-        wave_grid = trace_function(t0, x0, y0)
+        trace_function = partial(_poly_with_spatial_dependence, model=model)
+
+        # Create a grid of t0, x0, and y0 values
+        tt, yy = np.meshgrid(t0, y0, indexing="ij")
+        xx = np.meshgrid(t0, x0, indexing="ij")[1]
+        wave_grid = trace_function(tt, xx, yy)
```
emolter, line 1535: This change basically makes it so that the inputs all have the same shape as the eventual output before attempting to eval the trace function, rather than putting matrix multiplication into the trace function itself.  I did check the memory usage effects of this for my test NIRCam dataset, and the memory footprint of `wfss_contam` was nearly identical

PR #540, opened 2025-07-28
Deprecate None, array, tuple, and hdulist as valid inits to open()

### LINKED ISSUE #536: Disallow unused and confusing init types for datamodel open
Right now it's possible to do either of:
`dm.open("foo.fits", data=np.ones((3, 3), dtype="f4"))  # data is 3x3` or
`dm.ImageModel("foo.fits", data=np.ones((3, 3), dtype="f4"))  # data is 3x3`

These calls don't make a lot of sense: I don't think there's any good reason to open a model from file and reassign its data in the same line.  For `DataModel.__init__` it's desired that e.g. `dm.ImageModel(data=np.ones((3, 3))` should continue to work.

For `stdatamodels.jwst.datamodels.open`, in my view, it should NOT work to do any of the following:

- `dm.open(data=np.ones((3, 3)))  # returns <JwstDataModel> on main`
- `dm.open(np.ones((3,3)))  # returns <ImageModel(3, 3)> on main`
- `dm.open((3,3))  # returns <ImageModel(3, 3)> on main`
- `dm.open()  # returns <JwstDataModel> on main`

Plus it's pretty unexpected/unclear behavior that the first two of these calls return different model types...

emolter:
Closes #536 

This PR removes some init options from `open()`, basically making it so that `open` is no longer able to create datamodels from scratch; the desired model type should be instantiated directly instead.

This PR also makes it so initializing a model from a file-like input and also assigning its data arrays at the same time is no longer supported.

Examples of syntax that will be removed are in the issue comments.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/tests/test_models.py
+++ b/src/stdatamodels/jwst/datamodels/tests/test_models.py
@@ -76,10 +76,10 @@ def test_skip_fits_update(make_models, which_file):
     path = make_models[which_file]
     with fits.open(path) as hduls:
         hduls[0].header["exp_type"] = "FGS_DARK"
+        hduls.writeto(path, overwrite=True)
 
-        with pytest.warns(DeprecationWarning, match="Passing fits.HDUList to open is deprecated"):
-            with datamodels.open(hduls) as model:
-                assert model.meta.exposure.type == "FGS_DARK"
+    with datamodels.open(path) as model:
+        assert model.meta.exposure.type == "FGS_DARK"
 
 
 def test_asnmodel_table_size_zero():
--- a/src/stdatamodels/jwst/datamodels/util.py
+++ b/src/stdatamodels/jwst/datamodels/util.py
@@ -193,12 +193,6 @@ def open(init=None, guess=True, **kwargs):  # noqa: A001
 
         return ModelContainer(init, **kwargs)
 
-    elif isinstance(init, (dict, asdf.AsdfFile)):
-        raise TypeError(
-            f"Unsupported type for init argument to open {type(init)}. "
-            "To initialize a datamodel from a model tree, use the DataModel constructor directly."
-        )
-
     else:
         raise TypeError(f"Unsupported type for init argument to open {type(init)}")
 


emolter:
rerun after most recent commits https://github.com/spacetelescope/RegressionTests/actions/runs/16578708310

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -53,19 +66,26 @@ def open(init=None, guess=True, **kwargs):  # noqa: A001
...
-    """
-    from . import model_base
 
+    Notes
+    -----
+    The ``open`` function is primarily intended for opening files and creating models from them,
+    and is not intended for creating models from scratch. Use the DataModel constructors directly
+    for that purpose instead. Init types of None, shape tuple, and numpy array
+    are deprecated and will raise a TypeError in the future.
+    Use the DataModel constructors directly instead, i.e. :class:`JwstDataModel` for a generic model
```
melanieclarke, line 84: This description is slightly repetitive - can you rephrase?  Maybe remove the first "Use the DataModel constructors directly" sentence.

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -53,19 +66,26 @@ def open(init=None, guess=True, **kwargs):  # noqa: A001
...
+    or one of the many model subclasses (e.g. :class:`ImageModel`, :class:`MultiSlitModel`)
+    for specific applications. None, shape tuple, and numpy array are all valid inputs to those
+    constructors, and see the documentation for each model class for details on how to use them.
```
melanieclarke, line 87:
```suggestion
    constructors. See the documentation for each model class for details on how to use them.
```

src/stdatamodels/model_base.py
```diff
@@ -274,6 +274,17 @@ def __init__(
         # Initialize class dependent hidden fields
         self._no_asdf_extension = False
 
+        if (init is not None) and (not is_array) and (not is_shape) and (len(kwargs)) > 0:
+            warnings.warn(
+                "Unrecognized keyword arguments passed to DataModel.__init__. "
+                "DataModel init is file-like (e.g. filename, dict, HDUList, AsdfFile, etc.) "
+                "but keyword arguments were also passed, which are assumed to be attempting to "
```
melanieclarke, line 281: Can you add a note about this to the kwargs docs, above?  It should probably say that kwargs should only be passed if init is None.

melanieclarke (commented):
Looks good from my perspective, just a couple requests for docstring clarification.

src/stdatamodels/jwst/datamodels/util.py
emolter: whoops, yeah thanks for catching that. see https://github.com/spacetelescope/stdatamodels/pull/540/commits/22962b58872c0409405b101f2bc1be7333c8161b

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -151,6 +193,12 @@ def open(init=None, guess=True, **kwargs):  # noqa: A001
 
         return ModelContainer(init, **kwargs)
 
+    elif isinstance(init, (dict, asdf.AsdfFile)):
```
braingram, line 196: I'm ok if we allow the `else` below to catch these cases (`dict`, `asdf.AsdfFile`). Having a special `if` for these to issue a slightly different error message doesn't seem worth the added code and tests.

src/stdatamodels/jwst/datamodels/tests/test_models.py
```diff
@@ -77,8 +77,9 @@ def test_skip_fits_update(make_models, which_file):
     with fits.open(path) as hduls:
         hduls[0].header["exp_type"] = "FGS_DARK"
 
-        with datamodels.open(hduls) as model:
-            assert model.meta.exposure.type == "FGS_DARK"
+        with pytest.warns(DeprecationWarning, match="Passing fits.HDUList to open is deprecated"):
```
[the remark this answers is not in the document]
emolter: see https://github.com/spacetelescope/stdatamodels/pull/540/commits/deb63a19a5cd6e359a6b6bc76866b12bca237d6d

I think the right thing to do is to save and re-load the model, but I could be wrong about the original intent of that test so let me know if I misinterpreted it

braingram (approved):
LGTM. Thanks!

PR #552, opened 2025-08-11
Close hdulist during datamodel init and remove from file references

### LINKED ISSUE #300: close hdulist instead of keeping a reference in `DataModel`
When opening a `DataModel` stored in a `FITS` file:
https://github.com/spacetelescope/stdatamodels/blob/4041d5f75c2736f4c4131028232d62573c83d732/src/stdatamodels/model_base.py#L238-L244
`DataModel` will create a `FileReference` for the `hdulist` and keep it open. This appears to be unnecessary (adding a `hdulist.close()` on line doesn't break any jwst unit tests). This makes sense because the call to `from_fits` has to deal with the various quirks of `FITS` and results in loading all contents from the `FITS` file.

If we instead just close the hdulist a `FileReference` is no longer needed.

If we take this one step further and use `lazy_load=False` with `asdf` we could entirely remove `FileReference` and `DataModel.close` which would simplify the `DataModel` and `ModelContainer` api (by no longer needing to close or keep track of closing files).

emolter:
Closes #300 

This PR makes it so that that when a model is initialized or opened from .fits, the `hdulist` is closed during `DataModel.__init__` instead of being left open until datamodel close.

There is no good reason to keep the hdulist around, since converting the hdulist to asdf handles all fits-related quirks during init anyway.

Regression tests show that there are no pipeline failures, and that the memory usage of many pipelines is modestly but nontrivially lower with these changes for many of the memory-tracked tests: https://github.com/spacetelescope/RegressionTests/actions/runs/16886681733

## Tasks

</details>

### DIFF
--- a/changes/552.misc.rst
+++ b/changes/552.misc.rst
@@ -1,0 +1,1 @@
+When opening models from a .fits file, close the FITS HDUList on init instead of retaining a FileReference to it
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -235,15 +235,18 @@ class DataModel(properties.ObjectNode):
 
             if file_type == "fits":
                 hdulist = fits.open(init, memmap=False)
-                hdulist = self._migrate_hdulist(hdulist)
-                asdffile = fits_support.from_fits(
-                    hdulist,
-                    self._schema,
-                    self._ctx,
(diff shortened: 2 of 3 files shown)


braingram (approved):
LGTM.

PR #555, opened 2025-08-12
JP-3887: Add photometric time dependence tables

melanieclarke:
Partially resolves [JP-3887](https://jira.stsci.edu/browse/JP-3887)

Corresponding jwst PR is: https://github.com/spacetelescope/jwst/pull/9736

This PR adds time dependence coefficient tables to all photom models, except MirMrsPhotomModel, which already has time coefficient extensions that need a different form from the others.  MirImgPhotomModel also had a time coefficients table, but needs to move to the new format to allow more flexibility.

3 optional tables are added to each model: timecoeff_linear, timecoeff_exponential, and timecoeff_powerlaw.  Any, all, or none may be present, and the photom code will compose the correction from the appropriate functional forms.

I also added some validation code for all photom models (except MIRI MRS) to verify that if present, the timecoeff extensions match the phot_table rows.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/photom.py
+++ b/src/stdatamodels/jwst/datamodels/photom.py
@@ -29,16 +29,23 @@ class _PhotomModel(ReferenceFileModel):
         """
         Validate optional timecoeff extensions.
 
+        The phot_table extension must be present.
+
         If present, timecoeff_linear, timecoeff_exponential, and timecoeff_powerlaw
         extensions must match the length and descriptive columns present in the
         phot_table extension.
         """
         super().validate()
+
+        if not self.hasattr("phot_table"):
+            self.print_err("Model.phot_table is not present")
+            return
+
         try:
             timecoeff = ["timecoeff_linear", "timecoeff_exponential", "timecoeff_powerlaw"]
             for extension in timecoeff:
                 message = f"Model.phot_table and Model.{extension} do not match"
-                if self.hasattr("phot_table") and self.hasattr(extension):
+                if self.hasattr(extension):
                     table = getattr(self, extension)
                     assert len(self.phot_table) == len(table), message
 
--- a/src/stdatamodels/jwst/datamodels/tests/test_photom.py
+++ b/src/stdatamodels/jwst/datamodels/tests/test_photom.py
@@ -142,3 +142,24 @@ def test_invalid_timecoeff_mismatched_values(strict):
         # Empty extensions are not added
         assert not model.hasattr("timecoeff_linear")
         assert not model.hasattr("timecoeff_powerlaw")
+
+
+@pytest.mark.parametrize("strict", [True, False])
+def test_invalid_phot_table(strict):
+    """Test validation for missing phot_table."""
+    with mir_img_phot_hdulist() as phot_hdul:
+        del phot_hdul["PHOTOM"]
+        model = MirImgPhotomModel(phot_hdul, strict_validation=strict)
+        expected_message = "Model.phot_table is not present"
+        if strict:
+            with pytest.raises(ValueError, match=expected_message):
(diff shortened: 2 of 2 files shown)


melanieclarke:
Regression tests here:
https://github.com/spacetelescope/RegressionTests/actions/runs/16892547832

Attempt 1 shows one very small change to the background level value for one input file.  This appears to be precision related, from minor changes to the way the time dependence is calculated for MIRI imaging photom corrections: the background level fit in skymatch is sensitive to very small changes in the data.  There is otherwise no impact to the output data.

src/stdatamodels/jwst/datamodels/photom.py
```diff
@@ -18,7 +20,42 @@
 ]
 
 
-class FgsImgPhotomModel(ReferenceFileModel):
+class _PhotomModel(ReferenceFileModel):
```
emolter, line 23: This class appears to be causing docs build failures. It might need to be added into the nitpick ignores, or else not be underscore-protected

src/stdatamodels/jwst/datamodels/schemas/fgsimg_photom.schema.yaml
```diff
@@ -18,3 +18,51 @@ allOf:
...
+        description: Reference day in MJD
+      - name: year1value
+        datatype: float32
+        description: Relative throughput 1 year after t0
+      - name: tsoft
+        datatype: float32
+        description: Softening parameter for power law decline
+      - name: alpha
+        datatype: float32
+        description: Power law loss coefficient
```
emolter, line 68: Large fractions (all?) of this appears to be repeated for all the schemas.  Can it be made its own file, and a `$ref` used?

src/stdatamodels/jwst/datamodels/schemas/mirmrs_photom.schema.yaml
emolter, line 1: I assume it's fine to make this non backward-compatible?

src/stdatamodels/jwst/datamodels/schemas/fgsimg_photom.schema.yaml
melanieclarke: Alas, it cannot, as far as I know.  I took that approach first, but the science team wanted descriptive keywords to be included in the tables, to match the phot_table columns, and those are different for each mode.  I don't think there's a way to compose the datatype of a table from multiple sources.

src/stdatamodels/jwst/datamodels/photom.py
```diff
@@ -18,7 +20,44 @@
...
+        If present, timecoeff_linear, timecoeff_exponential, and timecoeff_powerlaw
+        extensions must match the length and descriptive columns present in the
+        phot_table extension.
+        """
+        super().validate()
+        try:
+            timecoeff = ["timecoeff_linear", "timecoeff_exponential", "timecoeff_powerlaw"]
+            for extension in timecoeff:
+                message = f"Model.phot_table and Model.{extension} do not match"
+                if self.hasattr("phot_table") and self.hasattr(extension):
```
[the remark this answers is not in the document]
melanieclarke: Sure, I can add that.  I thought about adding that check and decided to allow an empty photom file, since only the check on the time coeff tables was requested.  But you're right that it's definitely not a useful photom file without the phot_table, so we might as well check for it while we're validating.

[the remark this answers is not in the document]
melanieclarke: See [1ab79ec](https://github.com/spacetelescope/stdatamodels/pull/555/commits/1ab79ecc284ab006333fae8022895f7317198abd)

tapastro (approved):
👍

PR #567, opened 2025-08-15
Add `asdf_in_fits.to_hdulist` to add ASDF extension to an in memory or existing HDUList

### LINKED ISSUE #566: Add an option to `asdf_in_fits.write()` to return `HDUList` object as a memory object
I would like to write some ASDF data to a FITS file, but I don't want to create a written file on disk. It would be great if the `write` function had some parameter like `memory_only` to specify that the `HDUList` object should be returned only in memory.

braingram:
I also added a note to `asdf_in_fits` describing the API is mainly for non-pipeline uses.

Closes #566

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/asdf_in_fits.py
+++ b/src/stdatamodels/asdf_in_fits.py
@@ -7,6 +7,12 @@ from . import fits_support
 
 __all__ = ["write", "open", "to_hdulist"]
 
+# This API is intended to replace the removed asdf.AsdfInFits
+# for community (non-pipeline) usage. When considering changes
+# a wider search for usage beyond pipeline code and documentation
+# is recommended as well as longer deprecation periods that
+# aren't as closely linked to pipeline releases.
+
 
 def to_hdulist(tree, hdulist=None):
     """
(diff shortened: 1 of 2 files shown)


src/stdatamodels/asdf_in_fits.py
```diff
@@ -5,7 +5,34 @@
...
 from . import fits_support
 
-__all__ = ["write", "open"]
+__all__ = ["write", "open", "to_hdulist"]
+
+# This API is intended to replace the removed asdf.AsdfInFits
+# for community (non-pipeline) usage. When considering changes
+# a wider search for usage beyond pipeline code and documentation
+# is recommended as well as longer deprecation periods that
+# aren't as closely linked to pipeline releases.
```
emolter, line 14: Is it standard practice to have this in `#` comments instead of triple-quoted at the top of the file?

braingram: Thanks. I used a comment here so that it doesn't get auto-included in the RTD here:
https://stdatamodels--567.org.readthedocs.build/en/567/api.html#module-stdatamodels.asdf_in_fits
(which I think would be the case for a module level docstring/triple-quoted text at the top fo the file).

emolter (approved):

braingram:
Just in case I started a regtest run: https://github.com/spacetelescope/RegressionTests/actions/runs/16997005673

Issue #542, opened 2025-08-01
Miscellaneous small fixes to grism transform models


emolter:
While reviewing https://github.com/spacetelescope/stdatamodels/pull/493 a few issues were raised that are generically applicable to all the grism transform models, not just the new MIRI ones.  This issue is meant to ensure these useful suggestions don't get lost in the shuffle.
- Names of the transforms should be hard-coded, as there is no obvious use-case for renaming them
- Right now there is handling for setting attributes during `__init__` only if they are not None.  There doesn't seem to be any reason not to just let them be set to None.
- The NIRISS backward model uses the `lmodels` attribute to pass in a model that does the operation `t = model(wavelength)`. But that is really the inverse operation, so the attribute should be `inv_lmodels` instead.
- Docstrings can be improved, particularly the Parameters section of `__init__` for the various models.
- Some attributes of the specwcs should be made required in the schema.  This is already being applied for MIRI in the above-mentioned PR

PR #569, opened 2025-08-18
Improve documentation of grism transform models

emolter:
Partially addresses #542 

This PR improves the documentation of the grism transforms:

- Various improvements and clarifications have been added to the docstrings of the transform input parameters
- `stdatamodels.jwst.transforms` is now built into the public docs
- The transform models are now available through the namespace `stdatamodels.jwst.transforms`, e.g. `from stdatamodels.jwst.transforms import Slit2Msa`. They are still also available at `stdatamodels.jwst.transforms.models`.
- The `name` attributes of all the `GrismDispersion` transforms are now hard-coded, as there is no clear use-case for renaming them, and these names are sometimes used when the transform needs to be accessed from within a WCS object.

## Tasks

</details>

### DIFF
--- a/docs/source/jwst/transforms/index.rst
+++ b/docs/source/jwst/transforms/index.rst
@@ -5,12 +5,56 @@ Transforms
 .. toctree::
    :maxdepth: 2
 
+The :py:mod:`stdatamodels.jwst.transforms` submodule defines
+`Astropy-compatible models <https://docs.astropy.org/en/latest/modeling/index.html>`__
+that encode WCS transformations specific to JWST.
+
+Grism Transforms
+----------------
+
+In support of Wide-Field Slitless Spectroscopy (WFSS) modes, NIRISS, NIRCam, and MIRI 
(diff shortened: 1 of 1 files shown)


emolter:
@braingram no need to look at the details of the docstrings if you don't want to, but can you have a look at the structural changes to the docs and namespace here and let me know what you think?

emolter:
not fully, because I only changed the grism transform docstrings.  I think we will need to add some unit test coverage to the other ones in order to make sure that calls on numpy arrays actually works.  This already exists for the grism models.

tapastro (approved):
Looks good!

We can leave a more thorough explanation of how to use the transforms and interact with WCS frames to jwst docs updates (coordinated with the WCS documentation that Melanie proposed for hack day work) - but this is a good start, and greatly improves clarity for inputs/outputs and their shapes for the various transform classes. A good example of the sort of code that we should probably provide documentation for, even if it isn't explicitly part of the top-level API.

PR #577, opened 2025-08-29
Fix MIRIWFSSBackwardDispersion

nden:

Closes #

This PR addresses an issues with the transform from direct image to grism for MIRI WFSS mode.
It caused the bounding boxes of the spectra to be outside the image. 

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/transforms/models.py
+++ b/src/stdatamodels/jwst/transforms/models.py
@@ -2062,8 +2062,9 @@ class MIRIWFSSBackwardDispersion(_BackwardGrismDispersionBase):
         xmodel = self.xmodels[iorder]
         ymodel = self.ymodels[iorder]
         dx = xmodel(t)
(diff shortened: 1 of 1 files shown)


nden:
Since this is fixing code that's not released yet, does it need a change log?

jemorrison:
I agree with the adding dy. But I do not get the correct trace using t the way you did. I will send you my notebook how I test this

jemorrison:
The reason your plots have the spectra off is because in assign_wcs I add the source position to the dy determined values. I agree  it is better do it in the model (and I have local stdatamodels with that correction) . But the recommended change for the Backward transform is not consistent with how the polynomials were determined. Currently wavelength is not used in the fit to derive the polynomials in specwcs. I believe this was because for now they have just assumed a linear wavelength polynomial. The MIRI team now has real MIRIWFSS data - and improved methods are being determined.

nden:
Feel free to close this if you think it's incorrect.

PR #589, opened 2025-09-19
Remove deprecated V23ToSky and Rotation3D; add converter to UnsupportedConverters class

### LINKED ISSUE #546: Remove deprecated V23ToSky and Rotation3D
https://github.com/spacetelescope/stdatamodels/pull/539 deprecated these transforms in whatever version comes after stdatamodels 4.0.1.  This issue reminds us to remove it a few builds later

emolter:
Closes #546 

Relates to [JP-4086](https://jira.stsci.edu/browse/JP-4086)

Along with https://github.com/asdf-format/asdf/pull/1983, this PR implements a procedure for removing classes that have an ASDF converter.  These are typically WCS transforms.  Their removal requires special care because they are used to load files.

Say a file is created that has the `Foo` transform in its WCS.  On save, a serialized version of `Foo` (a `dict` of its input parameters plus an ASDF URI tag) will be saved.  On load, `asdf` will use the URI tag to find the instructions for how to convert `Foo` back into an Astropy-like transform; these instructions live in `stdatamodels` and are called the converter.

Now, let's remove the `Foo` class and its associated converter, then attempt to load the file.  Prior to this PR, a hard-to-understand exception would be raised by `gwcs`: `asdf` would fail to find the proper converter for `Foo`, so it would warn and just return the serialized version of it.  Then `gwcs` would attempt to incorporate that dict-like object into a WCS transform, fail to do so, and raise.

As of this PR, loading the file with default options will still raise an exception, but now it will be called `UnsupportedConverterError` and give an informative message for users explaining their options for accessing the data.  This is implemented by moving the URI tag that used to point to `Foo` into the `UnsupportedConverter` class instead. This has the added benefit of "squatting" on the tag and ensuring a new transform could never be written that points to that tag.

One of the user options for accessing their data, implemented with help from https://github.com/asdf-format/asdf/pull/1983, is to bypass the error and load the file anyway by setting the asdf config flag `warn_on_failed_conversion` to True.  With this option enabled, loading the file containing `Foo` will turn the `UnsupportedConversionError` into a warning, and the file will load successfully.  The user may then access their data and inspect the WCS, but the `Foo` part of the WCS will just look like a `dict` and the WCS object will not be callable.

This PR uses the machinery discussed above to remove the deprecated `V23ToSky` transform.  It also removes the deprecated `Rotation3D` transform that `V23ToSky` called, which is a bit more straightforward because `Rotation3D` did not have its own converter.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/transforms/converters/tests/test_unsupported.py
+++ b/src/stdatamodels/jwst/transforms/converters/tests/test_unsupported.py
@@ -5,8 +5,6 @@ from asdf.exceptions import AsdfConversionWarning
 import stdatamodels.jwst.datamodels as dm
 from stdatamodels.jwst.transforms.converters.jwst_models import UnsupportedConverterError
 
-"""Test attempting to load or save files with transforms that are no longer supported."""
-
 
 @pytest.fixture
 def model_with_unsupported_transform(tmp_path):


src/stdatamodels/jwst/transforms/converters/jwst_models.py
```diff
@@ -17,15 +18,56 @@
...
+    from the JWST pipeline, but which may still be present in old files.
+    When encountered, a warning will be issued and the WCS will not be deserialized.
+    """
+
+    tags = [
+        "tag:stsci.edu:jwst_pipeline/v23tosky-*",
+    ]
+    max_supported_version = [
+        "4.0.0",
+    ]
```
braingram, line 44:
```suggestion
    tag_to_max_version = {
        "tag:stsci.edu:jwst_pipeline/v23tosky-*": "4.0.0"
    }

    @property
    def tags(self):
        return self.tag_to_max_version.keys()
```

src/stdatamodels/jwst/transforms/converters/jwst_models.py
```diff
@@ -17,15 +18,56 @@
...
+            "We recommend downloading the latest reprocessed data from MAST, or else "
+            f"downgrading to stdatamodels version <= {self.max_supported_version[match_idx]}",
+        )
+        return
+
+    def _tag_idx(self, tag):
+        for i, pattern in enumerate(self.tags):
+            if tag.startswith(pattern[:-1]):
+                return i
+        return None
```
braingram, line 60:
```suggestion
    def max_version(self, tag):
        for tag_pattern, version in self.tag_to_max_version.items():
            if asdf.util.uri_match(tag_pattern, tag):
                return version
        raise ...
```

braingram (commented):
I think we can store the pattern/version pairs as a mapping and use an asdf util to do the match.

src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/v23tosky-0.7.0.yaml
```diff
@@ -1,41 +0,0 @@
...
-  coordinate system (V2,V3) to ICRS.
-
-examples:
-  -
-    - JWST V2, V3 to sky transform.
-
-    - |
-        !<tag:stsci.edu:jwst_pipeline/v23tosky-0.7.0>
-          angles: [-0.0193, -0.1432, -0.04, -65.60, 273.089]
-          axes_order: zyxyz
```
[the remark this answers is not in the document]
emolter: Thanks, yeah I was working on a similar test, I'm glad there is already a roughly well accepted way to write these tests.  I'll try to update it and let you know if I get stuck anywhere

src/stdatamodels/jwst/transforms/converters/tests/test_unsupported.py
```diff
@@ -0,0 +1,31 @@
...
+%YAML 1.1
+%TAG ! tag:stsci.edu:asdf/
+--- !core/asdf-1.1.0
+model: !<tag:stsci.edu:jwst_pipeline/v23tosky-0.7.0>
+  angles: [-0.0193, -0.1432, -0.04, -65.60, 273.089]
+  axes_order: zyxyz
+...
+    """
+    path = tmp_path / "model_with_unsupported_transform.asdf"
+    path.write_text(asdf_file)
```
emolter, line 22: I couldn't actually use your suggested asdf helper here in any way that was useful, because bytestrings are not a valid input to datamodels open.

This test is failing right now with
```asdf.exceptions.AsdfWarning: Unable to locate schema file for 'tag:stsci.edu:jwst_pipeline/v23tosky-0.7.0': 'http://stsci.edu/schemas/jwst_pipeline/v23tosky-0.7.0'```
I will continue trying to figure out why `open()` is getting far enough that it thinks it's supposed to validate against the schema - I had expected, and was hoping, that encountering this tag would more-or-less immediately raise the new error

Comment (braingram) [reply src/stdatamodels/jwst/transforms/converters/tests/test_unsupported.py:22]: Some benefits are:
- in case someone is inheriting from this schema the resource will still be available
- the URI will remain assigned to a valid resource/schema and
- the schema as a form of documentation will continue to be useful in describing the previously used transform

The downside would be:
- we have an extra file in our repository
- the example may not work if we unlink the schema and tag (so we may need to skip the example)

What do you think we should do?

Comment (emolter) [reply src/stdatamodels/jwst/transforms/converters/tests/test_unsupported.py:22]: > in case someone is inheriting from this schema the resource will still be available

If the transform is out-of-date and being removed, wouldn't we want to discourage this?

> the URI will remain assigned to a valid resource/schema

I thought the point of squatting on the tag is that we ensure that the URI points to something, and that thing is clearly marked as unsupported.  Or am I confusing the schema URI with the tag?

> the schema as a form of documentation will continue to be useful in describing the previously used transform

Ok, sure, but just comparing this with removing code, there is no expectation that the documentation will remain around for an expired deprecation.

> we have an extra file in our repository

I'm not necessarily opposed to this.  The maintenance cost feels like it's little to zero, as I doubt we would ever do something that meaningfully modifies all schemas in the repository.  I view this differently than code, where versions, pre-commit rules, etc might change regularly.

> the example may not work if we unlink the schema and tag (so we may need to skip the example)

Which example?


Overall I'm not sure I have a strong opinion.  I lean toward trying to remove it, just because I would have assumed that if the transform is deprecated we'd want to discourage folks from using the schema in any way.

Comment (braingram) [reply src/stdatamodels/jwst/transforms/converters/tests/test_unsupported.py:22]: > If the transform is out-of-date and being removed, wouldn't we want to discourage this?

Discouraging the usage sounds great but removing the resource will cause any dependent schemas to fail. I doubt there is any dependent schema in this case.

> I thought the point of squatting on the tag is that we ensure that the URI points to something, and that thing is clearly marked as unsupported. Or am I confusing the schema URI with the tag?

There are 2 URIs, one for the [schema](https://github.com/spacetelescope/stdatamodels/blob/e7215bf570ce53d46c8708de5b16e6cf647933ef/src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-0.7.0.yaml#L89), one for/is the [tag](https://github.com/spacetelescope/stdatamodels/blob/e7215bf570ce53d46c8708de5b16e6cf647933ef/src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-0.7.0.yaml#L88).

> Ok, sure, but just comparing this with removing code, there is no expectation that the documentation will remain around for an expired deprecation.

It looks like we don't even have the schemas as part of the documentation. 

> I'm not necessarily opposed to this. The maintenance cost feels like it's little to zero, as I doubt we would ever do something that meaningfully modifies all schemas in the repository. I view this differently than code, where versions, pre-commit rules, etc might change regularly.

Agreed and given this is a stable schema it shouldn't ever be changed.
 
> Which example?

https://github.com/spacetelescope/stdatamodels/blob/e7215bf570ce53d46c8708de5b16e6cf647933ef/src/stdatamodels/jwst/transforms/resources/schemas/stsci.edu/jwst_pipeline/v23tosky-0.7.0.yaml#L17-L20
This example is validated as part of the pytest runs via pytest-asdf-plugin.

> Overall I'm not sure I have a strong opinion. I lean toward trying to remove it, just because I would have assumed that if the transform is deprecated we'd want to discourage folks from using the schema in any way.

I don't have a strong opinion here either. I can see the benefits and downsides to keeping or removing the schema. My main concern is making sure that we continue to allow old files to be opened (which I believe we can do either way).

If there are no strong opinions I'd say we remove the schema in this case (given the age, lack of documentation, etc).

**Review by melanieclarke — COMMENTED** [review]
I'm not sure I totally follow the details here, but for what it's worth, in principle, this design looks good to me.

A question: the Slit2MsaLegacy transform will need similar removal at some point, but right now it is handled internally to the Slit2MsaConverter, which is still needed to handle the now-standard Slit2Msa transform.  Would we just then remove it, like you have done for Rotation3D?  Or do you imagine some other process for issuing an error for an unsupported transform without removing the converter that currently handles it?

File: src/stdatamodels/jwst/transforms/resources/manifests/jwst_transforms-0.7.0.yaml [hunk 5]
```diff
@@ -86,7 +86,7 @@ tags:
     - given incident angle and wavelength compute the refraction/diffraction angle
     - given an incident angle and a refraction angle compute the wavelength
 - tag_uri: "tag:stsci.edu:jwst_pipeline/v23tosky-0.7.0"
-  schema_uri: http://stsci.edu/schemas/jwst_pipeline/v23tosky-0.7.0
+  # schema_uri: http://stsci.edu/schemas/jwst_pipeline/v23tosky-0.7.0 # No longer supported
```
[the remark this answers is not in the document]
emolter: Brett and I went back and forth on this a bit, see https://github.com/spacetelescope/stdatamodels/pull/589#discussion_r2368572610 

If you don't drop the `schema_uri` then you need to keep the schema around too, because validation happens before conversion.  Keeping the schema around is an alternative option here, which we could implement if you feel strongly.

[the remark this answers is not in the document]
braingram: What fails without this modification?

src/stdatamodels/jwst/transforms/converters/jwst_models.py
```diff
@@ -17,15 +18,65 @@
...
+    """
+    Class to catch legacy transforms that are no longer supported.
+
+    This converter contains tags for all converters that have been removed
+    from the JWST pipeline, but which may still be present in old files.
+    When encountered, an error will be raised and the WCS will not be deserialized.
+    """
+
+    types = []
+    tag_to_max_version = {"tag:stsci.edu:jwst_pipeline/v23tosky-*": "4.0.0"}
```
braingram:
```suggestion
    tag_to_max_version = {"tag:stsci.edu:jwst_pipeline/v23tosky-*": "4.1.0"}
```
Should this be 4.1.0?

pyproject.toml
```diff
@@ -9,7 +9,7 @@ classifiers = [
   "Programming Language :: Python :: 3",
 ]
 dependencies = [
-  "asdf>=3.3.0",
+  "asdf @ git+https://github.com/asdf-format/asdf.git@main",
```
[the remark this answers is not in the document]
emolter: I realized we're going to need to run the unit test only when certain asdf versions are available.  Sorry if this is a noob question, but is there a generally accepted way to do that?

src/stdatamodels/jwst/transforms/converters/tests/test_unsupported.py
```diff
@@ -0,0 +1,50 @@
+import asdf
+import pytest
+from asdf.exceptions import AsdfConversionWarning
+
+import stdatamodels.jwst.datamodels as dm
+from stdatamodels.jwst.transforms.converters.jwst_models import UnsupportedConverterError
+
+"""Test attempting to load or save files with transforms that are no longer supported."""
```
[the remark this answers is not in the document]
emolter: what do you mean?

I do use Copilot but this was me.  Do you want me to remove it?

PR #598, opened 2025-10-03
Remove private _kwtool module

braingram:
This PR removes the private `stdatamodels.jwst._kwtool` module which has been migrated to grit.

I opened this PR now because mast is frequently returning timeouts when attempting to download the keyword dictionary. 
https://github.com/spacetelescope/stdatamodels/actions/runs/18233600855/job/51922625599?pr=593#step:10:559

This download is only required for a few tests for the private kwtool so it seems worthwhile to remove it now to avoid these test failures.

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/18234011018

There is one stdatamodels unit test that was updated to allow for this removall:
https://github.com/spacetelescope/stdatamodels/blob/85bdcb2f2e1388b10170f01384b04fce2e70ffea/src/stdatamodels/jwst/datamodels/tests/test_schema.py#L113-L115

## Tasks

</details>

### DIFF
--- a/docs/source/jwst/datamodels/schemas.rst
+++ b/docs/source/jwst/datamodels/schemas.rst
@@ -11,6 +11,16 @@ This package contains schemas of the following types:
 Data model schemas
 ------------------
 
+.. note::
+
+   For users familiar with the
+   `JWST Keyword Dictionary <https://mast.stsci.edu/portal/Mashup/Clients/jwkeywords/index.html>`_
+   it is important to note that the keyword dictionary is not used by stdatamodels.
+   Instead the datamodel schemas contain independent descriptions of the JWST files.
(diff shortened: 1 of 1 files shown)


docs/source/jwst/kwtool/keyword_dictionary.rst
emolter: While it seems like most of this should be migrated to that private tool, I wonder if we should maintain some documentation about where the keyword dictionary can be found and how their structure differs from stdatamodels

emolter (commented):
LGTM overall, one question about whether we should keep around a bit of the documentation that was removed

penaguerrero:
LGTM!

PR #625, opened 2025-12-02
JP-4184: Relax t range for NIRCam WFSS grism transforms

### LINKED ISSUE #10012: Artifacts at ends of simulated sources for wfss_contam
_Issue [JP-4184](https://jira.stsci.edu/browse/JP-4184) was created on JIRA by [Ned Molter](https://jira.stsci.edu/secure/ViewProfile.jspa?name=emolter):_

See images.  There is an unphysical spike in flux in the simulations coming out of wfss_contam.  This should be investigated - could be a pipeline issue or a photom reference file issue (or something else).

 

!Screenshot 2025-11-11 at 2.19.50 PM.png!

!image.png!

emolter:
Resolves [JP-4184](https://jira.stsci.edu/browse/JP-4184)

Closes spacetelescope/jwst#10012

This PR fixes a bug where artifacts were present at the short-wavelength ends of NIRCam WFSS spectra.  This was caused by a mismatch between the stdatamodels definition of the model transforms for NIRCam and how they are calibrated when input into the specwcs.  Specifically, NIRCam's dispersion models are calibrated such that the `t` parameter can be slightly outside the range `[0,1]`.

Prior to this PR, when a best-fitting `t`-value was found outside the `[0,1]` range, it was clipped to the edge of that range.  This was causing many flux contributions from different wavelengths to end up in the same pixel, which appeared like the artifacts shown on the JIRA ticket.  The team has instructed us that the limits `[-0.1, 1.1]` are what should be used.  When those limits are used instead, the problem goes away - the different wavelengths are properly dispersed into different locations on the detector.

A side-effect of this change is that the bounding boxes for the cutouts in `extract_2d` as well as the background mask are longer by a few hundred pixels in the along-dispersion direction for NIRCam.

## Tasks

</details>

### DIFF
--- a/changes/625.bugfix.rst
+++ b/changes/625.bugfix.rst
@@ -1,0 +1,1 @@
+Fix incorrect parametrization of trace models for NIRCam WFSS transforms
--- a/src/stdatamodels/jwst/transforms/models.py
+++ b/src/stdatamodels/jwst/transforms/models.py
@@ -1190,7 +1190,8 @@ class _NIRCAMForwardGrismDispersion(_ForwardGrismDispersionBase):
         if len(dx.shape) == 2:
             dx = dx[0, :]
 
-        t0 = np.linspace(0.0, 1.0, self.sampling)
+        # NIRCam t functions have been calibrated such that being just outside (0, 1) is okay
+        t0 = np.linspace(-0.1, 1.1, self.sampling)
 
         # handle multiple inverse model types
         if isinstance(model, (ListNode, list, tuple)):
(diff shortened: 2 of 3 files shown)


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/19870500114

tapastro:
How will this PR interact with the fix delivered as part of new reference files? The new files plan to restructure the delivered transforms such that they should span 0<t<1, right? Will this PR cause the new transform files to go "out of bounds"?

PR #638, opened 2025-12-15
MIRI RSCD reference file format update and adding RSCD step results to core schema.

jemorrison:
Is needed for JP-4119 [JP-4119](https://jira.stsci.edu/browse/JP-4119)

Closes #

This PR updates the RSCD reference file to allow the first integration entries in the table.
It also updates the core_schema adding RSCD output information on the number of groups skipped for an integration. This value can change from the reference file value depending on the number of groups in an integration.

## Tasks
  Regression test passed:
https://github.com/spacetelescope/RegressionTests/actions/runs/22241413789

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/rscd.py
+++ b/src/stdatamodels/jwst/datamodels/rscd.py
@@ -11,24 +11,8 @@ class RSCDModel(ReferenceFileModel):
     ----------
     rscd_group_skip_table : numpy table
         Reference table for RSCD correction baseline correction
-        A table with 3 columns that set the number of groups to skip for each
+        A table with 4 columns that set the number of groups to skip for each
         subarray and readpatt
-    rscd_gen_table : numpy table
-        Reference table for RSCD correction enhanced correction
-        A table with 5 columns that sets up general parameters for the enhanced
-        correction
-    rscd_int1_table : numpy table
-        Reference table for RSCD correction enhanced correction
-        A table with 7 columns that sets up correction parameters for the enhanced
-        correction for integration 1 based on subarray, readpatt, even or odd row
-    rscd_int2_table : numpy table
-        Reference table for RSCD correction enhanced correction
-        A table with 7 columns that sets up correction parameters for the enhanced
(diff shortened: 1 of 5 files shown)


src/stdatamodels/jwst/datamodels/rscd.py
```diff
@@ -1,8 +1,40 @@
...
+    for ext in hdulist:
+        # RSCD tables are usually in the first BinTableHDU
+        if ext.data is None or not isinstance(ext, fits.BinTableHDU):
+            continue
+
+        table_data = ext.data
+        if "group_skip1" not in table_data.dtype.names:
+            # Create the missing column.
+            # Defaulting to 1 as per the baseline requirement.
+            new_col = np.full(table_data.shape[0], 1, dtype=">i4")
```
braingram: Based on https://github.com/spacetelescope/stdatamodels/pull/638#issuecomment-3922571646 isn't filling this with 1 incorrect?

How about using the "group_skip" value for "group_skip1"?

Or alternatively, use a fill value of -1 and update the rscd step to perform no correct and issue an error when "group_skip1" is -1.

braingram (commented):
The failure looks unrelated and due to crds failing to correctly download files for the new context. Let's see if it goes away after this PR is updated.

braingram (approved):
Thanks for the updates on this. It LGTM.

I was able to open both an existing rscd file(https://jwst-crds.stsci.edu/browse/jwst_miri_rscd_0015.fits) and what I think is the new file (using the file Hunter shared with me that was failing to upload to the current test server).

braingram:
> @braingram What is the 'tpn' ? I just asked the MIRI team (who delivered the file to TEST) what was an associated CRDS ticket

tpns are sort the equivalent "schemas" for crds. I expect without a tpn change that the file delivery will throw errors. I vote for not delaying this PR for that work but it may delay delivering a file.

tapastro (approved):
Looks good, thanks Jane!

TPN updates discussed here: https://jira.stsci.edu/browse/CRDS-1064

PR #641, opened 2025-12-30
Move "legacy" schemas to separate directory, add missing schemas to docs

### LINKED ISSUE #618: Separate schemas that won't be updated
https://github.com/spacetelescope/stdatamodels/pull/589 further marked the end of maintaining the v23tosky schema (which was already out-of-date). During previous schema updates (https://github.com/spacetelescope/stdatamodels/pull/258) this schema was incorrectly seen as "maintained". It would be beneficial to separate "maintained" and "unmaintained" schemas/resources to make it explicit what schemas should be updated when upstream changes are introduced (and perhaps to adapt tests to consider if a schema is "maintained").

One option would be to move unmaintained schemas to a new directory and register that as a resource mapping. This should (and can) be done in a way that does not change schema URIs or contents making it have essentially no user-facing changes but improve the maintainability of this package.

braingram:
Closes #618

Moves "legacy" (no longer updated) transforms schemas to a separate directory and documents that these will not be updated.

Also adds tests to check that the docs contain mention of:
- the latest "legacy" schemas (to document that they will not be updated)
- the latest non-legacy schemas

The tests for latest non-legacy schemas revealed a few undocumented schemas. These were added to the docs.

jwst regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/20607402499
all pass

## Tasks

</details>

### DIFF
--- a/tests/test_schema_docs.py
+++ b/tests/test_schema_docs.py
@@ -5,13 +5,13 @@ import re
 
 import pytest
 
-repo_path = pathlib.Path(__file__).parent.parent
-docs_path = repo_path / "docs"
-transforms_docs = docs_path / "source" / "jwst" / "transforms" / "index.rst"
-transforms_resources_path = repo_path / "src" / "stdatamodels" / "jwst" / "transforms" / "resources"
-transforms_schemas_path = transforms_resources_path / "schemas" / "stsci.edu" / "jwst_pipeline"
-legacy_transforms_schemas_path = (
-    transforms_resources_path / "legacy" / "schemas" / "stsci.edu" / "jwst_pipeline"
+REPO_PATH = pathlib.Path(__file__).parent.parent
+DOCS_PATH = REPO_PATH / "docs"
+TRANSFORMS_DOCS = DOCS_PATH / "source" / "jwst" / "transforms" / "index.rst"
+TRANSFORMS_RESOURCES_PATH = REPO_PATH / "src" / "stdatamodels" / "jwst" / "transforms" / "resources"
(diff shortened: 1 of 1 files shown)


tests/test_schema_docs.py
```diff
@@ -0,0 +1,106 @@
...
+import re
+
+import pytest
+
+repo_path = pathlib.Path(__file__).parent.parent
+docs_path = repo_path / "docs"
+transforms_docs = docs_path / "source" / "jwst" / "transforms" / "index.rst"
+transforms_resources_path = repo_path / "src" / "stdatamodels" / "jwst" / "transforms" / "resources"
+transforms_schemas_path = transforms_resources_path / "schemas" / "stsci.edu" / "jwst_pipeline"
+legacy_transforms_schemas_path = (
```
emolter: nitpick, but isn't it more normal to have these defined constants in all caps?

.ruff.toml
```diff
@@ -58,7 +58,6 @@ ignore = [
   "TRY003", # prevents custom exception messages not defined in exception itself.
   "ISC001", # single line implicit string concatenation. formatter recommends ignoring this.
   "PTH123", # use Path.open instead of open
-  "UP038",  # isinstance with | instead of ,
```
emolter: I'm fine with removing this, and I think we may have already done so for jwst, but just curious - why was this done now? I don't see any `isinstance` calls in the PR

braingram: The rule is no longer present in ruff. When I ran pre-commit for this PR ruff warned me that this rule no longer exists so I removed it.

tests/test_schema_docs.py
braingram: Good call. I'll update the PR.

braingram: How does 4e78314f797a1cdc2fb0cf6a1141209a7c37b243 look?

PR #646, opened 2026-01-08
JP-3959: Add the WCS tag to datamodels schemas

### LINKED ISSUE #7020: `wcs` is missing from tab-completion of `meta`
To clarify the problem here I'm going to describe the workflow that lead to it:
1. I loaded a jwst fits file as downloaded from the archive
2. I loaded it with ``dm = jwst.datamodels.open('filename')`
3. I thought to myself "I know there's some way to get at the wcs object.  What is it again?"
4. I tried `dm.wcs`, which didn't work.
5. Something in my memory said "Oh, I think it's in meta somewhere"
6. I did `dm.meta.w<tab>`.
7. The only thing that came up was `wcsinfo`, so I spent a while looking in `wcsinfo` for how I get the GWCS object
8. I got fairly irritated, but then something jumped into my head that I'd seen this problem before, so I tried `dm.meta.gwcs`, and then on a whim, `dm.meta.wcs`.
9. The right answer was `dm.meta.wcs`

The core issue is that it was hard for me to figure out how to access the GWCS object, despite being someone who knows it should be there. So I see a few possible (not mutually-exclusive) solutions:

1. Make sure `wcs` shows up in tab-completion for the metadata node.  I tried to find where this gets set (it's probably a property or something), but couldn't find it on a quick glance.  If someone can give me a pointer to where `wcs` actually gets stuck into the metadictionary, I might be able to cook up a PR to fix this.
2. Consider also setting up some link so that `dm.wcs` works.  This might require WG review or something, but it seems important enough to justify IMHO. It's also a lot closer to how astropy's NDData, ndcube, etc, treat wcs.
3. Document much more clearly that `dm.meta.wcs` (or `dm.wcs` if you like my option 2) is the way to get the true canonical wcs.  It doesn't show up *anywhere* in https://jwst-pipeline.readthedocs.io/en/latest/jwst/datamodels/index.html which I think means its effectively un-discoverable if you don't know it's there already.

nden:
Resolves [JP-3959](https://jira.stsci.edu/browse/JP-3959)

Closes spacetelescope/jwst#7020

This PR addresses an oversight in the schemas where the WCS tag is missing from the schemas and never validated. The side effect of this change is that TAB-expansion now shows the `wcs` attribute.

## Tasks

</details>


tests/jwst/datamodels/test_read_metadata.py
```diff
@@ -19,9 +19,9 @@
 @pytest.fixture
 def recursive_tree():
     """Create a recursive tree to substitute for a WCS object"""
-    w = {"inputs": "test"}
-    w["outputs"] = w
-    return w
+    from gwcs import examples
```
emolter, line 22: should this be moved out of the test?

emolter (commented):
I'm on board with this philosophically.

Changelog fragment is needed, and regression tests.

I believe our next release in June is already slated to be major, 6.0, because we have expired a bunch of deprecations.  @melanieclarke @tapastro do you think we should go ahead and do this as part of that release?

tests/jwst/datamodels/test_read_metadata.py
emolter, line 1: It might be good to test the isinstance behavior, with a comment like "covers a bug where missing WCS tag led to inconsistent isinstance behavior" or something

emolter: I meant the hasattr behavior

PR #668, opened 2026-02-20
New Miri wfss photom datamodel

jemorrison:
Create a MIRI WFSS photom reference datamodel. We need this datamodel for JP-4253 and pr https://github.com/spacetelescope/jwst/pull/10280

Partially resolves [JP-4253](https://jira.stsci.edu/browse/JP-4253)
https://jira.stsci.edu/browse/CRDS-1083

Closes #

This PR addresses  adds a MIRI photom datamodel.

## Tasks
  - [x] [run `jwst` regression tests](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) with this branch installed (`"git+https://github.com/jemorrison/stdatamodels@MIRI_WFSS_PHOTOM"`)
The regression test passed:
https://github.com/spacetelescope/RegressionTests/actions/runs/22634145423

</details>


braingram (commented):
I left a few suggestions in the schema to consolidate some of the subschemas added in this PR.

Is there an example file on CRDS?

I only see `phot_table` being used from the file. Are the other tables used or is there a plan to use them (`timecoeff_linear`, `timecoeff_exponential`, `timecoeff_powerlaw`)?

braingram:
Thanks. That file fails to load with the new model:
```python
m = dm.MirWfssPhotomModel("jwst_miri_photom_0231.fits")
```
produces an exception:
```
ValueError: Array has wrong number of dimensions. Expected 1, got 0
```
Should the schema have a "subarray" column or does the reference file have an incorrect "subarray" column?

I don't see any of the time tables (only "PHOTOM"). @tapastro what's your advice here? Add the non-existent tables to the schema or wait until they're needed? FWIW having extra data in the FITS files is not an issue so the tables could be added at any point.

melanieclarke:
> So should those also have `subarray`?

Yes, if it's added to the phot_table it needs to be added to the time tables as well.  They need to have matching descriptive columns and entries in the same order.  The validation for photom models will raise an error in the photom step if the time table is present but doesn't match the phot_table.

jemorrison:
Well that is strange. I use that file in the JWST Pipeline and it works fine. 
So the JWST Pipeline is loading it differently ?

What is even stranger I made the file using this datamodel. I even validated it. 
So very strange.

melanieclarke (commented):
As Brett said, the time tables need subarray to match the phot_table.  Otherwise I think the schema looks okay.

melanieclarke (approved):
Everything looks good to me now.  I will merge after the downstream tests complete.

PR #670, opened 2026-02-23
JP-4228: Disable default array creation on access

### LINKED ISSUE #310: Investigate using a singleton in place of `None` to trigger `_make_default`
Currently setting a `DataModel` attribute to `None` (which is a common pattern for freeing up memory) results in creation of a default value for the attribute:
https://github.com/spacetelescope/stdatamodels/blob/ee4756ffe9ff71c1fb7b067f8998949a88d6a703/src/stdatamodels/properties.py#L324-L330

For large arrays (which are likely candidates for removal to free up memory) this results in creation of a large new array with default values.

```python
>>> import stdatamodels.jwst.datamodels as dm
>>> m = dm.ImageModel((1000, 1000))
>>> d0 = m._instance['data']  # avoid the getattr for data
>>> d0.shape
(1000, 1000)
>>> m.data = None
>>> d0 is m._instance['data'] 
False
>>> m._instance['data'].shape
(1000, 1000)
```

It may be helpful to use a singleton like `MAKE_DEFAULT` to signify that a default should be made.

```python
>>> import stdatamodels
>>> import stdatamodels.jwst.datamodels as dm
>>> m = dm.ImageModel((1000, 1000))
>>> m.data = stdatamodels.MAKE_DEFAULT  # to trigger a new default for data
```

This could help to make it more explicit when a default is requested. However it would require changes to code like the following:
https://github.com/spacetelescope/stdatamodels/blob/ee4756ffe9ff71c1fb7b067f8998949a88d6a703/src/stdatamodels/jwst/datamodels/slit.py#L65-L68

### LINKED ISSUE #436: array attribute defaults: always, not always, checking, how to instantiate
spacetelescope/jwst#1075 brought up a couple of issues with datamodels and the handling of the data arrays.  Currently, if an array is specified in the schema, the array will exist upon access, regardless of whether it was part of the initializing data.  From spacetelescope/jwst#1075, some issues with this approach have arisen:

1. Allow "optional" arrays. These arrays would not automatically instantiate when accessed. If an array was not in the initializing data, or explicitly set, an `AttributeError` should be thrown. This behavior could key off the existence of attribute's `default` property, or create another property.
2. Ability to check if an array exists without instantiating. This can currently be done using the `DataModel.instance`, but that is rather hacky/api-breaking.
3. Specify how an array is instantiated on a case-by-case basis. If an attribute is specified with `ndim` less than the primary data, only the _fastest moving_ dimensions are used in instantiating the array. However, as an example, `Level1bModel.zeroframe` does not follow this model, instead needing an arbitrary subset of the main data's dimensions. Note that this may be the only case and not worth consideration. It is mentioned here in case there are others, and that the current solution may break if the first two issues are dealt with.

Issue note: This is +1 on the esoteric scale. Just documenting here for later consideration.

### LINKED ISSUE #470: Default creation on access leads to unexpected `hasattr` behavior
Automatically creating attributes when accessed leads to "hasattr" always returning `True` and creating missing attributes (if they have a default).
```
In [1]: import stdatamodels.jwst.datamodels as dm
m
In [2]: m = dm.ImageModel()

In [3]: "data" in m
Out[3]: False

In [4]: hasattr(m, "data")
Out[4]: True

In [5]: "data" in m
Out[5]: True
```
This was previously known and `Model.hasattr` was added as a work around: https://github.com/spacetelescope/jwst/pull/2275

However `Model.hasattr` has only a few uses in jwst:
https://github.com/search?q=repo%3Aspacetelescope%2Fjwst%20%22.hasattr%22&type=code
whereas there are `hasattr` uses in jwst that are buggy because of this behavior.

For one example `match_nans_and_flags` creates missing extensions:
https://github.com/spacetelescope/jwst/blob/5fe2f38eb9bf36c2ea18f9736a5a51cffd8ff2e3/jwst/lib/pipe_utils.py#L115-L119

emolter:
Partially addresses [JP-4228](https://jira.stsci.edu/browse/JP-4228)

Closes #310 
Closes #436
Closes #470 

Same as https://github.com/spacetelescope/stdatamodels/pull/662 but reverting the change to metadata handling.  It made sense to keep these separate because the metadata changes created hundreds of new regression test differences, making it hard to see the differences related to the array changes.

Plan would be to merge this one and the associated JWST changes (see https://github.com/spacetelescope/jwst/pull/10267)
then add back the change to metadata handling, then merge a second PR and a much smaller second round of associated JWST changes.

This PR

- Allows setting of array-like attributes of models to `None`.  Previously, attempting this for an array with `ndim` in its schema would set the attribute to an ndarray of size 0 and that many dimensions.
- Makes `getattr` return `None` for un-initialized array-like attributes
- Fixes the bug where `getattr` and `hasattr` modified array-like attributes.  Attempting to access an undefined array-like attribute no longer sets that attribute to its schema-defined default.
- Preserves existing behavior that `hasattr` returns `True` for anything schema-defined, including data or metadata
- Adds `DefaultErrMixin` and `DQMixin` to replace `model.dq = model.dq` and `model.err = model.err` calls in class inits
- Adds `get_default` method to `ObjectNode` return the schema-defined default of an attribute.  This provides a way to explicitly generate a default array and replace e.g. `model.var_rnoise*=0`, `model.var_rnoise = model.var_rnoise`, or any of the other myriad ways that previously accessed an array trivially in order to set it to its default.
- Adds `get_dtype` method to `ObjectNode` to provide a convenient replacement for e.g. `model.spec_table = np.array(whatever, dtype=model.spec_table.dtype)` which is a common usage of the old default array setting behavior. This would now look like `model.spec_table = np.array(whatever, dtype=model.get_dtype("spec_table"))`

Example of the new hasattr and getattr behavior:
```
model = dm.ImageModel()
hasattr(model, "data")  # True
model.data is None  # True
"data" in model._instance  # False
hasattr(model.err)  # True
model.err is None  # True
"err" in model._instance  # False
model.info()  # does not contain data or err arrays at all
```

Example of new behavior when explicitly setting to None:
```
model = dm.ImageModel(strict_validation=True)
model.data = None
model.data is None  # True
model._instance["data"] is None  # True
model.validate()  # no errors
model.save("test.fits")

model2 = dm.open("test.fits")
data in model2._instance  # False
model2.data is None  # True
```
So the `None` in the instance is not preserved on save/load, but I think this is fine since `getattr` still returns None.

The changes here still allow `model.err = model.err` to be a valid construction in `__init__`, but it will no longer initialize to default; `model.err = model.err` is now equivalent to `model.err = None`.

The only side-effect that caused a failing stdatamodels unit test is that setting a tree that has a subtree to None makes it no longer accessible, i.e.
```
model = ImageModel()
hasattr(model.meta.observation, "date") # True
model.meta.observation = None
hasattr(model.meta.observation, "date")  # False
```
If this is not desired and we prefer it the old way, we can discuss options for that.

## Tasks

</details>

### DIFF
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -19,7 +19,8 @@ In this form, the memory for the primary array (typically "data") is
 allocated on init, but additional schema-defined arrays are not.
 This is useful if, for example, you don't
 need a data quality array -- the memory for such an array will not be
-consumed::
+consumed. We can use the :meth:`~stdatamodels.DataModel.info` method to
+see everything that's currently in the model::
 
   im.info()
   
@@ -33,11 +34,30 @@ consumed::
     ├─model_type (str): ImageModel
     └─date (str): 2026-02-26T14:56:12.278
 
+Notice that only the primary array ("data") is allocated, has the shape we put in,
+and has a data type of float32 as defined by the `ImageModel` schema.
 To set additional arrays to their default values, use the
 ``get_default`` method::
 
-    im.dq = im.get_default("dq")
-    print(im.dq) # shape (1024, 1024), zero-filled
+  im.dq = im.get_default("dq")
+  im.info()
+
+  root (AsdfObject)
+  ├─data (ndarray)
+  │ ├─shape (tuple)
+  │ │ ├─[0] (int): 1024
+  │ │ └─[1] (int): 1024
+  │ └─dtype (Float32DType): float32
+  ├─meta (dict)
+  │ ├─model_type (str): ImageModel
+  │ └─date (str): 2026-02-26T14:56:12.278
+  └─dq (ndarray)
+    ├─shape (tuple)
+    │ ├─[0] (int): 1024
+    │ └─[1] (int): 1024
+    └─dtype (UInt32DType): uint32
+
+As you can see, the ``dq`` array is now allocated with the same shape as the primary array.
 
 .. note::
 
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -7,8 +7,8 @@ Creating a data model from scratch
 ----------------------------------
 
 To create a new :class:`~stdatamodels.jwst.datamodels.ImageModel`,
-just call its constructor.  To create a
-new model where all of the arrays will have default values, simply
+just call its constructor.  For example, to create a
+new model with the primary array set to its default, simply
 provide a shape as the first argument::
 
     from stdatamodels.jwst.datamodels import ImageModel
@@ -23,7 +23,8 @@ consumed::
 
   # Print out the data array.  It defaults to being filled with zeros.
   print(im.data) # shape (1024,1024)
-  "dq" in im.instance # False
+  # The dq array is not allocated by default
+  print(im.dq) # None
 
 To set additional arrays to their default values, use the
 ``get_default`` method::
@@ -37,10 +38,8 @@ To set additional arrays to their default values, use the
   attribute and setting it to its default value could be achieved using e.g.
   ``im.dq = im.dq``. This syntax took advantage of a bug wherein data
   arrays were created on access, leading to unexpected behaviors. This construction
-  no longer works; now, if an attribute is schema-defined but unset,
-  accessing it returns ``None``, so ``im.dq = im.dq`` will set ``dq`` to ``None``.
-  Instead use ``get_default`` as shown above
-  to set the attribute to its default value.
+  no longer sets ``dq`` to a default array.
+  Use ``get_default`` as shown above instead.
 
 If you already have data in numpy arrays, you can also create a model
 using those arrays by passing it in as a data keyword argument::
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -15,17 +15,35 @@ provide a shape as the first argument::
     with ImageModel((1024, 1024)) as im:
         ...
 
-In this form, the memory for the arrays will not be allocated until
-the arrays are accessed.  This is useful if, for example, you don't
+In this form, the memory for the primary array (typically "data") is 
+allocated on init, but additional schema-defined arrays are not.
+This is useful if, for example, you don't
 need a data quality array -- the memory for such an array will not be
 consumed::
 
-  # Print out the data array.  It is allocated here on first access
-  # and defaults to being filled with zeros.
-  print(im.data)
+  # Print out the data array.  It defaults to being filled with zeros.
+  print(im.data) # shape (1024,1024)
+  "dq" in im.instance # False
 
-If you already have data in a numpy array, you can also create a model
-using that array by passing it in as a data keyword argument::
+To set additional arrays to their default values, use the
+``get_default`` method::
+
+    im.dq = im.get_default("dq")
+    print(im.dq) # shape (1024, 1024), zero-filled
+
+.. note::
+
+  In previous versions of ``stdatamodels``, allocating memory for an
+  attribute and setting it to its default value could be achieved using e.g.
+  ``im.dq = im.dq``. This syntax took advantage of a bug wherein data
+  arrays were created on access, leading to unexpected behaviors. This construction
+  no longer works; now, if an attribute is schema-defined but unset,
+  accessing it returns ``None``, so ``im.dq = im.dq`` will set ``dq`` to ``None``.
+  Instead use ``get_default`` as shown above
+  to set the attribute to its default value.
+
+If you already have data in numpy arrays, you can also create a model
+using those arrays by passing it in as a data keyword argument::
 
     data = np.empty((50, 50))
     dq = np.empty((50, 50))
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -21,10 +21,17 @@ This is useful if, for example, you don't
 need a data quality array -- the memory for such an array will not be
 consumed::
 
-  # Print out the data array.  It defaults to being filled with zeros.
-  print(im.data) # shape (1024,1024)
-  # The dq array is not allocated by default
-  print(im.dq) # None
+  im.info()
+  
+  root (AsdfObject)
+  ├─data (ndarray)
+  │ ├─shape (tuple)
+  │ │ ├─[0] (int): 1024
+  │ │ └─[1] (int): 1024
+  │ └─dtype (Float32DType): float32
+  └─meta (dict)
+    ├─model_type (str): ImageModel
+    └─date (str): 2026-02-26T14:56:12.278
 
 To set additional arrays to their default values, use the
 ``get_default`` method::
--- a/docs/source/jwst/datamodels/new_model.rst
+++ b/docs/source/jwst/datamodels/new_model.rst
@@ -299,7 +299,8 @@ reference value either to the provided array or to its default.
 We also want to handle ``dq`` and ``dq_def``. These are common enough that
 ``stdatamodels`` provides a convenient way to map the ``dq`` array to the standard
 values using its ``dq_def`` table on initialization. This is accomplished by letting
-the model inherit from the `~stdatamodels.jwst.datamodels.model_base.DefaultDQMixin` class::
+the model inherit from the
+:class:`~stdatamodels.jwst.datamodels.model_base.DefaultDQMixin` class::
 
   from stdatamodels.jwst.datamodels import JwstDataModel, DefaultDQMixin
 
--- a/docs/source/jwst/datamodels/new_model.rst
+++ b/docs/source/jwst/datamodels/new_model.rst
@@ -206,7 +206,7 @@ defined above:
           An array that is assumed to exist for all BadpixModel instances,
           and is set to default on init.
       """
-      schema_url = "bad_pixel_mask.schema.yaml"
+      schema_url = "badpix.schema.yaml"
 
 This is already a fully-defined model class and we could use it as-is.
 For example, we could do::
--- a/docs/source/jwst/datamodels/new_model.rst
+++ b/docs/source/jwst/datamodels/new_model.rst
@@ -4,30 +4,14 @@
 Defining a new model type
 =========================
 
-This tutorial describes the steps necessary to define a new model type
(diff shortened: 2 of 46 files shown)


src/stdatamodels/model_base.py
```diff
@@ -975,6 +974,8 @@ def hdu_names_from_schema(subschema, path, combiner, ctx, recurse):
 
         def included(cursor, part):
             # Test if part is in the cursor
+            if cursor is None:
+                return False
```
braingram: To help me understand the changes to `update` would you comment on the need for this change?

src/stdatamodels/properties.py
```diff
@@ -402,8 +412,6 @@ def __setattr__(self, attr, val):
             self.__dict__[attr] = val
         else:
             schema = _get_schema_for_property(self._schema, attr)
-            if val is None:
-                val = _make_default(attr, schema, self._ctx)
```
braingram: Is this (perhaps in part) what was preventing assigning `None` to things like `meta.observation` on main? I'm not suggesting we keep this (confusing) behavior.

braingram (commented):
I did not review the test changes (but will do so later). Thanks for making these improvements. The comments are mostly:
- to check I'm following the changes
- questions about the mixin usage

Overall things LGTM but I didn't approve since I didn't review the tests or downstream changes yet.

As far as informing users I think this warrants a section in the docs describing:
- how to migrate code from the old (`self.err = self.err`) to the new (`self.err = self.get_default("err")`)
- with an eye towards (if not some mention) follow-up changes (to metadata initialization, etc)

We could consider making the next release "5.0.0" to help downstream be aware that there are breaking changes (not that these are necessarily "breaking"). That way anyone pinning `<5` will hopefully get a dependency upgrade notice and look at the changelog.

src/stdatamodels/model_base.py
emolter: Yeah it was a confusing edge case.  Basically if you were trying to update model B from model A (give attributes from A to B) but A didn't have an entire tree (e.g., A doesn't have anything in `model.meta.observation`) then `cursor` would be `None` and the lines below that check for things in `cursor` would give "NoneType is not iterable".  I meant to write a unit test for this and forgot about it.  Will do so, hopefully tomorrow.

src/stdatamodels/jwst/datamodels/image.py
```diff
@@ -1,9 +1,9 @@
-from .model_base import JwstDataModel
+from .model_base import JwstDataModel, _DefaultDQMixin, _DefaultErrMixin
 
 __all__ = ["ImageModel"]
 
 
-class ImageModel(JwstDataModel):
+class ImageModel(JwstDataModel, _DefaultDQMixin, _DefaultErrMixin):
```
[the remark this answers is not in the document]
emolter: This is a good question.  The reason was primarily because not doing so led to many more unit and regression test failures.  ImageModel is by far the most used datamodel for testing, and there were so many places where we implicitly assume the model has a dq and err array that it felt easier to just give it one explicitly on init.

In regression test data, only a few intermediate files with extensions `_mask`, `_simul`, and `_profile` get extra (empty) arrays due to this change.  This is not ideal, but I'd say neither is it catastrophic.

The plan was to merge the current PR, then try slowly removing most or all default array setting on init and see what happens.
However that being said, now that this is in a state I'm fairly happy with, I could revert this particular change and see how many more differences we'd really need to account for here.

src/stdatamodels/jwst/datamodels/model_base.py
```diff
@@ -112,3 +115,39 @@ def update(self, d, only=None, extra_fits=False, cal_logs=True):
 
         # Call the parent update
         super().update(d, only=only, extra_fits=extra_fits)
+
+
+class _DefaultDQMixin(_DataModel):
```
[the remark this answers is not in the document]
emolter: good point. We do have such a page: https://stdatamodels.readthedocs.io/en/latest/jwst/datamodels/new_model.html#the-model-class

Skimming that, it does appear there's some outdated info there, and an example that looks like it's easily convertible to using the mixins.

braingram (commented):
Here are the next set of comments from reviewing the tests.

src/stdatamodels/jwst/datamodels/slit.py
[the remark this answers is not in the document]
emolter: Same reason as `ImageModel`.  I will also check the effect of undoing this change and see how many failures this causes.

tests/test_models.py
```diff
@@ -299,7 +429,7 @@ def test_skip_serializing_null(tmp_path, filename):
 
     with BasicModel(file_path) as model:
         # Make sure that 'telescope' is not in the tree
-        assert "telescope" not in model.meta._instance
+        assert "telescope" not in model.meta
```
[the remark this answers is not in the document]
braingram: Agreed. Thanks for asking that question :) For the scope of this PR I'd say let's just try to not introduce more uses of `_instance` if that matches with your understanding @emolter

tests/test_models.py
```diff
@@ -146,9 +154,130 @@ def test_base_model_has_no_arrays():
...
+
+@pytest.mark.xfail(reason="fails when reverting default meta behavior")
+def test_implicit_meta_none():
+    """
+    Test access to undefined metadata attributes.
+
+    Ensure returns None, even if there is a schema-defined default.
+    Ensure the attribute is not set during getattr."""
+    with DefaultsModel() as im:
+        assert hasattr(im.meta, "no_default_meta")
```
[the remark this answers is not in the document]
emolter: In https://github.com/spacetelescope/stdatamodels/pull/670/changes/e45c6340531e40292c616e2e4ae28a0009240f17 this should be removed.  I also added a test specific to the hasattr vs in inconsistency

src/stdatamodels/jwst/datamodels/image.py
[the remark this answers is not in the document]
emolter: https://github.com/spacetelescope/stdatamodels/pull/670/changes/c2cb5a681344c831ac1c667a5f23fcb450ad2218 removes the mixins from both `SlitModel` and `ImageModel`

docs/source/jwst/datamodels/models.rst
```diff
@@ -15,17 +15,35 @@ provide a shape as the first argument::
...
+This is useful if, for example, you don't
 need a data quality array -- the memory for such an array will not be
 consumed::
 
-  # Print out the data array.  It is allocated here on first access
-  # and defaults to being filled with zeros.
-  print(im.data)
+  # Print out the data array.  It defaults to being filled with zeros.
+  print(im.data) # shape (1024,1024)
+  "dq" in im.instance # False
```
melanieclarke: I found this example a little confusing on first read.  I kept seeing "dq" as "data" and wondering why it wasn't in the instance after it had just been printed.

In general, I think `in im.instance` is kind of advanced usage, so maybe instead this example could be:

```suggestion
  # Print out the data array.  It defaults to being filled with zeros.
  print(im.data) # shape (1024,1024), zero-filled
  # The dq array is not allocated by default
  print(im.dq) # None
```

braingram (commented):
A couple comments on the docs updates. Thanks for making these improvements!

braingram:
> It also occurs to me that if we're documenting `_DefaultDQMixin` maybe it should not be private?

If it's easy to make public as part of this PR that seems reasonable.

> Also, in the current state `get_default` and `get_dtype` don't render in the docs due to `ObjectNode` not being documented and living in the `nitpick_ignore` of `docs/conf.py`. This also seems less than ideal. Should we go ahead and document `ObjectNode`?

That's unfortunate. Maybe a follow-up PR for this one since I always doubt sphinx will do anything reasonable

docs/source/jwst/datamodels/new_model.rst
```diff
@@ -58,21 +40,24 @@ this minimal example.
...
-Let's start with the schema file, ``bad_pixel_mask.schema.yaml``.
+Let's start with the schema file, ``badpix_model.schema.yaml``.
 There are a few things it needs to do:
 
    1) It should contain all of the core metadata from the core schema
-      that ships with `jwst.datamodels`.  In JSON Schema parlance, this
+      that ships with ``stdatamodels``.  In JSON Schema parlance, this
       schema "extends" the core schema.  In object-oriented
       programming terminology, this could be said that our schema
       "inherits from" the core schema.  It's all the same thing.
```
melanieclarke: Unrelated to your changes here, but object-oriented inheritance is quite different from schema extension, so this sentence might be misleading.  I'd just leave out these last two sentences.

```suggestion
schema "extends" the core schema. 
```

docs/source/jwst/datamodels/new_model.rst
```diff
@@ -223,65 +246,89 @@ elements, by setting `schema_url` to::
...
+initializing it to the default if not provided. This pattern can be helpful
+if calling code assumes an attribute is always set and it's ok for the default
+value to be passed:
 
 .. code-block:: python
 
-    def __init__(self, init=None, dq=None, dq_def=None, **kwargs):
+    def __init__(self, init=None, reference_data=None, **kwargs):
         """
         A data model to represent MIRI bad pixel masks.
```
melanieclarke:
```suggestion
        A data model to represent bad pixel masks.
```

docs/source/jwst/datamodels/new_model.rst
melanieclarke: This won't return Boolean values, since `&` is bitwise.  It should be
 
```suggestion
        return (self.dq & field_value) > 0
```

melanieclarke (commented):
Code all looks good to me, although I agree it would be better to make DefaultDQMixin and DefaultErrMixin public.

A few comments on the new docs.  Thank you for updating the example!  It definitely needed it.

docs/source/jwst/datamodels/new_model.rst
emolter: nice catch, tbh I didn't even look at the code in the preexisting example

melanieclarke (commented):
Docs for the mixins looks fine to me.  

A couple more clean up comments.

melanieclarke (commented):
All my comments have been addressed, and I'm happy with both this and the corresponding jwst PR.

I'll hold off on approval to give Tyler a chance to review.  Also, I'd like to make sure we run regression tests again after all changes are in, before we merge.

tapastro (approved):
LGTM, thanks all for the hard work! 🐌

Issue #581, opened 2025-09-04
Combine `DataModel` and `JwstDataModel`


braingram:
[JwstDataModel](https://github.com/spacetelescope/stdatamodels/blob/3a77857ba4495f299f79046d5203bc98ce6a4922/src/stdatamodels/jwst/datamodels/model_base.py#L10) is the only class that inherits (directly) from [DataModel](https://github.com/spacetelescope/stdatamodels/blob/3a77857ba4495f299f79046d5203bc98ce6a4922/src/stdatamodels/model_base.py#L44). As `DataModel` is only intended for use by jwst investigate combining these.

PR #675, opened 2026-02-26
Docs updates for model base classes

### LINKED ISSUE #672: Docs do not report ObjectNode
See title.  Although direct instantiation of `ObjectNode` is discouraged in most cases, there's no harm in giving interested users access to what's documented there, especially since doing something as simple as
```
model = dm.ImageModel()
model.meta
```
gives `<stdatamodels.properties.ObjectNode object at memoryaddress>`

This will also enable us to expose the soon-to-be-merged `get_default` and `get_dtype` methods which are meant for public use but defined at the `ObjectNode` level.

emolter:
Closes #672

This PR:

- Exposes the `ObjectNode`  and `ListNode` classes in the docs (#672)
- Makes `JwstDataModel` show inherited members.  This change was made to ensure that once https://github.com/spacetelescope/stdatamodels/pull/670 is merged, the new `get_dtype` and `get_default` methods would be visible from the `JwstDataModel`  docs page instead of forcing users to search through `DataModel` to `ObjectNode` to find them.  Prior to merging that PR, we can see the same effect for `model.hasattr`.
- Moves `JwstDataModel` to its own doc page, which is linked from the top of the API page

New API page can be seen at https://stdatamodels--675.org.readthedocs.build/en/675/api.html 

## Tasks

</details>

### DIFF
--- a/docs/source/_templates/autosummary/class.rst
+++ b/docs/source/_templates/autosummary/class.rst
@@ -1,0 +1,8 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+   :members:
+   :inherited-members:
+   :show-inheritance:
--- a/docs/source/_templates/autosummary/class.rst
+++ b/docs/source/_templates/autosummary/class.rst
@@ -1,8 +1,0 @@
-{{ fullname | escape | underline}}
-
-.. currentmodule:: {{ module }}
-
-.. autoclass:: {{ objname }}
-   :members:
-   :inherited-members:
-   :show-inheritance:
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -9,7 +9,11 @@ stdatamodels API
 Model base class
 ----------------
 
-:class:`~stdatamodels.jwst.datamodels.JwstDataModel`
+.. list-table::
+   :widths: auto
+
+   * - :class:`~stdatamodels.jwst.datamodels.JwstDataModel`
+     - Base class for all JWST data models.
 
 .. automodapi:: stdatamodels
    :no-inheritance-diagram:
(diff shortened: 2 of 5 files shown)


docs/source/api.rst
```diff
@@ -1,8 +1,18 @@
...
-   :no-inheritance-diagram:
+Model base classes
+------------------
+
+.. autosummary::
+   :toctree: generated/
+   :nosignatures:
+
+   stdatamodels.properties.ObjectNode
+   stdatamodels.properties.ListNode
```
braingram:
```suggestion
.. automodapi:: stdatamodels.properties
   :no-inheritance-diagram:
```
Did automodapi for `properties` not work? That would be more consistent with the other submodules (which use automodapi).

emolter: I think automodapi for `properties` should work fine, but I thought it'd be nice to have `ObjectNode` in the same section as `DataModel` and `JwstDataModel`.  Arguably though just ensuring that `DataModel` and `JwstDataModel` show their inherited methods from `ObjectNode` might be good enough.

docs/source/_templates/autosummary/class.rst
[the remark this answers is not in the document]
emolter: Yes this would change again if we do something with #581.  Note though that although I am experimenting with #581 the changes in [this JWST PR](https://github.com/spacetelescope/jwst/pull/10288) (which I think is the one you're referencing) are good changes to make either way.

Thanks for the input.  I think I have a plan for how to make this less custom but still satisfy the requirements that:

1. `JwstDataModel` gets `:inherited-members:`
2. None of the other classes at `stdatamodels.jwst.datamodels` get `:inherited-members:`

which IMO are important requirements for this PR to do what I think it should do.
I'll push something up shortly.

braingram (approved):

Issue #682, opened 2026-03-09
Validate array shape on assignment


emolter:
As noted in #505 the following works without issue:
```a = datamodels.ImageModel(data=np.ones(shape=(1024, 1024)), dq=np.ones(shape=(512, 512)))```

as does
```
a = datamodels.ImageModel(data=np.ones(shape=(1024, 1024)))
a.dq = np.ones(shape=(512, 512))
```

It has always been an implicit requirement for the majority of extensions of a datamodel to have the same shape as the primary.  This issue is a suggestion to enforce that via some sort of validation on assignment.  This would take place alongside schema validation on assignment, and check against `model._shape`.

There are complications with certain arrays, e.g. the context array or certain "average" arrays, where the shape is expected to differ, and we still need to come up with a good way to handle those cases.

PR #683, opened 2026-03-10
JP-4097: Support for SOSS multistripe subarrays

tapastro:
Partially resolves [JP-4097](https://jira.stsci.edu/browse/JP-4097)

This PR addresses updates necessary to perform calibrations on multistripe data, particularly for the first multistripe implementation using NIRISS SOSS. Subarray names are added, and a new INT_TIMES table is defined to preserve per-stripe integration times after stripe collation into a parent frame (SUBSTRIP256 in these cases, perhaps FULL in future ones). 

Additionally, the dimension of pixeldq in RampModel was removed. This change has the potential for greatest general impact; it seemed like the path of least resistance to accomplish superstripe calibrations. Because we'll receive integrations per stripe, pixel (0, 0) will not correspond to the same location on detector across stripes. Because of this, a pixeldq array needs to have representation of each of stripe plane, pushing it into three dimensions. An alternative approach could be to store the pixeldq array explicitly in the detector frame, but then to apply a stripe transformation whenever accessing it. This seemed too complicated only to preserve the dimensionality constraint, but seems possible if the current approach encounters obstacles.

## Tasks

</details>


emolter (commented):
Overall the changes look more-or-less minimally invasive, so I think this is fine; however, it'll be helpful to see the JWST PR to understand usages better and decide if defining a new model type for this mode is warranted.

Just because I made it recently, I'm thinking about how this would interact with https://github.com/spacetelescope/stdatamodels/issues/682.  But we already have cases where we need to be able to handle arrays of different shapes/dimensions, so adding one more makes no difference.  Making a new model type _might_ make it easier to validate the shape of `pixeldq` for the purposes of that issue, but I'm not even sure it necessarily would, so best not to worry about it.

emolter (approved):
All the band-aids don't make me particularly happy but this appears at least more-or-less work for now and there's no time left in the build

PR #706, opened 2026-04-06
Fix failure to update model type

### LINKED ISSUE #705: Initializing a model from another model does not update meta.model_type
This looks like a bug:
```
model = dm.ImageModel()
model2 = dm.SlitModel(model)
model2.meta.model_type # 'ImageModel'
```

emolter:
Closes #705 

This PR fixes a bug where initializing a model from a different (but compatible) model type did not update ``model.meta.model_type``

Note that this also relates to https://github.com/spacetelescope/stdatamodels/issues/698 as it should make it so e.g.
```
model = dm.ImageModel()
model2 = dm.IFUImageModel(model)
model == model2
```
will be False instead of True.

## Tasks

</details>

### DIFF
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -322,8 +322,9 @@ It's also possible to convert between compatible model types this way, e.g.::
     old_model = ImageModel()
     new_model = IFUImageModel(old_model)
 
-and in this case, a deep copy of the metadata will be performed but the data
-arrays will be shallow-copied for memory efficiency.
+and in this case, ``new_model is old_model`` will be ``False``. Simple metadata will be copied and
+the ``meta.model_type`` attribute will be updated, but data arrays and other complex types (e.g.,
+the WCS object) will be shared between the two models.
 
 
 History information
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -306,6 +306,25 @@ model into a new one using the :meth:`~stdatamodels.DataModel.update` method::
 
     new_model.update(old_model)
 
+While it's typically not recommended for general use due to the possibility of
+unexpected behavior, copying a model can be achieved by passing it into
+a constructor of the same type::
+
+    old_model = ImageModel()
+    new_model = ImageModel(old_model)
+
+This simply creates a new reference to the model object, i.e. ``new_model is old_model`` will be ``True``.
+Changes to one model will affect the other.
+
+It's also possible to convert between compatible model types this way, e.g.::
+
+    old_model = ImageModel()
+    new_model = IFUImageModel(old_model)
+
+and in this case, a deep copy of the metadata will be performed but the data
+arrays will be shallow-copied for memory efficiency.
+
+
 History information
 -------------------
 
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -35,7 +35,8 @@ see everything that's currently in the model::
     └─date (str): 2026-02-26T14:56:12.278
 
 Notice that only the primary array ("data") is allocated, has the shape we put in,
-and has a data type of float32 as defined by the `ImageModel` schema.
+and has a data type of float32 as defined by the
+:class:`~stdatamodels.jwst.datamodels.ImageModel` schema.
 To set additional arrays to their default values, use the
 ``get_default`` method::
 
--- a/docs/source/jwst/datamodels/models.rst
+++ b/docs/source/jwst/datamodels/models.rst
@@ -314,17 +314,18 @@ a constructor of the same type::
     old_model = ImageModel()
     new_model = ImageModel(old_model)
 
-This simply creates a new reference to the model object, i.e. ``new_model is old_model`` will be ``True``.
-Changes to one model will affect the other.
+This will perform a shallow copy: in this case ``old_model.instance == new_model.instance``,
+and changes to one model will affect the other.
 
 It's also possible to convert between compatible model types this way, e.g.::
 
     old_model = ImageModel()
     new_model = IFUImageModel(old_model)
 
-and in this case, ``new_model is old_model`` will be ``False``. Simple metadata will be copied and
-the ``meta.model_type`` attribute will be updated, but data arrays and other complex types (e.g.,
-the WCS object) will be shared between the two models.
+This will also perform a shallow copy, except
+that some metadata elements (such as ``meta.model_type``) will be updated and therefore
+``old_model.instance != new_model.instance``. Almost all attributes will still be shared
+between models, so changes to one will affect the other.
 
 
 History information
--- a/changes/706.bugfix.rst
+++ b/changes/706.bugfix.rst
@@ -1,1 +1,1 @@
-Fix a bug where initializing a model from a different (but compatible) model type did not update model.meta.model_type
+Fix a bug where initializing a model from a different (but compatible) model type did not update ``model.meta.model_type``
--- a/changes/706.bugfix.rst
+++ b/changes/706.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a bug where initializing a model from a different (but compatible) model type did not update model.meta.model_type
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -221,7 +221,7 @@ class DataModel(properties.ObjectNode):
                 # but share the same data arrays.
                 self._instance = copy.copy(self._instance)
                 if "meta" in self._instance:
-                    self._instance["meta"] = copy.deepcopy(self._instance["meta"])
+                    self._instance["meta"] = copy.copy(self._instance["meta"])
 
                 current_validate_arrays = self._validate_arrays
                 self._validate_arrays = True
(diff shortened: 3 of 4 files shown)


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/24042063573

braingram (changes requested):
Thanks for working on this.

Do we have expectations/assumptions about what `DataModel(some_model)` should do?

Testing on main:
```python
>> import stdatamodels.jwst.datamodels as dm, numpy as np
>> m = dm.ImageModel((5, 5))
>> m2 = dm.IFUImageModel(m)
>> m._instance is m2._instance
True
```

So the original ImageModel `m` and new IFUImageModel `m2` share the backing `_instance`. On main this can explain why `m2.meta.model_type` is `ImageModel`.

With this PR the `m._instance is m2._instance` is still true but this means that now instead of `m2` having `ImageModel` the original model `m` now has the incorrect model type `IFUImageModel`.

```python
>> import stdatamodels.jwst.datamodels as dm, numpy as np
>> m = dm.ImageModel((5, 5))
>> m2 = dm.IFUImageModel(m)
>> m.meta.model_type
'IFUImageModel'
```

I'm marking this as requested changes since I think the behavior in this PR is more confusing than on main.

Is `SomeModel(instance_of_different_model)` used anywhere?

emolter:
Good catch, thanks.  I should have added those tests.  I'll need to re-think how to do this.  I believe this is indeed used but can look into details of where and how.

emolter:
@braingram To answer your question about usages , at minimum this is used every time `Step.prepare_output` is called using the `open_as_type` option, which is [quite a lot](https://github.com/search?q=repo%3Aspacetelescope%2Fjwst+open_as_type&type=code).

This also somewhat informs what to do about it.  For example, I would hesitate to make it so this case triggered a full deepcopy because specifically `Level1bModel` --> `RampModel` is a common occurrence and those models can be large.

I think the best solution would be to somehow make the model instances different but have them share the same data arrays. e.g. the unit test would satisfy:
```
    input_model = FitsModel((50, 50))
    with BasicModel(input_model) as dm:
        assert dm.data.shape == (50, 50)
        assert dm.meta.model_type == "BasicModel"
        assert input_model._instance is not dm._instance
        assert input_model.data is dm.data
    assert input_model.meta.model_type == "FitsModel"
    input_model.close()
```

What do you think?

emolter:
https://github.com/spacetelescope/stdatamodels/pull/706/commits/9318a157e26db402c55b9a80172268bb68756716 achieves those requirements, not sure of the other impacts if any though.

braingram:
Thanks for making the update. This looks reasonable to me. To track I'm following the changes, with this PR the `meta` and everything under it (including `wcs`) was identical on main and now will be a deepcopy?

emolter:
yes that's correct

emolter:
@tapastro @melanieclarke what do you think of the proposed behavior here?

melanieclarke:
A deepcopy on the wcs can be an expensive operation.  If the data arrays are shallow copies, I wouldn't necessarily expect that the wcs would be independent.  If we did a shallow copy on the meta, so that the top level dictionary is independent but complex types within it are not, would that also serve the purpose?

emolter:
@melanieclarke are you suggesting something like https://github.com/spacetelescope/stdatamodels/pull/706/commits/ff2cf4486fd15bea94a01c203ece73ac902954cb ?  

From what I can tell that also seems to achieve the desired results, which I'm taking here to basically be the unit test assert statements: models are not equal, their model types are different, their data arrays are references to each other.  And now it would seem their WCSs are also just references.

melanieclarke:
> @melanieclarke are you suggesting something like [ff2cf44](https://github.com/spacetelescope/stdatamodels/commit/ff2cf4486fd15bea94a01c203ece73ac902954cb) ?

Yes, exactly!

> From what I can tell that also seems to achieve the desired results, which I'm taking here to basically be the unit test assert statements: models are not equal, their model types are different, their data arrays are references to each other. And now it would seem their WCSs are also just references.

That sounds ideal to me.

emolter:
ok thanks I'll update the docs accordingly

emolter:
To note also, the fact that `on_init` is being called for models that are initialized from other models is new behavior.  IMO this is just better behavior, but let me know if you have thoughts or see potential issues.

emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/24042063573

tapastro:
I am in agreement with the final state this PR has reached - if tests are clean, I'm happy to approve.

docs/source/jwst/datamodels/models.rst
```diff
@@ -306,6 +307,26 @@ model into a new one using the :meth:`~stdatamodels.DataModel.update` method::
...
     new_model.update(old_model)
 
+While it's typically not recommended for general use due to the possibility of
+unexpected behavior, copying a model can be achieved by passing it into
+a constructor of the same type::
+
+    old_model = ImageModel()
+    new_model = ImageModel(old_model)
+
+This simply creates a new reference to the model object, i.e. ``new_model is old_model`` will be ``True``.
```
braingram: Does this need updating with recent changes? With this PR following this example does not result is `new_model is old_model` being true.
```python
m = dm.ImageModel()
m2 = dm.ImageModel(m)
assert m is not m2
```

docs/source/jwst/datamodels/models.rst
```diff
@@ -306,6 +307,26 @@ model into a new one using the :meth:`~stdatamodels.DataModel.update` method::
...
+Changes to one model will affect the other.
+
+It's also possible to convert between compatible model types this way, e.g.::
+
+    old_model = ImageModel()
+    new_model = IFUImageModel(old_model)
+
+and in this case, ``new_model is old_model`` will be ``False``. Simple metadata will be copied and
```
braingram: Would you clarify what "simple" metadata is here?
For example if I create a new model from a model then modify `meta.exposure.nframes` of the original it modifies the new one.

emolter: This was my original concern with Melanie's suggestion and I sort of forgot about it. I think the behavior is that anything at the top level of `meta` is now copied, but beyond that is not copied.  Does that gel with what you're seeing?  (I didn't re-test that recently).

What do you think is the desired behavior here?

docs/source/jwst/datamodels/models.rst
emolter: hmm, I wonder why this changed.  And it seems we should have a test for this specific case.

Again I would raise the question of what we think the desired behavior is.

emolter: ok I think I may have just confused myself here, and then written the wrong thing.  `m1 is m2` is False but `m1._instance is m2._instance` and `m1 == m2` are both True here.  I think that's fine for now, pending any future changes to `==` per https://github.com/spacetelescope/stdatamodels/issues/698

docs/source/jwst/datamodels/models.rst
emolter: ok I tested this again and indeed that's the behavior.  e.g. now the test looks like:
```
    input_model = FitsModel((50, 50))
    input_model.meta.exposure.type = "foo"
    input_model.meta.telescope = "test"
    with BasicModel(input_model) as dm:
        assert dm.data.shape == (50, 50)
        assert dm.meta.model_type == "BasicModel"
        assert input_model != dm
        assert input_model._instance is not dm._instance
        assert input_model.data is dm.data
        assert dm.meta.exposure.type == "foo"
        dm.meta.exposure.type = "bar"
        assert dm.meta.telescope == "test"
        dm.meta.telescope = "changed"
    assert input_model.meta.exposure.type == "bar" # did get modified
    assert input_model.meta.telescope == "test"  # did not get modified
    assert input_model.meta.model_type == "FitsModel" # did not get modified
    input_model.close()
```
which seems a bit confusing/arbitrary to me.

What if we instead somehow added special handling for the WCS, where all other "meta" would get deep-copied except WCS?  (this wouldn't be the first instance of special handling of the wcs attribute within stdatamodels fwiw)

melanieclarke: We could, but my preference would be to treat it as a shallow-as-possible copy.  If only meta.model_type (and date, filename) needs to be distinct, then the shallow copy on the top-level meta dictionary accomplishes the purpose.

I'd prefer to either go full deepcopy on the entire model so you can trust everything is distinct, or else to assume that it's a shallow copy and modifying most things about one will modify the other.   I think the most common use case here is to make a new model of the right type and discard the old one anyway, so it doesn't matter much if the original gets modified when the new one is updated.

emolter: Fine by me.  If others agree then it sounds like no code changes, just docs changes, are needed.

emolter: see https://github.com/spacetelescope/stdatamodels/pull/706/commits/bd20e0524caa9f076315b8b175576627de27637e

docs/source/jwst/datamodels/models.rst
emolter: see https://github.com/spacetelescope/stdatamodels/pull/706/commits/bd20e0524caa9f076315b8b175576627de27637e

docs/source/jwst/datamodels/models.rst
melanieclarke: That looks good to me.

braingram (approved):
Sounds like there is agreement on this as the new behavior. Changes LGTM.

emolter:
Cool along with Tyler's "if the tests pass I'll approve" it seems most people are on board, will go ahead and merge

PR #707, opened 2026-04-06
Expire deprecations for stdatamodels 6.0

### LINKED ISSUE #538: Remove deprecated **kwargs in asdf_in_fits.open and fits_support.from_fits_asdf
https://github.com/spacetelescope/stdatamodels/pull/533 deprecated the usage of **kwargs in the two methods mentioned in the title, as of whatever release comes after 4.0.1.  In a few releases after that one, the deprecation should turn into a removal.

### LINKED ISSUE #559: Remove `interpret_bit_flags`
This is to track when we can remove the unused `interpret_bit_flags`.

#486 deprecated it and as noted in https://github.com/spacetelescope/stdatamodels/issues/460 this deprecation should be in place for "2 builds". The change was released in 3.1.0 which is above the [1.18.1 requirements](https://github.com/spacetelescope/jwst/blob/b039b5f3426fe42f74e34cbec17033e05c8b49ed/pyproject.toml#L36) and [below the 1.19.0 ones](https://github.com/spacetelescope/jwst/blob/2b54b7de3b616edfc6062bd51f50facce40a3811/pyproject.toml#L35) so 1.19.0 is the first build with this deprecation. I'm assuming 1.19.1 doesn't count as the "next build" so after 1.20 or whatever the next build is released we can remove these.

@emolter can you confirm this schedule?

emolter:
Closes #538 
Closes #559 
Closes #703 

This PR removes all code and functionality that has been deprecated since before 5.0

## Tasks

</details>


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/24045135100

src/stdatamodels/dqflags.py
```diff
@@ -6,64 +6,7 @@
 If a given bit is set, the flag assigned to that bit is interpreted as being set or active.
 """
 
-import warnings
-
-from astropy.nddata.bitmask import interpret_bit_flags as ap_interpret_bit_flags
-
-from stdatamodels.basic_utils import multiple_replace
```
braingram, line 13: Was this the last use of `multiple_replace`? If so we can now deprecate all of basic_utils.

src/stdatamodels/jwst/datamodels/util.py
```diff
@@ -104,13 +78,12 @@ def open(init=None, guess=True, **kwargs):  # noqa: A001
...
-    if init is None:
-        warnings.warn(
-            "Passing None to open is deprecated; use the DataModel constructor directly instead.",
-            DeprecationWarning,
-            stacklevel=2,
+    if init is None or isinstance(init, (tuple, np.ndarray, fits.HDUList)):
+        raise TypeError(
+            f"Init type of {type(init).__name__} is not supported. "
+            "To initialize a model from scratch, from array, or from a `~astropy.io.fits.HDUList`, "
+            "use the DataModel constructors directly instead."
```
braingram: How about we allow these to fall through to the default?:
https://github.com/spacetelescope/stdatamodels/blob/869fabcfae68db762cb4697515882db9a3b4cb53/src/stdatamodels/jwst/datamodels/util.py#L196
That seems easier to track for future updates.

braingram (commented):
Thanks for removing these. 2 minor comments. 1 is to track a follow-up deprecation and 1 is a question about a new `TypeError` vs using an existing one.

src/stdatamodels/dqflags.py
emolter: Looks like yes.  Do you want me to do that right away here or in a follow-up?

src/stdatamodels/jwst/datamodels/util.py
emolter: https://github.com/spacetelescope/stdatamodels/pull/707/commits/d6a3811c05a97b990deec6b0aa495524021157b3

I had to catch HDUList separately still since it's list-like, otherwise the code would try to send it into the ModelContainer constructor

src/stdatamodels/dqflags.py
braingram: A follow-up would be easier to review.

braingram: Here's a follow-up issue https://github.com/spacetelescope/stdatamodels/issues/711

src/stdatamodels/jwst/datamodels/util.py
emolter: and https://github.com/spacetelescope/stdatamodels/pull/707/commits/3fa88c751ea2a0ffacaefc83ad628a587cfee232 or else we get `ValueError: Cannot open an association without the jwst package installed` instead of `TypeError` for tuple input if jwst isn't installed.

tapastro (approved):

PR #712, opened 2026-04-08
Add wcsinfo to multislitmodel schema

emolter:
Required to merge #710 and see discussion on that PR.

This PR adds `wcsinfo` to the `MultiSlitModel` schema.  It turns out that we request `wcsinfo` from `MultiSlitModel` in the pipeline and it seems it should therefore be part of the expected metadata (and as such get saved with the outputs).

## Tasks

</details>

### DIFF
--- a/changes/712.feature.rst
+++ b/changes/712.feature.rst
@@ -1,0 +1,1 @@
+Add wcsinfo to the MultiSlitModel schema
--- a/src/stdatamodels/jwst/datamodels/schemas/multislit.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/multislit.schema.yaml
@@ -4,7 +4,7 @@ $schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
 id: "http://stsci.edu/schemas/jwst_datamodel/multislit.schema"
 allOf:
 - $ref: core.schema
-- $ref: wcsinfo.schema
+- $ref: multislit_wcsinfo.schema
 - type: object
   properties:
     slits:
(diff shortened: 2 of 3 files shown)


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/24156781396

There is a lot of new metadata for this change.  I wonder if we are generally ok with that or not.

melanieclarke:
> https://github.com/spacetelescope/RegressionTests/actions/runs/24156781396
> 
> There is a lot of new metadata for this change. I wonder if we are generally ok with that or not.

The NIRSpec spec2 changes look okay – the only difference is in some empty comment-style headers. 

But it looks like this is now writing an imaging-type FITS WCS to WFSS  and NIRCam DHS exposures, which isn’t right.  That will probably need explicit handling.

melanieclarke (commented):
I think this set looks good, except possibly for the SYS keywords.  RADESYS, VELOSYS, and SPECSYS are FITS WCS keywords.  I don't think they'd do any harm on their own, but they shouldn't be needed, detached from any other WCS keywords.

emolter:
ok, thanks.  I left them in because they do exist already in the NIRSpec multislitmodel headers, so removing them would (along with the proposed model.update changes) lead to them no longer being there.  I assume based on your comment we are ok with that.

melanieclarke:
> ok, thanks. I left them in because they do exist already in the NIRSpec multislitmodel headers, so removing them would (along with the proposed model.update changes) lead to them no longer being there. I assume based on your comment we are ok with that.

Oh, good to know, thanks.  If they're there currently, let's keep them just in case.

emolter:
https://github.com/spacetelescope/jwst/pull/10464 handles this differently, moving relevant metadata into the slits of the multislitmodel instead of updating the multislitmodel schema.

PR #713, opened 2026-04-13
JP-4289: Add superstripe ramp model

melanieclarke:
Partially resolves [JP-4289](https://jira.stsci.edu/browse/JP-4289)

Add a new ramp model for superstripe mode to better reflect the data structure.

Allows pixeldq to be 3D for superstripe data, 2D for regular ramp models.

Requires coordination with jwst PR https://github.com/spacetelescope/jwst/pull/10446

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/ramp.py
+++ b/src/stdatamodels/jwst/datamodels/ramp.py
@@ -2,7 +2,7 @@ import numpy as np
 
 from .model_base import JwstDataModel
 
-__all__ = ["RampModel"]
+__all__ = ["RampModel", "SuperstripeRampModel"]
 
 
 class RampModel(JwstDataModel):
@@ -44,8 +44,6 @@ class RampModel(JwstDataModel):
         """
         Retrieve the schema-defined default value of an attribute.
 
-        Overrides the parent method to set pixeldq to a 2D array if
-        data is defined.
         Override the parent method to give zeroframe the correct shape
         if data is defined.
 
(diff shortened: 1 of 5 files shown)


melanieclarke:
Downstream and tests with jwst main are expected to fail for the superstripe case.

Regtests with the jwst PR:
https://github.com/spacetelescope/RegressionTests/actions/runs/24261764800

Expected diffs only: the intermediate products from the earliest stages of detector1 now have DATAMODL value `SuperstripeRampModel` instead of `RampModel`.

tapastro (approved):
LGTM!

src/stdatamodels/jwst/datamodels/ramp.py
```diff
@@ -65,16 +63,85 @@ def get_default(self, attr):
...
-            return np.zeros(self.data.shape[-2:], dtype=np.uint32)
-
         if attr == "zeroframe" and self.data is not None:
             shp = (self.data.shape[0],) + self.data.shape[-2:]
             return np.zeros(shp, dtype=np.float32)
 
         return super().get_default(attr)
+
+
+class SuperstripeRampModel(JwstDataModel):
```
emolter, line 73: Have you played with/considered making SuperstripeRampModel inherit directly from RampModel?  It looks like it would reduce repeats here, and I could imagine some cases where it'd be easier to check for model type.  But I also know that we don't use much inheritance in the datamodels generally

src/stdatamodels/jwst/datamodels/schemas/superstripe_ramp.schema.yaml
```diff
@@ -0,0 +1,44 @@
...
+      fits_hdu: ZEROFRAME
+      default: 0.0
+      ndim: 3
+      datatype: float32
+    refout:
+      title: Reference Output
+      fits_hdu: REFOUT
+      default: 0.0
+      ndim: 4
+      datatype: float32
```
emolter, line 41: `average_dark_current` is missing c.f. RampModel schema.  I assume that's intentional and that will never be needed here?

braingram, line 41: This drops average_dark_current. Would you comment on that change and add it to the changelog?

braingram: Resolving this as a duplicate to https://github.com/spacetelescope/stdatamodels/pull/713#discussion_r3079997022

src/stdatamodels/jwst/datamodels/ramp.py
melanieclarke: I thought about it, but wasn't sure it would be useful.  It's only a few repeated lines in the `__init__` and `get_default` that would be saved, and it introduces potential complexity.  I can add that if you think it's useful.

src/stdatamodels/jwst/datamodels/schemas/superstripe_ramp.schema.yaml
melanieclarke: Yes, it shouldn't be needed.  The `SuperstripeRampModel` is only used for the first few steps of detector1.  It gets replaced with a regular `RampModel` at the end of the refpix step, before the `dark_current` step.

emolter (commented):
I started reading through the JWST PR and I'm getting a little concerned we're doing too much here to just handle a single array that can be 2-D or 3-D.  We now have an entirely new copy of a schema, a new model type, and a new option in `Step.prepare_output` to choose the model type based on the input model's metadata.

What if we instead just added a custom `validate()` method to `RampModel` that checks the dimensionality of the `pixeldq` array based on if it's a superstripe mode or not?   Isn't this sort of what custom validate() overrides are for? And keep the `get_default` override.

Then superstripe modes would then still have `RampModel` type and `prepare_output` would need no changes (possible `GuiderRawModel` improvements could be a separate PR), but the pixel DQ array would still get validated.

I know this might be going back on what I said I preferred earlier but now I don't see much problem with it.

melanieclarke:
> I started reading through the JWST PR and I'm getting a little concerned we're doing too much here to just handle a single array that can be 2-D or 3-D. We now have an entirely new copy of a schema, a new model type, and a new option in `Step.prepare_output` to choose the model type based on the input model's metadata.
> 
> What if we instead just added a custom `validate()` method to `RampModel` that checks the dimensionality of the `pixeldq` array based on if it's a superstripe mode or not? Isn't this sort of what custom validate() overrides are for? And keep the `get_default` override.
> 
> Then superstripe modes would then still have `RampModel` type and `prepare_output` would need no changes (possible `GuiderRawModel` improvements could be a separate PR), but the pixel DQ array would still get validated.
> 
> I know this might be going back on what I said I preferred earlier but now I don't see much problem with it.

Yep, that's another way we could do it.  I liked the idea of a different datamodel for semantic reasons as much as the pixeldq handling.  It's not just the pixeldq that's different, the structure of all the data arrays has a different meaning for superstripe ramps than for regular ramp models, since the stripes are interleaved on the first data dimension.  It's not possible to interpret a superstripe ramp correctly if you're not aware that has superstripes, so it seemed useful to me to distinguish it as a different data type.

But you're right that there are pragmatic impacts to making a new datamodel that might be more important than the semantics.  I can explore the validation option so we can compare.  

@tapastro - do you have opinions on which direction is preferable?

melanieclarke:
> I can explore the validation option so we can compare.

@emolter - here's the alternate solution that does not make a new RampModel:
https://github.com/spacetelescope/stdatamodels/pull/715
https://github.com/spacetelescope/jwst/pull/10465

Let me know what you think.

tapastro:
I think both approaches look reasonable to me; one utilizes the flexibility that a datamodel can provide to let RampModel contain ramps of standard readouts and stripe readouts, while the other is more clear about the distinct content differences between the two. I'd have been happy with the flexible model approach, but I think for the sake of clarity and ease of future maintenance I prefer this one. Thank you for exploring both options!

PR #714, opened 2026-04-15
Speed up validation on assignment for simple types

emolter:
Relates to #690 

This PR makes validation on assignment a bit faster.

- Does not change `DataModel.validate()` runtime at all, because that passes a whole instance dict into `validate.value_change`.
- Improves runtime of all `ObjectNode.__setattr__` instances for simple types, if `validate_on_assignment` is True, which it usually is.  For my test case of just setting the instrument name 1e5 times, it was a bit better than a factor of 2 faster.

This is especially helpful if we move forward with https://github.com/spacetelescope/stdatamodels/pull/710 because that would make it so all attributes are set using `ObjectNode.__setattr__` during an `update` call.

## Tasks

</details>

### DIFF
--- a/changes/714.misc.rst
+++ b/changes/714.misc.rst
@@ -1,0 +1,1 @@
+Reduce overhead in validation of scalar values on assignment, minorly improving runtime of ObjectNode.__setattr__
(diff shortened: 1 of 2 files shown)


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/24462061014 

AMI issues are not related.

braingram (approved):
Thanks! Looks reasonable to me to pass those scalar types directly to validate.

PR #716, opened 2026-04-27
Compatibility with numpy 2.5

pllim:
This PR addresses deprecation warning seen in jwst devdeps, as reported by Melanie Clarke.

## Tasks
- [ ] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [x] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
- [x] [run `jwst` regression tests](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) with this branch installed (`stdatamodels@git+https://github.com/pllim/stdatamodels@numpydev-dtype`)
  - stable: https://github.com/spacetelescope/RegressionTests/actions/runs/25008333494 (passed)
  - devdeps: https://github.com/spacetelescope/RegressionTests/actions/runs/25009988560 (failures not related to deprecation warnings anymore)

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details>

PR #718, opened 2026-04-28
avoid dtype setting deprecated in numpy 2.5

braingram:
Numpy 2.5 will deprecate setting dtype. This update our handling of a fitsrec issue where the dtype is incorrectly reported to avoid setting the dtype.

Regression tests with this PR and `requirements-dev-thirdparty.txt` show unrelated (photutils 3.0?) failures:
https://github.com/spacetelescope/RegressionTests/actions/runs/25018852701

This is an alternative to and closes https://github.com/spacetelescope/stdatamodels/pull/716

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/util.py
+++ b/src/stdatamodels/util.py
@@ -68,10 +68,9 @@ def gentle_asarray(a, dtype, allow_extra_columns=False):
 
     # When a FITS file includes a pseudo-unsigned-int column, astropy will return
     # a FITS_rec with an incorrect table dtype.
+    # It's also unsafe to directly cast any FITS_rec with a
+    # pseudo-unsigned column.
     # https://github.com/astropy/astropy/issues/8862
-    # Due to an issue in astropy, it's not safe to directly cast
-    # a FITS_rec with a pseudo-unsigned column.
-    # See https://github.com/astropy/astropy/issues/8862
     if isinstance(a, fits.fitsrec.FITS_rec):
         if any(c.bzero is not None for c in a.columns):
             return _safe_asanyarray(a, out_dtype)
--- a/src/stdatamodels/util.py
+++ b/src/stdatamodels/util.py
@@ -67,14 +67,14 @@ def gentle_asarray(a, dtype, allow_extra_columns=False):
                 raise ValueError(msg)
 
     # When a FITS file includes a pseudo-unsigned-int column, astropy will return
-    # a FITS_rec with an incorrect table dtype.  The following code rebuilds
-    # in_dtype from the individual fields, which are correctly labeled with an
-    # unsigned int dtype.
-    # We can remove this once the issue is resolved in astropy:
+    # a FITS_rec with an incorrect table dtype.
     # https://github.com/astropy/astropy/issues/8862
+    # Due to an issue in astropy, it's not safe to directly cast
+    # a FITS_rec with a pseudo-unsigned column.
+    # See https://github.com/astropy/astropy/issues/12112
     if isinstance(a, fits.fitsrec.FITS_rec):
-        a.dtype = _rebuild_fits_rec_dtype(a)
-        in_dtype = a.dtype
+        if any(c.bzero is not None for c in a.columns):
+            return _safe_asanyarray(a, out_dtype)
 
     if in_dtype == out_dtype:
         return a
--- a/src/stdatamodels/util.py
+++ b/src/stdatamodels/util.py
@@ -71,7 +71,7 @@ def gentle_asarray(a, dtype, allow_extra_columns=False):
     # https://github.com/astropy/astropy/issues/8862
     # Due to an issue in astropy, it's not safe to directly cast
     # a FITS_rec with a pseudo-unsigned column.
-    # See https://github.com/astropy/astropy/issues/12112
+    # See https://github.com/astropy/astropy/issues/8862
     if isinstance(a, fits.fitsrec.FITS_rec):
         if any(c.bzero is not None for c in a.columns):
             return _safe_asanyarray(a, out_dtype)
@@ -161,7 +161,7 @@ def _safe_asanyarray(a, dtype):
         if any(c.bzero is not None for c in a.columns):
             # Due to an issue in astropy, it's not safe to directly cast
             # a FITS_rec with a pseudo-unsigned column.
-            # See https://github.com/astropy/astropy/issues/12112
+            # See https://github.com/astropy/astropy/issues/8862
             result = np.zeros(a.shape, dtype=dtype)
             for old_col, new_col in zip(a.dtype.names, result.dtype.names, strict=False):
                 result[new_col] = a[old_col]


src/stdatamodels/util.py
```diff
@@ -67,14 +67,14 @@ def gentle_asarray(a, dtype, allow_extra_columns=False):
...
-    # in_dtype from the individual fields, which are correctly labeled with an
-    # unsigned int dtype.
-    # We can remove this once the issue is resolved in astropy:
+    # a FITS_rec with an incorrect table dtype.
     # https://github.com/astropy/astropy/issues/8862
+    # Due to an issue in astropy, it's not safe to directly cast
+    # a FITS_rec with a pseudo-unsigned column.
+    # See https://github.com/astropy/astropy/issues/12112
     if isinstance(a, fits.fitsrec.FITS_rec):
-        a.dtype = _rebuild_fits_rec_dtype(a)
```
braingram, line 76: `_rebuild_fits_rec_dtype` works around https://github.com/astropy/astropy/issues/8862 where astropy misreports dtypes for unsigned integer columns.
If the fitsrec here contains an unsigned integer column, the safest thing to do is to call `_safe_asanyarray` which contains special code to handle this case (and another issue with astropy):
https://github.com/spacetelescope/stdatamodels/blob/ffb7c6f3a7254eb5c2e1c4b152961da895e47899/src/stdatamodels/util.py#L162-L164
Without handling that, simple operations like creating views can result in unsigned columns being silently changed to incorrect values.

pllim (commented):
I don't have enough FITS internal knowledge to know that this patch is a one-to-one replacement. There could be edge cases that this new check might overlook but maybe you are not affected anyway, so I will defer to @melanieclarke .

src/stdatamodels/util.py
melanieclarke: It looks like 12112 is the same issue, since it's marked as a duplicate to 8862, so maybe we can clean up the comments here and in the `_safe_asanyarray` function to just point to 8862.

melanieclarke (commented):
I also am not familiar enough with this section of stdatamodels to feel confident reviewing, but let me take a stab at understanding what's going on here.

It looks like `gentle_asarray` is used for casting data to match the schema when setting datamodel attributes, and it looks like `_safe_asanyarray` is used as a fallback only if the input dtype is not compatible with the output dtype.  That function makes a new copy of the array, so it should be avoided if possible.  

The change here is to always call `_safe_asanyarray` when a FITS_rec table contains a pseudo-unsigned integer, instead of attempting to correct the bad input dtype and return the original array if possible.  Is that right?

Do we use pseudo-unsigned integers anywhere in our own code or products?  If not, this seems reasonable to me as a safe workaround for a rare edge case.  If we do use them, say in uncal products, we might need to look at whether this handling impacts performance for large files.

braingram:
Thanks for taking a look at this.

> The change here is to always call `_safe_asanyarray` when a FITS_rec table contains a pseudo-unsigned integer, instead of attempting to correct the bad input dtype and return the original array if possible. Is that right?

Yes, for any fitsrec with an unsigned column call `_safe_asanyarray`.
 
> Do we use pseudo-unsigned integers anywhere in our own code or products? If not, this seems reasonable to me as a safe workaround for a rare edge case. If we do use them, say in uncal products, we might need to look at whether this handling impacts performance for large files.

We do have unsigned integer columns in some data (unsigned arrays not in tables aren't impacted here). TSOSpecModel is one:
https://github.com/spacetelescope/stdatamodels/blob/ffb7c6f3a7254eb5c2e1c4b152961da895e47899/src/stdatamodels/jwst/datamodels/schemas/tso_spectable.schema.yaml#L61-L62
IRS2Model is another:
https://github.com/spacetelescope/stdatamodels/blob/ffb7c6f3a7254eb5c2e1c4b152961da895e47899/src/stdatamodels/jwst/datamodels/schemas/irs2.schema.yaml#L37

`gentle_asarray` is only called from `_cast`:
https://github.com/spacetelescope/stdatamodels/blob/ffb7c6f3a7254eb5c2e1c4b152961da895e47899/src/stdatamodels/properties.py#L92
which is called with potentially a fitsrec in `from_fits_hdu`:
https://github.com/spacetelescope/stdatamodels/blob/ffb7c6f3a7254eb5c2e1c4b152961da895e47899/src/stdatamodels/fits_support.py#L919
In this case the rebuilt dtype will always fail the later `in_dtype == out_dtype` due to endian differences which will then trigger `_safe_asanyarray` here:
https://github.com/spacetelescope/stdatamodels/blob/ffb7c6f3a7254eb5c2e1c4b152961da895e47899/src/stdatamodels/util.py#L92
but another way, when opening a fits file the endian difference makes rebuilding the dtype unnecessary since the data is later copied.

The only other place `_cast` is called is within `properties.py` when a FITS_rec is assigned to the model. This is pretty rare (thankfully since FITS_rec misbehaves in many ways). There are 2 unnecessary uses in jwst unit tests:
https://github.com/spacetelescope/jwst/blob/a9d22c955d0e5ff68f3478bf2039fdca7b41ce04/jwst/datamodels/utils/tests/test_flat_multispec.py#L256
https://github.com/spacetelescope/jwst/blob/a9d22c955d0e5ff68f3478bf2039fdca7b41ce04/jwst/spectral_leak/tests/test_spectral_leak_step.py#L19
(I have a regtest run trying to ferret out further usage). It's dangerous to take views for these assignments as can be seen with the following example.
```python
import stdatamodels.jwst.datamodels as dm, numpy as np
from astropy.io import fits
m = dm.IRS2Model()
fr = np.array([(1, 1, 1), (2, 2, 2)], dtype=m.get_dtype("dq_table")).view(fits.FITS_rec)
m.dq_table = fr
assert m.dq_table['OUTPUT'][0] == 1
m.save("bar.fits")
assert m.dq_table['OUTPUT'][0] == 65537
```
since the assignment did not copy and the fits_rec contained an unsigned column the file saving silently corrupted the data. With this PR the last assert fails since the data is no longer corrupted and `m.dq_table['OUTPUT'][0] == 1`.

I don't see any significant performance differences as reported in the regtest summary page. Let me know if any other tests would be helpful.

I'll add a new comment when the regtest run attempting to find other view of fitsrecs with unsigned columns.

melanieclarke:
Thanks for the detailed explanation!  I think I almost understand.

> when opening a fits file the endian difference makes rebuilding the dtype unnecessary since the data is later copied.

Okay, thanks.  Opening FITS files was the use case I was most concerned about, and it sounds like we were always using the copy in that case anyway.

> The only other place `_cast` is called is within `properties.py` when a FITS_rec is assigned to the model. This is pretty rare (thankfully since FITS_rec misbehaves in many ways). 

You mean, as opposed to assigning a numpy record to the model to be converted to a FITS_rec?  We do that frequently. But that case avoids the problem by never round tripping through a FITS_rec cast?

braingram:
> You mean, as opposed to assigning a numpy record to the model to be converted to a FITS_rec? We do that frequently. But that case avoids the problem by never round tripping through a FITS_rec cast?

Yeah as opposed to assigning a numpy record array. In that case the thing being "cast" is a numpy record array and fails the `isinstance(a, fits.fitsrec.FITS_rec)` check and doesn't hit the code on main that incompatible with numpy 2.5.

braingram:
https://github.com/spacetelescope/RegressionTests/actions/runs/25066483257/attempts/1 is a regtest run with jwst main and a modified stdatamodels branch which raises an exception if the cast fitsrec:
- contains an unsigned column
- would pass the in_dtype == out_dtype check if the rebuilt dtype was assigned

https://github.com/spacetelescope/stdatamodels/compare/main...braingram:stdatamodels:risky_view

it shows only unit test failures for the 2 lines of code commented above. I'll rerun after pushing updates to those 2 tests which remove the fitsrec casts. EDIT: rerunning with those 2 tests updated (https://github.com/spacetelescope/jwst/pull/10495) all passes indicating no other instances where a fitsrec with an unsigned column is cast during assignment.

These failures should be the only cases where a fitsrec is copied on an assignment with this PR. So with this PR those tests pass but with this PR result in an extra copy.

melanieclarke:
Okay, got it.  I'm satisfied that these changes are adding an extra data copy only in very rare edge cases, so we should be safe to proceed.

src/stdatamodels/util.py
braingram: Updated issue numbers in https://github.com/spacetelescope/stdatamodels/pull/718/commits/e614fd5f38e10110b7062a0ad92063f36ab13024

src/stdatamodels/util.py
```diff
@@ -67,14 +67,14 @@ def gentle_asarray(a, dtype, allow_extra_columns=False):
...
     # When a FITS file includes a pseudo-unsigned-int column, astropy will return
-    # a FITS_rec with an incorrect table dtype.  The following code rebuilds
-    # in_dtype from the individual fields, which are correctly labeled with an
-    # unsigned int dtype.
-    # We can remove this once the issue is resolved in astropy:
+    # a FITS_rec with an incorrect table dtype.
     # https://github.com/astropy/astropy/issues/8862
+    # Due to an issue in astropy, it's not safe to directly cast
+    # a FITS_rec with a pseudo-unsigned column.
+    # See https://github.com/astropy/astropy/issues/8862
```
melanieclarke: One small nitpick: this new comment is redundant with the one above.  Can you combine them?

melanieclarke (approved):
This looks good to me now.  I have one small quibble about the added comment, but it doesn't need to hold this PR up, so I will approve now.

src/stdatamodels/util.py
braingram: How does cbb07a7c look?

melanieclarke: Better, thanks!

pllim:
Thanks! Also FWIW I noticed https://github.com/astropy/astropy/pull/19585

PR #720, opened 2026-04-28
update changelog categories

zacharyburnett:
updates changelog categories to be more in line with pipeline towncrier fragments:
- `breaking`
- `schema`
- `feature`
- `bugfix`
- `docs`
- `other`

these are ordered in the changelog output, and should make it easier to follow semantic versioning for releases if we want to do that

## Tasks

</details>


.github/pull_request_template.md
```diff
@@ -12,15 +12,6 @@ This PR addresses ...
 ## Tasks
 - [ ] update or add relevant tests
 - [ ] update relevant docstrings and / or `docs/` page
-- [ ] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
-  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
-  - [ ] [run `jwst` regression tests](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) with this branch installed (`"git+https://github.com/<fork>/stdatamodels@<branch>"`)
-
-<details><summary>news fragment change types...</summary>
-
-- ``changes/<PR#>.feature.rst``: new feature
```
emolter, line 21: I know removing these from the template is consistent with other repositories, but I have to say I really don't like their removal.  Normally when I'm writing a changelog fragment I'm already on the PR summary page because I need the PR number.

With these in the template, I need only click the dropdown from the same page to see my options.

With these no longer in the template as for `jwst`, I now find myself navigating away from this page, scrolling down in the jwst repo GitHub readme, failing to find what I want, and then running a cmd+F search on "news fragment" in my VSCode instance to see my options.  I know it's a small thing but it bugs me.

changes/README.md
emolter, line 1: We should probably add some text somewhere here that defines public API.  It's more in-line with standard open-source standard practices than `jwst`, but because `jwst` is different it may be worth specifying our policy.

changes/README.md
```diff
@@ -0,0 +1,22 @@
...
+News fragment filenames consist of the pull request number and the changelog category (see below).
+A single change can have more than one news fragment, if it spans multiple categories.
+
+## Change Log Categories
+
+Make a news fragment for every relevant category affected by your change.
+
+- `<PR#>.breaking.rst`: Add this fragment if your change **breaks public API**, describing what the user needs to change
+- `<PR#>.feature.rst`
+- `<PR#>.fix.rst`
```
emolter: Why is "bugfix" becoming "fix"?

pyproject.toml
emolter, line 1: Only semi-related to this PR, but should we additionally consider adding more of these and/or making them more specific?  I'm thinking mostly about all the "feature" fragments that are very minor updates to schemas of specific model types.  It seems logical to separate these somehow from new features in the base code.

emolter (commented):
@braingram should probably also comment

changes/README.md
zacharyburnett: i figure more concise (conveying the same info) is better and reduces the chance of a typo (that's why I used `fix` in the upcoming package template). But we can revert this too if it's too much of a change

pyproject.toml
zacharyburnett: oh good point, what categories were you thinking of?

.github/pull_request_template.md
zacharyburnett: added back in cd1b0dd

braingram: FWIW I never trust the PR template or docs and instead open the pyproject.toml file before making a fragment.

braingram:
With any change here we need to rename all the existing fragments.

To be honest I liked that this package used the standard fragment types but I see the point about the proposed new ones making it easier to tell what requires a major version bump.

zacharyburnett:
> With any change here we need to rename all the existing fragments.
> 
> To be honest I liked that this package used the standard fragment types but I see the point about the proposed new ones making it easier to tell what requires a major version bump.

We could also just use the default ones with `removal` replaced with `breaking` and reordered to be first. I figured it might be easier to remember if we use a similar set of fragment names across all our repos, but we will indeed have some custom ones per-repository and so we'll need to check the definitions anyway when making new PRs

changes/README.md
emolter: I'm fine either way it just might take a bit of adjusting to not forget this

pyproject.toml
emolter: If we (mostly) go with Brett's suggestion of just sticking with standard ones, then this is moot.  But I was thinking something along the lines of "schema_update".

zacharyburnett: ok, done!

melanieclarke: Just chiming in to say I like this idea.  I always felt a bit weird about marking a minor schema change as a "feature", and it will be a more useful category for users trying to track down what changed and why.

pllim:
As mentioned in Team Chartreuse tag up, I am curious on who is the intended audience of these change logs. Are they for pipeline devs or end-users (e.g., Jdaviz users who need this package to open data without the pipeline)?

If the latter, category like "schema" would mean little to them. They would want to know which instrument changed what data format and if it is backward compatible with the data they already have.

emolter:
@pllim I find it a bit doubtful that most end-user scientists read these changelogs in detail.  So to me, the primary consumers of this would be:

- STScI folks who e.g. create reference files
- pipeline devs
- devs of downstream packages like Jdaviz
- possibly external users, if they are power-users

The name wouldn't have to be "schemas".  My idea was simply to separate new code features from schema updates somehow.  But given that providing schemas is a huge part of what stdatamodels does, I'm not sure there's any rename that's easier to understand.

A problem with separating this by instrument or mode is that many changes don't fall into a single category.  Have a look at https://github.com/spacetelescope/stdatamodels/blob/main/CHANGES.rst#new-features 
The first three entries there, all of which were schema updates, affected:

1. all extract1d files
2. all files containing spectra
3. the core schema

pllim:
So if changes to a schema is a breaking change, do we then need 2 fragments, one in breaking and one in schema?

zacharyburnett:
> So if changes to a schema is a breaking change, do we then need 2 fragments, one in breaking and one in schema?

yeah exactly; the `schema` fragment explains the change, and the `breaking` fragment explains which schemas will break existing code (and what users can do to update)

pllim:
I am not sure how splitting the change summary into 2 different sections is gonna make things less confusing...

emolter:
@pllim even if we don't change the category name, in order to be in-line with what's already done in `jwst` and elsewhere we will need duplicated fragments.  AFAIK in jwst it's the new standard practice for, say, the removal of a step-level parameter in JumpStep to have both a "breaking" and "jump" fragment.

changes/README.md
```diff
@@ -0,0 +1,23 @@
...
+A single change can have more than one news fragment, if it spans multiple categories.
+
+## Change Log Categories
+
+Make a news fragment for every relevant category affected by your change.
+
+- `<PR#>.breaking.rst`: Add this fragment if your change **breaks public API**, describing what the user needs to change
+- `<PR#>.schema.rst`: schema updates
+- `<PR#>.feature.rst`
+- `<PR#>.fix.rst`
```
braingram:
```suggestion
- `<PR#>.bugfix.rst`
```
Let's stick with the standard naming.

pyproject.toml
```diff
@@ -147,16 +147,32 @@ ignore = [".gitkeep"]
...
-name = "Documentation"
-
-[tool.towncrier.fragment.removal]
-name = "Deprecations and Removals"
-
-[tool.towncrier.fragment.misc]
+showcontent = true
+
+[[tool.towncrier.type]]
+directory = "fix"
```
braingram:
```suggestion
directory = "bugfix"
```

braingram (changes requested):
Now may be an opportune time to put this in (since we have no fragments).

Would you resolve the conflicts and rename "fix" to "bugfix"? Otherwise LGTM.

changes/773.schema.rst
braingram, line 1: Would you rename this to schema instead of breaking?

.github/pull_request_template.md
```diff
@@ -12,15 +12,16 @@ This PR addresses ...
...
 
-- ``changes/<PR#>.feature.rst``: new feature
-- ``changes/<PR#>.bugfix.rst``: fixes an issue
-- ``changes/<PR#>.doc.rst``: documentation change
-- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
-- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
+- `<PR#>.breaking.rst`: Add this fragment if your change **breaks public API**, describing what the user needs to change
+- `<PR#>.schema.rst`: schema updates
+- `<PR#>.feature.rst`
+- `<PR#>.fix.rst`
```
braingram:
```suggestion
- `<PR#>.bugfix.rst`
```

braingram (approved):

PR #722, opened 2026-04-29
JP-4103: Allow FAILED status for all cal steps

melanieclarke:
Resolves [JP-4103](https://jira.stsci.edu/browse/JP-4103)

Add "FAILED" as an option for all cal_step status keywords.

## Tasks

</details>

### DIFF
--- a/changes/722.feature.rst
+++ b/changes/722.feature.rst
@@ -1,0 +1,1 @@
+Add "FAILED" as an option for all ``cal_step`` status keywords.
--- a/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
@@ -2379,37 +2379,37 @@ properties:
           adaptive_trace_model:
             title: Adaptive Trace Model
             type: string
-            enum: [SKIPPED, COMPLETE]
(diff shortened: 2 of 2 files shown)


melanieclarke:
Regtests with jwst on main:
https://github.com/spacetelescope/RegressionTests/actions/runs/25123363235

All passing.

braingram (approved):
Schema changes LGTM.

One thing to consider (perhaps as a follow up or not, I'm not sold on the idea). Is to consolidate these into a definition. Adding the following to the top of the core schema:
```yaml
definitions:
  cal_step:
    type: string
    enum: [SKIPPED, COMPLETE, FAILED]
    blend_table: True
```
would allow updating all the cal_step entries to something like the following:
```yaml
           adaptive_trace_model:
             allOf:
               - title: Adaptive Trace Model
                 fits_keyword: S_TRCMOD
               - $ref: "#/definitions/cal_step"
```
which is arguably harder to read but does make changes like the one in this PR easier to do. I did a basic test of this locally (modified the schema, set a cal_step checking that it was validated, saved the file, confirmed it was written to the fits header) so at least the infrastructure for it works.

melanieclarke:
Thanks @braingram.  I'll go ahead and merge the schema update now, and we can follow up on the definition consolidation later if desired.

PR #730, opened 2026-05-06
Add SUBPRISM_IP/S subarray names

tapastro:
Partially resolves [JP-4150](https://jira.stsci.edu/browse/JP-4150)

This PR addresses a missed mapping from the SLITLESSPRISM_IP/S subarray names to intermediate SUBPRISM_IP/S names - to support processing of the data, we will include these subarray names. We expect DMS builds 13.0+ will have SDP translate these back to SLITLESSPRISM_IP/S, but that has not yet been completed.

## Tasks
- [ ] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [x] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [x] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
  - [x] [run `jwst` regression tests](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) with this branch installed (`"git+https://github.com/<fork>/stdatamodels@<branch>"`)

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details>

PR #731, opened 2026-05-06
Revert "JP-4189: Keyword title changes to match stdatamodels with jwstkd"

tapastro:
Reverts spacetelescope/stdatamodels#621


braingram (approved):

tapastro:
@braingram Due to the urgency of the requested DMS patch, I'm going to revert this PR and run tests on the patch PR, #730.  I'll get this back in when time pressure is not so high.

PR #732, opened 2026-05-06
Revert "Revert "JP-4189: Keyword title changes to match stdatamodels with jwstkd""

braingram:
Reverts spacetelescope/stdatamodels#731


emolter:
image

PR #734, opened 2026-05-06
JP-4273: Allow multiple apertures in PSF file

melanieclarke:
Resolves [JP-4273](https://jira.stsci.edu/browse/JP-4273)

To support optimal extraction for NIRSpec FS, we need to allow the SpecPsfModel to contain multiple apertures, to match all possible fixed slits.

Existing SpecPsfModels for MIRI LRS have some keywords in the primary header that will need to be moved to the extension header, so I've added some migration code for that.  Otherwise, existing PSF files should read in as is just fine.

JWST spectral extraction code will need to be updated to match the new schema. PR is: https://github.com/spacetelescope/jwst/pull/10516

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
@@ -12,6 +12,8 @@ allOf:
 - $ref: keyword_pexptype.schema
 - $ref: keyword_filter.schema
 - $ref: keyword_pfilter.schema
+- $ref: keyword_grating.schema
+- $ref: keyword_pgrating.schema
 - type: object
   properties:
     apertures:
--- a/src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
@@ -5,46 +5,49 @@ id: "http://stsci.edu/schemas/jwst_datamodel/specpsf.schema"
 allOf:
 - $ref: referencefile.schema
 - $ref: keyword_band.schema
+- $ref: keyword_pband.schema
 - $ref: keyword_channel.schema
+- $ref: keyword_pchannel.schema
 - $ref: keyword_exptype.schema
-- $ref: keyword_readpatt.schema
-- $ref: keyword_psubarray.schema
+- $ref: keyword_pexptype.schema
 - $ref: keyword_filter.schema
-- $ref: subarray.schema
-
+- $ref: keyword_pfilter.schema
 - type: object
   properties:
-    meta:
-      type: object
-      properties:
-        psf:
-          title: PSF parameters
-          type: object
-          properties:
-            subpix:
-              title: oversampling factor
-              type: number
-              fits_keyword: SUBPIX
-            center_col:
-              title: column ePSF shifted to
-              type: number
-              fits_keyword: CENTCOL
-            center_row:
-              title: row ePSF shifted to
-              type: number
-              fits_keyword: CENTROW
-- type: object
-  properties:
-    data:
-      title: The PSF image
-      fits_hdu: PSF
-      default: 0.0
-      ndim: 2
-      datatype: float64
-    wave:
-      title: Wavelength image
-      fits_hdu: WAVE
-      default: 0
-      ndim: 1
-      datatype: float64
-
+    apertures:
+      type: array
+      title: Array of PSF models
+      items:
+        type: object
+        properties:
+          name:
+            title: Aperture name
+            fits_keyword: APERTURE
+            fits_hdu: PSF
+          subpix:
+            title: Oversampling factor
+            type: number
+            fits_keyword: SUBPIX
+            fits_hdu: PSF
+          center_col:
+            title: Center column (vertical dispersion)
+            type: number
+            fits_keyword: CENTCOL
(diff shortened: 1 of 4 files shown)


src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
```diff
@@ -5,46 +5,49 @@ id: "http://stsci.edu/schemas/jwst_datamodel/specpsf.schema"
...
 - $ref: keyword_channel.schema
+- $ref: keyword_pchannel.schema
 - $ref: keyword_exptype.schema
-- $ref: keyword_readpatt.schema
-- $ref: keyword_psubarray.schema
+- $ref: keyword_pexptype.schema
 - $ref: keyword_filter.schema
-- $ref: subarray.schema
-
+- $ref: keyword_pfilter.schema
```
emolter, line 14: Are all these changes related to the PR?  Why does the PR require the keywords queried by CRDS to change?  My (limited) understanding of the goal here is just to allow multiple apertures, not to change the mapping between science data files and associated reference files

src/stdatamodels/jwst/datamodels/specpsf.py
```diff
@@ -3,19 +3,63 @@
...
+    hdulist : HDUList
+        FITS HDUList to read.
+
+    Returns
+    -------
+    hdulist : HDUList
+        The updated HDUList.
+    """
+    keys_to_copy = ["APERTURE", "CENTCOL", "CENTROW", "SUBPIX"]
+    if len(hdulist) == 3:
```
emolter: Wouldn't a real (out-of-date) data product have a fourth extension, the ASDF table?  Is there a different way to ensure that this is the expected old ref file type besides checking the number of extensions?  Wouldn't this conditional also be triggered if this is a new-style model that just happened to have the right number of apertures?

src/stdatamodels/jwst/datamodels/specpsf.py
```diff
@@ -3,19 +3,63 @@
...
 
+    PSF models are contained in a list, under the ``apertures``
+    attribute.
+
     Attributes
     ----------
-    data : numpy float32 array
-         The PSF image
-    wave : numpy float32 array
-         Wavelength image
```
emolter, line 15: old docs were out of sync with the float64 data type in the schema.  Have we confirmed which one was intended, float32 or float64?

emolter (commented):
Thanks Melanie!  I had a few questions.

src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
melanieclarke: This was intended as a clean up for the existing keywords.  CRDS matching for MIRI LRS currently uses filter, channel, band, and exp_type.  It does not use readpatt or subarray so I removed those.  I also removed the subarray schema because the reference file doesn't need it.

I added the corresponding `p*` schemas to allow more flexibility in CRDS matching, in case MIRI or NIRSpec needs it in the future, but I'm not totally sure that's how those work, and can remove them if that's preferable.

src/stdatamodels/jwst/datamodels/specpsf.py
melanieclarke: All good points.  This worked for the existing MIRI LRS PSF files, but I can try to find a more general way to identify them.

src/stdatamodels/jwst/datamodels/specpsf.py
melanieclarke: float64 is correct.

src/stdatamodels/jwst/datamodels/specpsf.py
melanieclarke: Updated in [c6234de](https://github.com/spacetelescope/stdatamodels/pull/734/commits/c6234de800bf32d494d66c705c62c227869cbb7a), should be more robust now.

src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
emolter: can't say I know how the `p*` schemas interact with CRDS either unfortunately

src/stdatamodels/jwst/datamodels/specpsf.py
emolter: this does look better to me, thanks!

src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
melanieclarke: Following up with NIRSpec, they also need the grating keyword to match to, and may also use the `pexptype` and `pgrating` extended matching capability.

CRDS ticket for initial NIRSpec PSF definition is: https://jira.stsci.edu/browse/CRDS-1002

tapastro (approved):

src/stdatamodels/jwst/datamodels/schemas/specpsf.schema.yaml
tapastro: My understanding is that the keyword pattern schemas are required to match rmap selectors with pipes, but I have never been confident in why we have them. Perhaps something for Jonathan to enlighten us on?

PR #738, opened 2026-05-11
Fix handling of logical int8 columns

braingram:
Closes https://github.com/spacetelescope/stdatamodels/issues/735

Logical/int8 columns are causing warnings for astropy dev and do not work as expected (see above issue).

Update handling of structured arrays/fits_rec with logical/int8 columns copying data where needed to handle the mismatch between numpy bool storage and astropy/fits logical bool mapping.

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/25576691180/job/75368198893

github is having issues so CI is unreliable. Will trying to restart some tests.

## Tasks

</details>

### DIFF
--- a/tests/jwst/datamodels/test_models.py
+++ b/tests/jwst/datamodels/test_models.py
@@ -807,6 +807,19 @@ def test_amioi_model_extra_meta(tmp_path, oifits_ami_model):
     oifits_ami_model.save(fn)
 
 
+@pytest.mark.parametrize("value", [True, False])
+def test_amioi_logical_flag(value):
+    """
+    Test that int8 stored "FLAG" columns accept bools.
+
+    This is a targeted test to reproduce a bug found in:
+    https://github.com/spacetelescope/stdatamodels/issues/735
+    """
+    model = datamodels.AmiOIModel()
+    model.t3 = [(0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, [0, 0, 0], value)]
+    assert model.t3[0][-1] == value
+
+
 def test_dq_def_roundtrip(tmp_path, test_data_path):
     """
     Open a MaskModel with a defined DQ array and dq_def that modifies the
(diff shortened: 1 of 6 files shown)


tests/jwst/datamodels/test_models.py
```diff
@@ -807,6 +807,19 @@ def test_amioi_model_extra_meta(tmp_path, oifits_ami_model):
     oifits_ami_model.save(fn)
 
 
+@pytest.mark.parametrize("value", [True, False])
```
emolter: Do numeric values 0 and 1 also work here? (I assume they would get converted to bool by Python, just making sure?)

tests/schemas/table_model.yaml
```diff
@@ -8,8 +8,10 @@ allOf:
     properties:
       table:
         fits_hdu: TABLE
-        default: [0, 42, NaN, "foo", 42.0, 0.0, 0.0, 0.0]
+        default: [False, 0, 42, NaN, "foo", 42.0, 0.0, 0.0, 0.0]
```
emolter, line 11: Related to my last comment, this concerns me a bit - why can't it be 0?

emolter (commented):
looks good overall but I had a question

tests/jwst/datamodels/test_models.py
braingram: Good question! They do. I updated the test to check some including ord(T) ord(F) and test round-tripping to a file: 9454993d

tests/schemas/table_model.yaml
braingram: It can be. I updated this since I figured a default of `False` would be easier to understand since that's what the value becomes (or more precisely the numpy `False`). I'm happy to revert this change if preferred.

emolter: ok, good.  I'm fine with this change.  I incorrectly assumed you made the change because it would fail otherwise.

tests/jwst/datamodels/test_models.py
emolter: Thanks!  The tests look good to me

emolter (approved):
This looks good to me until we can remove fits_rec entirely

tests/schemas/table_model.yaml
braingram: Sounds good. FYI I tested locally and reverting this change doesn't cause any unit test failures.

PR #739, opened 2026-05-11
JP-4313: Add NRC superstripe subarrays

melanieclarke:
Toward [JP-4313](https://jira.stsci.edu/browse/JP-4313)

Add new subarray names for upcoming NIRCam superstripe modes

Also add `meta.exposure.read_times` to the core schema to allow specification of explicit read times for unevenly sampled readouts.  This is needed to support jump detection and ramp fitting for in-frame reads of repeated substripes.

## Tasks

</details>


src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1016,6 +1016,13 @@ properties:
             title: Snowballs/showers per million pixels per second
             type: number
             fits_keyword: EXTNCRS
+          read_times:
```
braingram, line 988: Is there any expectation that a user might want to access these values directly from a FITS file? If so what about storing this in a table in a separate extension?

melanieclarke: It seems unlikely to me that users would need to access these values -- they're only relevant for some of the detailed calculations in detector1.  FITS tables are such a pain to use I hate to make one we don't need.  But maybe?  @tapastro - opinions?

melanieclarke: The more I think about it, the less I want to put these values into a table.  In the most common use case, this list will contain exactly 4 numbers, which are easily recalculated from metadata if a user needs them for some out-of-pipeline purpose.  A whole FITS table seems excessive for the purpose, and much harder to structure correctly than a list of lists.

The calculation is in the `jwst.lib.stripe_utils.stripe_read_times` function in the corresponding jwst PR:
https://github.com/spacetelescope/jwst/pull/10534/changes#diff-cf14a77d25397212c8babf6f62540e157125a0bac4af4c3d67c47b8d70518482R857

braingram (approved):
Changes LGTM. I consulted https://jira.stsci.edu/browse/APT-94701 for the subarray names but if there is a more definitive ticket let me know.

tapastro (approved):

PR #740, opened 2026-05-12
Ensure shape attribute stays in-sync with primary array

### LINKED ISSUE #681: Model shape attribute can become out-of-sync with primary array
Consider the following:
```
import numpy as np
data = np.ones((10,10))
model = dm.ImageModel(data=data)
model.shape # 10, 10
model.data = np.ones((5,5))
model.shape # 10, 10
model.data.shape # (5,5)
```

This happens because the `model.shape` property definition only queries the primary array shape if `model._shape` is `None`; otherwise it just returns `model._shape`.  This should be fixed in some way that prevents these from being different.

`model._shape` is set either on `__init__` (if the primary array was provided on initialization of the model) or else the first time `model.shape` is requested, e.g.
```
model = dm.ImageModel()
model.shape # None
model.data = np.ones((10,10))
model._shape # None
model.shape # (10,10)
model._shape # (10,10)
```

Relates to https://github.com/spacetelescope/stdatamodels/issues/505

emolter:
Partially addresses [JP-4293](https://jira.stsci.edu/browse/JP-4293)

Closes #681 

This PR causes the `model.shape` attribute to always query the primary array shape and return that, fixing a bug where `model.shape` could become stale when the primary array was re-set.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -674,12 +674,16 @@ class DataModel(properties.ObjectNode):
     @property
     def shape(self):
         """Return the shape of the primary array."""
-        if self._shape is None:
-            primary_array_name = self.get_primary_array_name()
-            if primary_array_name and self.hasattr(primary_array_name):
-                primary_array = getattr(self, primary_array_name)
-                self._shape = primary_array.shape
-        return self._shape
+        primary_array_name = self.get_primary_array_name()
+        if primary_array_name and self.hasattr(primary_array_name):
+            primary_array = getattr(self, primary_array_name)
+            self._shape = primary_array.shape
+            return self._shape
+        return None
+
+    @shape.setter
+    def shape(self, value):
+        raise AttributeError("Model shape is read-only.")
 
     def __setattr__(self, attr, value):
         if attr in frozenset(("shape", "history", "_extra_fits", "schema")):
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -176,7 +176,7 @@ class DataModel(properties.ObjectNode):
         # Determine what kind of input we have (init) and execute the
         # proper code to initialize the model
         self._file_references = []
-        self._shape = None  # temporary seed used by _make_default_array via ctx.shape
+        self._shape = None
         is_array = False
         is_shape = False
         shape = None
@@ -266,8 +266,6 @@ class DataModel(properties.ObjectNode):
             raise TypeError(f"Can't initialize datamodel using {str(type(init))}")
 
         # Initialize object fields as determined from the code above
-        # _shape seeds ctx.shape for _make_default_array before the primary array exists.
-        # It is cleared to None once the primary array is stored in _instance.
         self._shape = shape
         self._instance = asdffile.tree
         self._asdf = asdffile
@@ -291,7 +289,6 @@ class DataModel(properties.ObjectNode):
                     "has no primary array in its schema"
                 )
             setattr(self, primary_array_name, init)
-            self._shape = None
 
         # If a shape has been given, initialize the primary array.
         if is_shape:
@@ -302,9 +299,8 @@ class DataModel(properties.ObjectNode):
                     "has no primary array in its schema"
                 )
 
-            # _shape seeds ctx.shape so _make_default_array knows the requested size.
+            # _shape seeds ctx.shape so _make_default_array knows the requested size
             setattr(self, primary_array_name, self.get_default(primary_array_name))
-            self._shape = None
 
         # initialize arrays from keyword arguments when they are present
         for attr, value in kwargs.items():
--- a/src/stdatamodels/model_base.py
+++ b/src/stdatamodels/model_base.py
@@ -462,6 +462,7 @@ class DataModel(properties.ObjectNode):
                 file_reference.increment()
                 target._file_references.append(file_reference)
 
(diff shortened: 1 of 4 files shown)


src/stdatamodels/properties.py
```diff
@@ -211,8 +211,8 @@ def _make_default_array(attr, schema, ctx):
     primary_array_name = ctx.get_primary_array_name()
 
     if attr == primary_array_name:
-        if ctx.shape is not None:
-            shape = ctx.shape
+        if ctx._shape is not None:
```
melanieclarke: Why is this changed?  I would think it should use `ctx.shape`, to make sure `_shape` is updated if needed?

emolter: This piece of code is run when a datamodel is initialized from a tuple, e.g. `model = ImageModel((10,10))`.  Without changing this, the changes to the `shape` property meant that during init, the `self.hasattr(primary_array_name)` would return False, and the `shape` property would return `None`.  That meant here that the shape seen here was `None` which led to an empty array.

However you are right that there's a bug here:
```
    with DefaultsModel((10, 10)) as im:
        im.primary = np.zeros((9,9))
        default_primary = im.get_default("primary")
        assert default_primary.shape == (9, 9)
```
raises an AssertionError, whereas
```
    with DefaultsModel((10, 10)) as im:
        im.primary = np.zeros((9,9))
        default_primary = im.get_default("primary")
        im.shape
        assert default_primary.shape == (9, 9)
```
is fine.  I will rethink this.

emolter: The last few commits should fix this, sorry for the messy commit history

src/stdatamodels/model_base.py
```diff
@@ -674,11 +675,10 @@ def to_fits(self, init, *args, **kwargs):
...
-        if self._shape is None:
-            primary_array_name = self.get_primary_array_name()
-            if primary_array_name and self.hasattr(primary_array_name):
-                primary_array = getattr(self, primary_array_name)
-                self._shape = primary_array.shape
+        primary_array_name = self.get_primary_array_name()
+        if primary_array_name and self.hasattr(primary_array_name):
+            primary_array = getattr(self, primary_array_name)
+            return primary_array.shape
         return self._shape
```
melanieclarke: To check my understanding -- since this doesn't update self._shape, the intent is now to leave the internal value set to the input shape?  

So, if I set up a model with an initial shape, it should remember, like this?
```
dm = ImageModel((10,10))
print(dm.shape) # is (10,10)
dm.data = np.zeros((5,5))
print(dm.shape)  # is (5,5)
dm.data = None
print(dm.shape) # is (10,10), not None
```

I think that seems reasonable, since I set up the model with an initial value to begin with.

This particular example raises an AttributeError, though, on the last line:
```
File src/stdatamodels/properties.py:396, in ObjectNode.__getattr__(self, attr)
    394 except KeyError as err:
    395     if schema == {}:
--> 396         raise AttributeError(f"No attribute '{attr}'") from err
    398     if "max_ndim" in schema or "ndim" in schema or "datatype" in schema:
    399         # data-like attributes should return None if not set
    400         # not AttributeError, because we want hasattr(model, 'data') to be True
    401         # but also don't want to set the array in getattr
    402         return None

AttributeError: No attribute 'shape'
```

emolter: Thanks, yes the intent is that `_shape` is the shape that the model was initialized with.  I have added a test that makes that explicit.  The `AttributeError` bug you found, and thanks for catching it, should be fixed now - I had to catch explicit None as well.

melanieclarke: Yep, that's working well now, thanks!

melanieclarke (commented):
This looks good to me now, pending more reviews and clean tests.

tests/test_models.py
```diff
@@ -40,6 +40,52 @@ def test_set_shape():
...
+    with BasicModel((50, 50)) as dm:
+        # deleting primary should reset shape to init value
+        del dm.data
+        assert dm.shape == (50, 50)
+
+        # making primary None should reset shape to init value
+        # this happens even if primary array is set in the meantime
+        dm.data = np.zeros((10, 10))
+        dm.data = None
+        assert dm.shape == (50, 50)
```
braingram: Isn't this testing a bug?
The docstring for shape is "Return the shape of the primary array." So shouldn't dm.shape be None here?

emolter: I can see that argument.  I'll see what can be done about it

emolter: https://github.com/spacetelescope/stdatamodels/pull/740/commits/abf9e64824b57f26d80595037b0e2ab74530d9c2 addresses this, going to resolve since we talked about it in-person

tests/test_models.py
```diff
@@ -40,6 +40,60 @@ def test_set_shape():
...
+
+        dm.data = np.zeros((42, 23))
+        assert dm.shape == (42, 23)
+
+        # modifying non-primary does not change shape
+        dm.area = np.zeros((10, 10))
+        assert dm.shape == (42, 23)
+
+        # since BasicModel was init from nothing, shape reverts to None
```
braingram:
```suggestion
        # primary array deleted, shape reverts to None
```
Nitpick on comment.

braingram (approved):
Thanks!

tapastro (approved):

PR #741, opened 2026-05-12
Raise AttributeError when attempting to set ObjectNode.instance 

### LINKED ISSUE #704: setting `DataModel.instance` (or `Node.instance`) results in unexpected contents
`DataModel` inherits an `instance` property from `Node` that exposes the underlying `_instance` dict (for a `DataModel`):
https://github.com/spacetelescope/stdatamodels/blob/d73ecd971142c1c6694639a982592690fcbb7837/src/stdatamodels/properties.py#L365-L367

`Node.instance` does not have a setter but `ObjectNode` [contains a custom `__setattr__`](https://github.com/spacetelescope/stdatamodels/blob/d73ecd971142c1c6694639a982592690fcbb7837/src/stdatamodels/properties.py#L412) which makes setting `instance` possible.

```python
m = dm.ImageModel()
m.instance = {"meta": {"foo": "bar"}}
assert "foo" not in m.meta  # no foo in meta
assert "foo" not in m.instance["meta"]

assert "instance" in m._instance  # m._instance now has an "instance" key

m.instance = m.instance  # creates a new "instance" key in _instance with the contents of _instance (recursive)
```

As discussed with @emolter there are at least 2 ways to address this.

1) Make setting `instance` an error
2) Allow setting `instance` directly and have it set `_instance` (not sure if we ever want to delete `_instance` or throw an exception)

emolter:
Closes #704 

This PR causes attempted setting of ObjectNode (and therefore datamodel) instance attribute to raise an AttributeError.

## Tasks

</details>


braingram (approved):
Thanks!

PR #743, opened 2026-05-15
JP-3944: New ref file type for NIRSpec chromaticity correction

emolter:
Relates to [JP-3944](https://jira.stsci.edu/browse/JP-3944)

See companion JWST PR here: https://github.com/spacetelescope/jwst/pull/10556 

This PR adds a new reference file type in support of the NIRSpec IFU chromaticity correction.

## Tasks

</details>

### DIFF
--- a/changes/743.feature.rst
+++ b/changes/743.feature.rst
@@ -1,0 +1,1 @@
+Add ChromCorrModel type to support NIRSpec IFU chromaticity correction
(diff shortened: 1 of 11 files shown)


melanieclarke (commented):
LGTM, but would be helpful to have another datamodels expert review.

@braingram - this is still in draft because it needs CRDS support before we can move forward, but an advance review would be helpful if you have time.

braingram (approved):
Thanks. Changes look complete to me.

I pulled down one of the example files from the JP ticket and it has a defined lamp_mode of N/A but the rmap for https://jira.stsci.edu/browse/CRDS-1113 doesn't list that as part of the selector so likely if it is an "issue" that lamp_mode exists it's one with the reference file and not with the datamodel. My initial reaction is that's it's inconsequential but I figured I'd mention it here in case I'm mistaken.

melanieclarke (approved):
Let's go ahead and merge this since it will have no impact with jwst on main, and is working with the files available on CRDS test.  One less thing to remember to do later when CRDS test merges with ops...

Issue #405, opened 2025-02-18
Refactor wcs_ref_models.py to reduce repeated methods


emolter:
All of the models defined in `wcs_ref_models` need identical `on_save` and `to_fits` as well as much of the same functionality in their respective `validate()` methods.  There is already a base class named `_SimpleModel` that contains these methods.  However, there are 11 different models that do not inherit from `_SimpleModel`, so lots of repeated code remains.

It would be useful to refactor this (using, e.g., a mixin pattern) so that all the models can benefit from the implementations in the base class or classes.

PR #744, opened 2026-05-15
Fix to_fits signature to correctly raise NotImplementedError

karlhillx:
Closes #742

**Description**

When `save()` calls `to_fits(output_path, *args, **kwargs)` from the base model, any subclasses that override `to_fits` without accepting `*args, **kwargs` will throw a `TypeError` (e.g., unexpected keyword argument 'overwrite') before the intended `NotImplementedError` can be raised.

This PR simply updates the signature of `to_fits` in the affected reference models (in `wcs_ref_models.py` and `tsophot.py`) to accept `*args` and `**kwargs`, allowing the `NotImplementedError` to surface correctly with the intended message.

### DIFF
--- a/tests/jwst/datamodels/test_wcs.py
+++ b/tests/jwst/datamodels/test_wcs.py
@@ -44,9 +44,10 @@ def test_simple_model_to_fits_not_implemented():
     not TypeError (which was the bug before the *args, **kwargs signature fix).
     See https://github.com/spacetelescope/stdatamodels/pull/744
     """
-    import tempfile
     import os
+    import tempfile
     import warnings
+
     from stdatamodels.jwst.datamodels import DistortionModel
 
     with warnings.catch_warnings():
--- a/tests/jwst/datamodels/test_wcs.py
+++ b/tests/jwst/datamodels/test_wcs.py
@@ -36,3 +36,36 @@ def test_wcs_ref_models():
         fo.meta.instrument.channel = "SHORT"
         fo.meta.instrument.module = "A"
         fo.validate()
+
+
+def test_simple_model_to_fits_not_implemented():
+    """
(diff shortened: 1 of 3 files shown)


emolter (commented):
Thanks @karlhillx this looks good.

A unit test would be helpful.  I know it's redundant that all of these methods have the same override (and there's an [open issue](https://github.com/spacetelescope/stdatamodels/issues/405) about how to clean up this part of the code),  so I'm not suggesting unit-testing all 13 instances, although with a parametrized test it may not be that hard.  However I think it would be good to directly test this for at least one of the models that inherits from `_SimpleModel`.

My suggestion would be to put it here: https://github.com/spacetelescope/stdatamodels/blob/main/tests/jwst/datamodels/test_wcs.py

karlhillx:
Added a test using DistortionModel covering the full call path through save() with a .fits path, which is what triggers the overwrite=True kwarg. Both tests in test_wcs.py pass. Ready for re-review.

tests/jwst/datamodels/test_wcs.py
```diff
@@ -36,3 +36,36 @@ def test_wcs_ref_models():
         fo.meta.instrument.channel = "SHORT"
         fo.meta.instrument.module = "A"
         fo.validate()
+
+
+def test_simple_model_to_fits_not_implemented():
```
emolter, line 41: the structure of the test looks good overall, and thanks for adding it.  A few small things:

- we have a fixture `tmp_path` that will set up a temporary directory for the test without needing to import tempfile and unlink at the end. It would be helpful here, and you'd be able to avoid the try...finally
- I don't know what warnings you were attempting to ignore here, but we avoid catching and ignoring warnings in tests, as they may hide a real issue.  Warnings are upgraded to errors in our testing CI job.
- I don't believe it's necessary for this test to explicitly call `validate()`

PR #746, opened 2026-05-21
JP-3959: Add the WCS tag to datamodels schemas

### LINKED ISSUE #7020: `wcs` is missing from tab-completion of `meta`

emolter:
Resolves [JP-3959](https://jira.stsci.edu/browse/JP-3959)

Fixes https://github.com/spacetelescope/jwst/issues/7020

This PR addresses an oversight in the schemas where the WCS tag is missing from the schemas and never validated.

The side effects of this change are:
- TAB-expansion now shows the `wcs` attribute.
- `hasattr(model.meta, "wcs")` now returns `True` if the wcs attribute is schema-defined but not present, consistent with the behavior of other attributes.  Prior to this PR it would return False in that case, which was a unique behavior of `model.meta.wcs` specifically.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -13,7 +13,7 @@ from asdf import schema as asdf_schema
 from asdf import tagged, treeutil
 from asdf.exceptions import ValidationError
 from asdf.tags.core import HistoryEntry, NDArrayType, ndarray
-from asdf.util import HashableDict
+from asdf.util import HashableDict, uri_match
 from astropy import time
 from astropy.io import fits
 from astropy.utils.exceptions import AstropyWarning
@@ -376,17 +376,32 @@ def _get_validators(hdulist):
     fits_context = FitsContext(hdulist)
 
     validators = HashableDict(asdf_schema.YAML_VALIDATORS)
-    tag_validator = validators["tag"]
 
-    def validate_tag(validator, tag_pattern, instance, schema):
-        af = asdf.AsdfFile()
-        if af.extension_manager.handles_type(instance.__class__):
-            for tag in af.extension_manager.get_converter_for_type(instance.__class__).tags:
-                tag_validator(validator, tag_pattern, tag, schema)
+    # create an extension_manager for validate_tag
+    extension_manager = asdf.AsdfFile().extension_manager
+
+    def validate_tag(validator, tag_pattern, instance, schema, extension_manager=extension_manager):
+        if not extension_manager.handles_type(instance.__class__):
+            yield ValidationError(
+                f"Unsupported {instance} does not have a tag matching pattern {tag_pattern}"
+            )
             return
-        raise ValidationError(
-            f"Unsupported {instance} does not have a tag matching pattern {tag_pattern}"
-        )
+
+        # we don't know yet which of these tags will be used so check them all
+        tags = extension_manager.get_converter_for_type(instance.__class__).tags
+
+        # did we find tags?
+        if not tags:
+            yield ValidationError(
+                f"type '{instance.__class__}' has no valid tags to check against '{tag_pattern}'"
+            )
+            return
+
+        # select_tag requires a SerializationContext which we don't have so check all tags
+        match = all(uri_match(tag_pattern, tag) for tag in tags)
+        if not match:
+            yield ValidationError(f"tags '{tags}' do not match pattern '{tag_pattern}'")
+            return
 
     partial_fits_array_writer = partial(_fits_array_writer, fits_context)
 
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -11,6 +11,7 @@ import asdf
 import numpy as np
 from asdf import schema as asdf_schema
 from asdf import tagged, treeutil
+from asdf.exceptions import ValidationError
 from asdf.tags.core import HistoryEntry, NDArrayType, ndarray
 from asdf.util import HashableDict
 from astropy import time
@@ -375,6 +376,17 @@ def _get_validators(hdulist):
     fits_context = FitsContext(hdulist)
 
     validators = HashableDict(asdf_schema.YAML_VALIDATORS)
+    tag_validator = validators["tag"]
+
+    def validate_tag(validator, tag_pattern, instance, schema):
+        af = asdf.AsdfFile()
+        if af.extension_manager.handles_type(instance.__class__):
+            for tag in af.extension_manager.get_converter_for_type(instance.__class__).tags:
+                tag_validator(validator, tag_pattern, tag, schema)
+            return
+        raise ValidationError(
+            f"Unsupported {instance} does not have a tag matching pattern {tag_pattern}"
+        )
 
     partial_fits_array_writer = partial(_fits_array_writer, fits_context)
 
@@ -387,6 +399,7 @@ def _get_validators(hdulist):
             "items": partial(_fits_item_recurse, fits_context),
             "properties": partial(_fits_comment_section_handler, fits_context),
             "type": partial(_fits_type, fits_context),
+            "tag": validate_tag,
         }
     )
 
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -11,6 +11,7 @@ import asdf
 import numpy as np
 from asdf import schema as asdf_schema
 from asdf import tagged, treeutil
+from asdf.exceptions import ValidationError
 from asdf.tags.core import HistoryEntry, NDArrayType, ndarray
 from asdf.util import HashableDict
 from astropy import time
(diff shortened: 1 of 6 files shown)


melanieclarke (commented):
LGTM, but will leave to @tapastro to approve, since this needs communication to other teams.

src/stdatamodels/fits_support.py
```diff
@@ -373,6 +374,17 @@ def _get_validators(hdulist):
...
+    tag_validator = validators["tag"]
+
+    def validate_tag(validator, tag_pattern, instance, schema):
+        af = asdf.AsdfFile()
+        if af.extension_manager.handles_type(instance.__class__):
+            for tag in af.extension_manager.get_converter_for_type(instance.__class__).tags:
+                tag_validator(validator, tag_pattern, tag, schema)
+            return
+        raise ValidationError(
+            f"Unsupported {instance} does not have a tag matching pattern {tag_pattern}"
```
braingram: Looking at this with fresh eyes it's currently not working as expected. The validators yield errors so the raise need to be a yield and we need a yield from for the tag_validator. That exposes that the use of tag_validator here doesn't work as needed since when validation is used for mapping fits hdus the passed object is a `WCS` and not a tagged object.

There is an issue of scope of the `AsdfFile` used here where there is one created in `_create_asdf_hdu` which should ideally be the same one used here but that would require a more major refactoring. Since for this PR all that's really needed is to figure out what tag asdf will use for the WCS object I think we can:
- create a throw-away AsdfFile instance attached to the fits_context (so we don't make a new one for every tag)
- use that to look up the tag (I think we can use select_tag here)
- do the uri_match ourselves (we could also make a throw-away tagged object but that seems more brittle depending on asdf internals)

Since the suggestions touch a few things in this file I'll make a PR targeting yours.

emolter: thanks for finding this.  I'd appreciate the PR very much - I have to admit I haven't fully understood what's wrong here, and it was originally Nadia's work.

braingram: Happy to help. It looks like we also want some tests for this where invalid wcses are provided for assignment and save (which I think would have revealed the issue). I'll add that to the PR.

src/stdatamodels/jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2944,3 +2944,8 @@ properties:
             type: number
             fits_keyword: BKGD_SCL
             blend_table: True
+      wcs:
+        title: WCS object
+        anyOf:
+          - tag: tag:stsci.edu:gwcs/wcs-*
+          - type: "null"
```
braingram: Is this `null` needed? I'm running into an odd issue where if we get the `tag` validator to fail the `type` validator fails due in part to the override in this package. Providing a ndarray as a wcs produces an error:
```python
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
on save.

emolter: You're the expert here, along with maybe @WilliamJamieson , but I have to say I'd agree at least that the specific error message is not very illuminating

tests/jwst/datamodels/test_validation.py
```diff
@@ -28,3 +32,57 @@ def test_strict_validation_date():
...
+    ],
+)
+@pytest.mark.parametrize("strict", [True, False])
+@pytest.mark.parametrize("ext", ["fits", "asdf"])
+def test_wcs_validation_on_save(tmp_path, ext, value, error, strict):
+    fn = tmp_path / f"test.{ext}"
+    if error:
+        ctx = pytest.raises(ValidationError)
+        if ext == "asdf":
+            pytest.xfail(
```
braingram, line 80: These are xfailed because of inconsistent validation on save for asdf and fits: https://github.com/spacetelescope/stdatamodels/issues/750

tests/jwst/datamodels/test_read_metadata.py
```diff
@@ -19,9 +20,7 @@
 @pytest.fixture
 def recursive_tree():
     """Create a recursive tree to substitute for a WCS object"""
-    w = {"inputs": "test"}
-    w["outputs"] = w
-    return w
+    return examples.gwcs_2d_shift_scale()
```
braingram, line 23: This no longer produces a recursive tree. Is it necessary to update this?

emolter: looks like all tests pass with this reverted.  Do you want me to revert it? or were you just making sure it wasn't hiding some bug?

braingram: If I recall correctly this was added to cover a bug where read_metadata failed to read a recursive tree (due to asdf.util.load_yaml failing for a recursive tree). asdf 4.2.0+ doesn't have the bug and I believe we worked around this by passing tagged for asdf files:
https://github.com/spacetelescope/stdatamodels/blob/433b2d003146e021a627649e70b3b6bae8b7b9c3/src/stdatamodels/jwst/datamodels/util.py#L535

It seems beneficial to keep a recursive object as part of these tests until we set the minimum asdf version to 4.2.0. An alternative would be:
- bump the min asdf to 4.2.0
- rename this fixture to wcs (or some other more descriptive name)

What do you think?

emolter: reverted in https://github.com/spacetelescope/stdatamodels/pull/746/commits/8a62fe27bcbd7c5674030efec19a600fa7928c79

braingram: Thanks. Unfortunately it's failing in the CI. Maybe we go with the alternative? 4.2.0 is over a year old at this point and already the min version in jwst.

emolter: whoops yeah I didn't notice the warnings, for some reason they aren't upgraded to errors for me locally.  And it's funny we happened to look at this again at exactly the same moment.

I think we should just go ahead with the unit test update Nadia made and move forward with this PR.

Then, separately, it may make sense to bump asdf to 4.2

braingram: Sounds good. Would you rename the fixture in this PR to something like `example_wcs`?

emolter: done

tests/jwst/datamodels/test_read_metadata.py
```diff
@@ -17,20 +18,18 @@
 
 
 @pytest.fixture
-def recursive_tree():
+def example_wcs():
     """Create a recursive tree to substitute for a WCS object"""
```
braingram, line 22:
```suggestion
    """Create an example WCS object"""
```

braingram (approved):
Thanks!

tests/jwst/datamodels/test_read_metadata.py
braingram: Nitpick here on the docstring for this fixture.

tapastro (approved):
Smoke signals sent to jdaviz and JP notebooks folks about possible breaking changes due to difference in hasattr behavior; I'm happy to merge this now.

PR #748, opened 2026-05-27
Fix FITS bug for list-type metadata

melanieclarke:
Toward [JP-4313](https://jira.stsci.edu/browse/JP-4313)

To produce accurate rate values, some multistripe modes will require explicit read times for all readout frames in each group.  See #739 for the draft implementation to support this, which stores the read time values in a list of lists inside `meta.exposure.read_times`.  This is similar to the `meta.exposure.read_pattern` metadata stored for Roman datamodels: https://github.com/spacetelescope/rad/blob/b7879d2c36bdac7006a57c9971b8ae8dbb157ee4/src/rad/resources/schemas/meta/exposure-1.2.0.yaml#L197

Using this draft implementation, I noticed that RampModels saved as FITS files with `meta.exposure.read_times` set generated many extra FITS extensions, mostly empty.  I tracked it down to the `_fits_item_recurse` validator, which sets a sequence index whenever it encounters list-type attributes.  This is appropriate for lists of objects, like slits or spectra, since this is how it knows to start a new extension version for a new slit or spectrum.  It is not appropriate for list-type primary metadata: when encountered, the update to the sequence index makes any metadata defined later in the schema think it's supposed to go into a new primary header extension.

The fix here is to update the sequence index only if the list-type item contains an object.   This works for the use case I need it for, and shouldn't break any other existing JWST schemas, but it may be fragile to other kinds of stored metadata.  I'm open to alternate fixes if anyone can think of a more general one, or a better way to store the necessary metadata in #739.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -338,15 +338,12 @@ def _fits_item_recurse(fits_context, validator, items, instance, schema):
         return
 
     if validator.is_type(items, "object"):
+        initial_index = fits_context.sequence_index
         for index, item in enumerate(instance):
-            # Update the FITS sequence index if it is a list of objects,
-            # to support things like lists of slits or spectra. No need
-            # if it is just a simple list of items.
-            if validator.is_type(item, "object"):
-                fits_context.sequence_index = index
-
+            fits_context.sequence_index = index
             for error in validator.descend(item, items, path=index):
                 yield error
+        fits_context.sequence_index = initial_index
     else:
         # We don't do the index trick on "tuple validated" sequences
         for (index, item), subschema in zip(enumerate(instance), items, strict=False):
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -339,7 +339,12 @@ def _fits_item_recurse(fits_context, validator, items, instance, schema):
 
     if validator.is_type(items, "object"):
         for index, item in enumerate(instance):
-            fits_context.sequence_index = index
+            # Update the FITS sequence index if it is a list of objects,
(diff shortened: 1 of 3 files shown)


src/stdatamodels/fits_support.py
```diff
@@ -339,7 +339,12 @@ def _fits_item_recurse(fits_context, validator, items, instance, schema):
 
     if validator.is_type(items, "object"):
         for index, item in enumerate(instance):
-            fits_context.sequence_index = index
+            # Update the FITS sequence index if it is a list of objects,
+            # to support things like lists of slits or spectra. No need
+            # if it is just a simple list of items.
+            if validator.is_type(item, "object"):
```
braingram: Rather than rely on item type, can we store the index outside of the for loop then restore it after?

braingram: I am failing to generate a reasonable suggestion through github :-/ something like:
```python
        initial_index = fits_context.sequence_index
        for index, item in enumerate(instance):
            fits_context.sequence_index = index
            for error in validator.descend(item, items, path=index):
                yield error
        fits_context.sequence_index = initial_index
```
Testing it locally with the test added in this PR (thanks for adding that!) the tests pass but I did not run jwst tests or examine the expected output.

melanieclarke: Clever!  Yes, it looks like that works for both the simple list metadata and the data arrays that need extension versions to be recorded properly.

melanieclarke: Fixed with your suggestion.  I'll rerun regression tests.

braingram (approved):
Thanks for finding and fixing the bug!

PR #756, opened 2026-06-10
JP-4326: New model type for WFSS multi-band direct imaging data

emolter:
Relates to [JP-4326](https://jira.stsci.edu/browse/JP-4326)

This PR adds a new model type to hold multiple WFSS direct images at different wavelengths inside a single datamodel.

The model is being added in part in response to comments from testers of JP-4326 that re-using `IFUCubeModel` for this purpose is confusing.
The new `WFSSCubeModel` looks broadly similar to `IFUCubeModel` but has a pared-down set of metadata and extensions.

## Tasks

</details>

### DIFF
--- a/src/stdatamodels/jwst/datamodels/schemas/wfssmultiband.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/wfssmultiband.schema.yaml
@@ -27,10 +27,8 @@ allOf:
       default: 0.0
       ndim: 3
       datatype: float32
-    wavetable:
-      title: Wavelength value for slices
-      fits_hdu: WCS-TABLE
-      datatype:
-      - name: wavelength
-        datatype: float32
-        ndim: 2
+    wavelength:
+      title: Wavelength value for planes
+      fits_hdu: WAVELENGTH
+      ndim: 1
+      datatype: float32
--- a/src/stdatamodels/jwst/datamodels/schemas/wfssmultiband.schema.yaml
+++ b/src/stdatamodels/jwst/datamodels/schemas/wfssmultiband.schema.yaml
@@ -1,0 +1,36 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.1.0"
+id: "http://stsci.edu/schemas/jwst_datamodel/wfssmultiband.schema"
+allOf:
+- $ref: core.schema
+- $ref: bunit.schema
+- $ref: photometry.schema
+- $ref: wcsinfo.schema
+- type: object
+  properties:
+    data:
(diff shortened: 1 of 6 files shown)


src/stdatamodels/jwst/datamodels/schemas/wfssmultiband.schema.yaml
```diff
@@ -0,0 +1,36 @@
...
+      title: Error array
+      fits_hdu: ERR
+      default: 0.0
+      ndim: 3
+      datatype: float32
+    wavetable:
+      title: Wavelength value for slices
+      fits_hdu: WCS-TABLE
+      datatype:
+      - name: wavelength
```
tapastro: I'm not well-versed in IFUCubes, but I took a look at some I had on disk to understand this extension. My local examples all have an (N, 1) shaped wavetable, where the data shape is (N, x, y). Is there a reason the wavetable is 2D and not 1D?

emolter: I noticed this too.  I think it should be fine, and less confusing, to make it 1-D.  Let me try it

melanieclarke: There are FITS spectral WCS requirements on the shape of this table for IFU data.

emolter: ok thanks that's helpful context. Am I understanding correctly that this isn't relevant for this new type?

melanieclarke: I haven't been following the discussion on why this new type is needed.  This table is intended for providing a spectral WCS in FITS data, for cases where the wavelength can't easily be expressed with standard FITS WCS keywords (e.g. it's a nonlinear function with different values for each slice along the wavelength dimension).  If that's not a thing you need to provide for this data type, then you shouldn't need the table.

emolter: @tapastro how about https://github.com/spacetelescope/stdatamodels/pull/756/commits/d3b652cf00cc9853f9edc63082dfb82443c96d53 which just makes the wavelengths an array.  I think this is the most user-friendly way to define it.

tapastro: It seems reasonable to me - and I think it should still be able to support any variety of wavelength sampling. Perhaps there is a WCS limitation I'm not wrapping my head around, but we can make updates if we run into a future snag.

tapastro (approved):

PR #640, opened 2025-12-18
DOC: Remove unrendered ref that looks ugly

pllim:
This PR addresses ugly rendered doc because the ref linking failed for some reason when docstring inherited.

Example: https://stdatamodels.readthedocs.io/en/latest/api/stdatamodels.jwst.datamodels.NRMModel.html

Screenshot 2025-12-18 154826

Also fixed a formatting error for a code snippet. No need for RT.

With this patch: https://stdatamodels--640.org.readthedocs.build/en/640/api/stdatamodels.jwst.datamodels.NRMModel.html

## Tasks
- [ ] update or add relevant tests
- [x] update relevant docstrings and / or `docs/` page
- [x] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
  - [ ] [run `jwst` regression tests](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) with this branch installed (`"git+https://github.com/<fork>/stdatamodels@<branch>"`)

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details>

PR #772, opened 2026-06-30
DOC: Fix default role for interpreted text

pllim:
Resolves [JP-4210](https://jira.stsci.edu/browse/JP-4210)

This PR addresses https://github.com/spacetelescope/jwst/issues/10093 so we can revert https://github.com/spacetelescope/stdatamodels/pull/640 .

Turns out the default role for interpreted text for this package was overwritten somewhere. This is probably unintentional side effect of something else because I do not see it explicitly set in this package, and things work "out of the box" downstream in `jwst` without explicitly resetting default role like this.

*  Pro: Interpreted text works again.
* Con: Because it has been broken for so long, other violations crept in and must be also fixed in this PR.

Only docs, so no need for RT.

Please just ignore the change log check failure because I cannot find a way to skip it. This PR does not need a change log but I have to fix old entries.

https://stdatamodels--772.org.readthedocs.build/en/772/

## Tasks
- [x] Fix the other 86 warnings 😭 

</details>


docs/source/api.rst
```diff
@@ -34,3 +34,9 @@ Model base class
    :no-inheritance-diagram:
 .. automodapi:: stdatamodels.jwst.transforms
    :no-inheritance-diagram:
+
+Utilities
+---------
```
emolter:
```suggestion
```
Not sure this is a helpful header, as there's already a header for stdatamodels.history, and also HistoryList is the only thing in the section

src/stdatamodels/jwst/datamodels/multicombinedspec.py
```diff
@@ -47,10 +49,11 @@ class WFSSMultiCombinedSpecModel(JwstDataModel):
     """
     A data model for NIRCam and NIRISS WFSS exposure-averaged 1D spectra.
 
-    This model differs from the other MultiCombinedSpecModel classes in that
+    This model differs from the other
+    `~stdatamodels.jwst.datamodels.MultiCombinedSpecModel` classes in that
```
emolter:
```suggestion
    This model differs from the
    `~stdatamodels.jwst.datamodels.MultiCombinedSpecModel` class in that
```

CHANGES.rst
```diff
@@ -349,10 +349,10 @@ Deprecations and Removals
 - Deprecate passing kwargs into asdf.open from asdf_in_fits.open (`#533
   <https://github.com/spacetelescope/stdatamodels/issues/533>`_)
 - Deprecate V23ToSky transform. Use the function
-  `jwst.assign_wcs.pointing.v23tosky` to construct an equivalent
-  `CompoundModel` that relies only on models from `astropy.modeling`.
+  ``jwst.assign_wcs.pointing.v23tosky`` to construct an equivalent
+  ``CompoundModel`` that relies only on models from `astropy.modeling`.
```
emolter, line 353: should this go to https://docs.astropy.org/en/stable/api/astropy.modeling.CompoundModel.html ?

emolter (commented):
This looks like a very helpful cleanup!  Just a few small comments.

CHANGES.rst
pllim: IMHO we should not have put live links in change log in the first place. This is to avoid having to change historical entries (as I am doing here) in the future when things break. So I would say we keep the double backticks.

docs/source/api.rst
pllim: Removed.

emolter (approved):

PR #779, opened 2026-07-23
JP-4395: Update sampling for MIRI WFSS transform

drlaw1558:
Partially resolves [JP-4395](https://jira.stsci.edu/browse/JP-4395)

## Tasks

- `<PR#>.feature.rst`
- `<PR#>.bugfix.rst`
- `<PR#>.docs.rst`
- `<PR#>.other.rst`
</details>

### DIFF
--- a/changes/779.feature.rst
+++ b/changes/779.feature.rst
@@ -1,1 +1,1 @@
-Update number of sampling points for MIRI WFSS dy to t trasnform.
+Update number of sampling points for MIRI WFSS dy to t transform.
--- a/changes/779.feature.rst
+++ b/changes/779.feature.rst
@@ -1,0 +1,1 @@
+Update number of sampling points for MIRI WFSS dy to t trasnform.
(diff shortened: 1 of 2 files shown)


emolter:
Was finally able to get downstream tests running, and the unit test `test_NIRISSForwardRowGrismDispersion` in `jwst.assign_wcs.tests` is failing because it has some hard-coded expected values.

The maximum absolute wavelength difference between this PR and the old expectation is 0.08500112, which is bigger than your quoted 1 part in 10^6.

On the other hand, for some reason the wavelengths in this test are large negative values near -41.  So I think the test should be updated and moved into stdatamodels or else removed.  I'll look into what's best.

The regression test diffs for NIRISS with real data look to be in-line with your quoted differences: ~1e-4 is the magnitude of the relative flux differences.

emolter (approved):

tapastro (approved):

PR #787, opened 2026-08-10
Fix jwst downstream by configuring tox/pytest use jwst pytest configuration

braingram:
Update the jwst tox environment to use the jwst pytest configuration (to fix the current failures due to a warning ignored in the jwst configuration not being ignored here and instead turning into an error).

This also drops the coverage measurement of stdatamodels code during jwst tests. It seems reasonable to not include downstream tests as part of coverage measurement and this only drops coverage by 4%. I will update the branch protection rules after merging.

## Tasks

- `<PR#>.feature.rst`
- `<PR#>.bugfix.rst`
- `<PR#>.docs.rst`
- `<PR#>.other.rst`
</details>

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,8 @@ envlist =
 
 [testenv]
 allowlist_externals =
-    jwst: touch
+    jwst: bash
+    jwst: git
 description =
     run tests
     xdist: using parallel processing
@@ -23,19 +24,22 @@ extras =
 deps =
     xdist: pytest-xdist
     cov: pytest-cov
-    jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
 package=
     cov: editable
     !cov: wheel
+change_dir =
+    jwst: {env_tmp_dir}
 commands_pre =
-    devdeps: pip install -r requirements-dev.txt -U --upgrade-strategy eager
+    jwst: bash -c "pip freeze -q | grep 'stdatamodels @' > {env_tmp_dir}/requirements.txt"
+    jwst: git clone https://github.com/spacetelescope/jwst.git
+    jwst: pip install -e jwst[test]
+    devdeps: pip install -r {toxinidir}/requirements-dev.txt -U --upgrade-strategy eager
+    jwst: pip install -r {env_tmp_dir}/requirements.txt
     pip freeze
-# make an empty pytest.ini so jwst tests don't use the settings in pyproject.toml
-    jwst: touch pytest.ini
 commands =
     pytest \
     xdist: -n auto \
     cov: --cov=src --cov-report=term-missing --cov-report=xml \
-    jwst: --pyargs jwst --ignore-glob=*/scripts/* -Werror \
+    jwst: jwst \
     nocrds: --no-crds
     {posargs}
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -23,19 +24,20 @@ extras =
 deps =
     xdist: pytest-xdist
     cov: pytest-cov
-    jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
 package=
     cov: editable
     !cov: wheel
 commands_pre =
+    jwst: bash -c "pip freeze -q | grep 'stdatamodels @' > {env_tmp_dir}/requirements.txt"
+    jwst: git clone https://github.com/spacetelescope/jwst.git
```
zacharyburnett, line 34: just FYI this will clone a copy of `jwst/` within the `stdatamodels/` working copy, which won't conflict _right now_ but might in the future. A more maintainable solution might be doing `change_dir:` to move to an empty dir outside of `stdatamodels/` and then installing `stdatamodels` and `jwst` from source instead of having to make a `requirements.txt`

braingram: Thanks. I pushed a change in https://github.com/spacetelescope/stdatamodels/pull/787/commits/85f7a5baf279bc1e4c82523795a655dac36269c2

which copies more of the tox.ini from stcal (where most of this code was taken from).

zacharyburnett (approved):
otherwise this seems to work so I'm also fine with merging it as is

tox.ini
braingram: Turns out that didn't work due to failing to find requirement-dev.txt (from this package). I pushed a new commit to try and get both to work together. If it's still failing I'll convert this back to draft until a solution can be found.

tox.ini
```diff
@@ -23,19 +24,22 @@ extras =
...
     !cov: wheel
+change_dir =
+    jwst: {env_tmp_dir}
 commands_pre =
-    devdeps: pip install -r requirements-dev.txt -U --upgrade-strategy eager
+    jwst: bash -c "pip freeze -q | grep 'stdatamodels @' > {env_tmp_dir}/requirements.txt"
+    jwst: git clone https://github.com/spacetelescope/jwst.git
+    jwst: pip install -e jwst[test]
+    devdeps: pip install -r {toxinidir}/requirements-dev.txt -U --upgrade-strategy eager
+    jwst: pip install -r {env_tmp_dir}/requirements.txt
```
zacharyburnett, line 37: does this work?
```suggestion
    jwst: git clone https://github.com/spacetelescope/jwst.git
    jwst: pip install -e ./jwst[test]
    jwst: pip install -e ./stdatamodels
    devdeps: pip install -r {toxinidir}/requirements-dev.txt -U --upgrade-strategy eager
```

braingram: Thanks. I did something similar in https://github.com/spacetelescope/stdatamodels/commit/03638a6b6deb9c7ecd22cb7fca45d752e2ae9d82 which appears to have worked. I reopened this for review.

braingram: Excellent. Now the coverage doesn't work...

braingram: I'm going to remove coverage from those tests. It seems more helpful to have the unit tests here contribute to coverage instead of the entire jwst test suite.

braingram: Looks like the jwst tests only contribute ~4% to coverage. Seems reasonable to change the coverage to not include downstream tests.

emolter (approved):

PR #793, opened 2026-08-14
JP-2518: Improve round-trip accuracy for NIRCam grism

melanieclarke:
Resolves [JP-2518](https://jira.stsci.edu/browse/JP-2518)

Closes #687 

Related but independent jwst PR: https://github.com/spacetelescope/jwst/pull/10783

The NIRCam backward grism dispersion transform currently uses a linear interpolation over 40 sampled points to invert the wavelength model to recover the x index.   This can lead to round trip errors that vary over the dispersion axis.  For one example (jw01366002001_04103_00001), round trip error at x0 = [100, 1000, 1500], y0 = 34 with 
```x, y, _ = wcs.backward_transform(*wcs(x0, y0, 1))```
gives x - x0 of [-0.00420179  0.01182850  0.02396645].

Similar to the changes for WFSS transform in #779, we can use a spline fit for a smoother inversion instead of the linear interpolation.  With these changes, the round trip error in x is smaller and more even:
`x - x0 = [0.00020418 0.00020306 0.00020273]`

Also, this transform currently does not correctly handle input 1D vectors: it should be possible to input paired x, y, lambda values in 1D vector form and get back a 1D vector.  This PR also updates output handling to support more general 1D and 2D cases, replacing draft PR #687.  For simplicity, the transform still requires that if any input array is 2D, the input arrays all match in shape.

## Tasks

- `<PR#>.feature.rst`
- `<PR#>.bugfix.rst`
- `<PR#>.docs.rst`
- `<PR#>.other.rst`
</details>

### DIFF
--- a/changes/793.feature.rst
+++ b/changes/793.feature.rst
@@ -1,0 +1,2 @@
+For the NIRCam grism backward dispersion model, update linear interpolation method to cubic spline for better roundtrip accuracy.
+Also, allow paired 1D vector inputs to the backward transform, returning a 1D vector output.
--- a/src/stdatamodels/jwst/transforms/models.py
+++ b/src/stdatamodels/jwst/transforms/models.py
@@ -1483,7 +1483,6 @@ class NIRCAMBackwardGrismDispersion(_BackwardGrismDispersionBase):
             The inverse dispersion value for the given wavelength
         """
         t0 = np.linspace(0.0, 1.0, int(self.sampling))
-
         if len(model) < 2:
             # Handle legacy versions of the trace model
             xr = _evaluate_transform_guess_form(model, x=x0, y=y0, t=t0)
@@ -1507,13 +1506,16 @@ class NIRCAMBackwardGrismDispersion(_BackwardGrismDispersionBase):
         if (x0.ndim > 1 or wavelength.ndim > 1) and x0.shape != wavelength.shape:
             raise ValueError("Inputs are >1D but array shapes do not match")
 
-        # Most common use case is a single unique x0, y0 value for all input,
-        # but it is technically possible to transform different x0, y0 values
-        # at the same time. Fit t as a function of wavelength
+        # It is possible to transform different x0, y0 values at the same time
+        # (e.g., for WFSS sources). Fit t as a function of wavelength
         # for each set of x0, y0 in the input and use it to get the output t value.
         x_y_pairs = np.unique(np.stack([x0.flat, y0.flat], axis=1), axis=0)
         t_out = np.full_like(wavelength * x0, np.nan)
         for xref, yref in x_y_pairs:
+            # If invalid, skip this pair: t_out will be NaN for this input.
+            if np.isnan(xref) or np.isnan(yref):
+                continue
+
             # Get w for each sampled t0
             w_t = trace_function(t0, np.full_like(t0, xref), np.full_like(t0, yref))
             idx = np.argsort(w_t)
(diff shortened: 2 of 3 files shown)


melanieclarke:
Regtests with jwst on main:
https://github.com/spacetelescope/RegressionTests/actions/runs/32157357489

emolter:
pinging @drlaw1558 because he was the one who originally reported the 1-D vector issue

emolter (changes requested):
This approach is adversely affecting runtime too much.  Using test data from `jw04043-o001_20251224t153004_spec2_00031_asn.json` with default options, it takes <1 second to do `assign_wcs.util.create_grism_bbox` on main and over 2 minutes on the PR.  I think the primary driver of this is that the trace function is now evaluated within a `for` loop over pixels, which basically undoes the vectorization.

Is the main reason the round-trip is more accurate due to the change to using spline interpolation?  If so, do we need all the other structural changes here, or would it make sense to make a drop-in replacement of `_find_min_with_linear_interpolation` that uses cubic spline interpolation in a vectorized way instead of linear interpolation?

I think it might also be helpful, given the complexity here, to separate the PRs into two - one that improves the roundtrip, and one that supports 1-D vector input.

melanieclarke:
Thanks for testing!  I'll rethink this with the WFSS case more in mind.  I originally developed it for TSGRISM, for which it is usually called with a single x0/y0 pair -- for this use case, the spline is generated once and the transform is vectorized over wavelengths.

I'll also see if it makes sense to separate out the 1D vector support.

melanieclarke:
After some investigation, I've found some faster ways to do the spline interpolation than my first attempt.  I pushed up a small change that makes the default spec2 case you tested run just a couple seconds longer than the implementation on main, but running wfss_contam is still prohibitively slow with this implementation (~hours instead of minutes).

I tried a few other spline interpolation options as well, including looking into ways to vectorize over x0, y0 like the current implementation, but I haven't found anything that comes close to current performance without major development effort.  

I also tried just increasing the sampling for the linear interpolation.  This won't fix the spiky, uneven nature of the inversion, but it does reduce the absolute roundtrip error, at the cost of more runtime.  Changing it from 40 to 100 decreases max roundtrip errors in my TSGRISM case from 0.03 to 0.006 but increases runtime for spec2 with wfss_contam from 10m to 16m for your sample case, on my computer.  I'm not sure if that level of improvement is worth the performance hit for wfss_contam.

The other thing I could do is separate out TSGRISM from WFSS implementation, since the spline interpolation is plenty fast for TSGRISM purposes, but I'm not sure that's worth the extra maintenance burden either.

In short, I don't see any satisfactory options at this point.  I'm going to close this PR and suggest instead that we consider looking into smoother inversion options later alongside other optimizations, perhaps as part of https://github.com/spacetelescope/jwst/issues/10820.

drlaw1558:
@melanieclarke Just to understand the issue a little better; is it something about the NIRCam WFSS WCS implementation that's making the spline interpolation adversely affect wfss_contam runtime, or is the spline implemented for NIRISS/MIRI in https://github.com/spacetelescope/stdatamodels/pull/779 similarly problematic?

melanieclarke:
The issue with the NIRCam implementation is that the wavelength model depends on the reference spatial position, so it's not just one spline model, it's a spline model at each input x,y the transform is evaluated at.  A single spline doesn't take too long to fit, but 10,000 at once gets a bit expensive.

PR #794, opened 2026-08-14
JP-4432: Allow extra_fits to use caching and clean up FITS writer a bit

emolter:
Closes #792 

This PR refactors the FITS writer.

## Tasks

- `<PR#>.feature.rst`
- `<PR#>.bugfix.rst`
- `<PR#>.docs.rst`
- `<PR#>.other.rst`
</details>

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -340,12 +340,10 @@ class FitsContext:
             self.hdu_cache = {("PRIMARY", 0): fits.PrimaryHDU()}
         else:
             self.hdu_cache = {}
-            nameless_idx = 0
-            for hdu in hdulist:
+            for i, hdu in enumerate(hdulist):
                 if hdu.name in [None, ""]:
-                    # give any nameless HDUs a ver to differentiate them
-                    nameless_idx += 1
-                    hdu.ver = nameless_idx
+                    # give nameless HDUs a ver equal to their index in the file
+                    hdu.ver = i
                 self.hdu_cache[(hdu.name, hdu.ver - 1)] = hdu
 
 
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -970,7 +970,7 @@ def _can_skip_fits_update(hdulist, asdf_struct, context):
 
     # Check for FITS hash and compare to current. If equal, automatically skip.
     if asdf_struct.tree.get(FITS_HASH_KEY, None) is not None:
-        if asdf_struct.tree[FITS_HASH_KEY] == fits_hash(list(hdulist)):
+        if asdf_struct.tree[FITS_HASH_KEY] == fits_hash(hdulist):
             log.debug("FITS hash matches. Skipping FITS updating.")
             return True
 
@@ -986,7 +986,7 @@ def fits_hash(hdus):
 
     Parameters
     ----------
-    hdus : list[HDU]
+    hdus : `astropy.io.fits.HDUList` or plain list of HDUs
         All the HDUs.
 
     Returns
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -886,14 +886,12 @@ def _map_hdulist_to_arrays(hdulist, af):
                 "((?P<name>[ -~]+),)?(?P<ver>[0-9]+)",
                 source[len(_FITS_SOURCE_PREFIX) :],
             )
-            if parts is not None:
-                ver = int(parts.group("ver"))
-                if parts.group("name"):
-                    pair = (parts.group("name"), ver)
-                else:
-                    pair = ("", ver)
-            data = hdu_index[pair].data
-            return data
+            ver = int(parts.group("ver"))
+            if parts.group("name"):
+                pair = (parts.group("name"), ver)
+            else:
+                pair = ("", ver)
+            return hdu_index[pair].data
         return node
 
     # don't assign to af.tree to avoid an extra validation
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -339,7 +339,14 @@ class FitsContext:
         if hdulist is None:
             self.hdu_cache = {("PRIMARY", 0): fits.PrimaryHDU()}
         else:
-            self.hdu_cache = {(hdu.name, hdu.ver - 1): hdu for hdu in hdulist}
+            self.hdu_cache = {}
+            nameless_idx = 0
+            for hdu in hdulist:
+                if hdu.name in [None, ""]:
+                    # give any nameless HDUs a ver to differentiate them
+                    nameless_idx += 1
+                    hdu.ver = nameless_idx
+                self.hdu_cache[(hdu.name, hdu.ver - 1)] = hdu
 
 
 def _get_validators(fits_context):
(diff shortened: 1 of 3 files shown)


emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/32053869263 new regtests

emolter:
back to draft - blocked until we can merge https://github.com/spacetelescope/stdatamodels/pull/797

src/stdatamodels/fits_support.py
```diff
@@ -1009,24 +968,24 @@ def _can_skip_fits_update(hdulist, asdf_struct, context):
 
     # Check for FITS hash and compare to current. If equal, automatically skip.
     if asdf_struct.tree.get(FITS_HASH_KEY, None) is not None:
-        if asdf_struct.tree[FITS_HASH_KEY] == fits_hash(hdulist):
+        if asdf_struct.tree[FITS_HASH_KEY] == fits_hash(list(hdulist)):
```
braingram: Why is the hdulist converted to a list here?

src/stdatamodels/fits_support.py
```diff
@@ -908,7 +869,7 @@ def from_fits_asdf(
 
 def _map_hdulist_to_arrays(hdulist, af):
     # hdulist[key] is very inefficient so pre-index hdus
-    hdu_index = {(h.name or i, h.ver): h for i, h in enumerate(hdulist)}
+    hdu_index = {(h.name, h.ver or i): h for i, h in enumerate(hdulist)}
```
braingram: I think this (in part) causes issues for old files. If I generate a file with 3 non-primary HDUS (2 unnamed)
```
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU       4   ()
  1                1 ImageHDU         6   (3,)   int64
  2  SCI           1 ImageHDU         7   (3,)   int64
  3                1 ImageHDU         6   (3,)   int64
```
and store them in a tree with:
```
└─arrs (list)
  ├─[0] (ndarray)
  │ ├─shape (tuple)
  │ │ └─[0] (int): 3
  │ └─dtype (Int64DType): >i8
  ├─[1] (ndarray)
  │ ├─shape (tuple)
  │ │ └─[0] (int): 3
  │ └─dtype (Int64DType): >i8
  └─[2] (ndarray)
    ├─shape (tuple)
    │ └─[0] (int): 3
    └─dtype (Int64DType): >i8
```
This round-trips on main without issue. However if I attempt to load the file saved on main with this PR it fails with:
```
KeyError: ('', 3)
```
Would you undo the changes to this function? I don't think they should be necessary for changes to write behavior and are hiding a bug in the writing code.

braingram (changes requested):
I did not do a full review after running into an issue with opening a file produced on main with this PR. I suspect it will require additional code changes so I figured we could address that issue first before proceeding with review.

src/stdatamodels/fits_support.py
emolter: How are you generating the file exactly, and how are you reading it?  I'm trying to add a test for this case to ensure the problem doesn't recur or persist.  I have

```
from astropy.io import fits
from stdatamodels import DataModel

def test_unnamed(tmp_path):
    shape = (3,)
    hdul = fits.HDUList()
    hdul.append(fits.PrimaryHDU())
    hdul.append(fits.ImageHDU(data=np.zeros(shape, dtype=np.int64)))
    hdul.append(fits.ImageHDU(data=np.ones(shape, dtype=np.int64), name="SCI"))
    hdul.append(fits.ImageHDU(data=np.ones(shape, dtype=np.int64) * 2))
    file_path = tmp_path / "test.fits"
    hdul.writeto(file_path)
    print(hdul.info())

    with DataModel(file_path) as model:
        assert isinstance(model, DataModel)
        assert hasattr(model.extra_fits, "SCI")
```

The print statement looks identical to what you showed above, and the test is passing.

braingram: Unfortunately the issue is seen when a file is written from main and read with the PR (which we don't have a system for testing). One option would be to generate the test file without using stdatamodels and instead only use astropy and asdf. If it's helpful I can try to come up with an example. In essence the issue is that on main nameless hdus are written with `fits:<hdu_index>` which this PR changes to `fits:'',<hdu_index>`. (note the blank name) I suspect there is also a backwards incompatibility where files like the above written with this PR aren't readable with old versions since the above test file writes out a link to a nameless hdu with version 3 which does not exist in the hdulist.

emolter: I'm happy reverting it, that's not the problem.  But it would be very helpful to have a test case.

This was supposed to be what you said, I generated the file using exclusively astropy as you can see.  Not clear to me why that works while nameless HDUs on stdatamodels main does not.

braingram: Here's an example:
```python
import astropy.io.fits as fits
import numpy as np

ff = fits.HDUList()
ff.append(fits.PrimaryHDU())
ff.append(fits.ImageHDU([0]))
ff.append(fits.ImageHDU([1], name="SCI"))
ff.append(fits.ImageHDU([2]))

asdf_bytes = b"""#ASDF 1.0.0
#ASDF_STANDARD 1.6.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
arrs:
- !core/ndarray-1.1.0
  source: fits:1
  datatype: int64
  byteorder: big
  shape: [1]
- !core/ndarray-1.1.0
  source: fits:SCI,1
  datatype: int64
  byteorder: big
  shape: [1]
- !core/ndarray-1.1.0
  source: fits:3
  datatype: int64
  byteorder: big
  shape: [1]
..."""

data = np.frombuffer(asdf_bytes, dtype=np.uint8)[None, :]
fmt = f"{len(data[0])}B"
column = fits.Column(array=data, format=fmt, name="ASDF_METADATA")
ff.append(fits.BinTableHDU.from_columns([column], name="ASDF"))
ff.writeto("foo.fits", overwrite=True)

from stdatamodels import asdf_in_fits
af = asdf_in_fits.open("foo.fits")
assert af["arrs"][0][0] == 0
assert af["arrs"][1][0] == 1
assert af["arrs"][2][0] == 2
```

The "foo.fits" file is essentially what stdatamodels main will produce for an asdf_in_fits file with the above hdu/tree structure. Does this help to illustrate the issue?

emolter: Thanks, yes the problem is clearer now, and it's helpful to have a test case we can include

emolter: ok I think I finally fixed this in https://github.com/spacetelescope/stdatamodels/pull/794/commits/30fe7b089e537a0bc0f462b7f411e2bf4ba6dad4 

All changes to `_map_hdulist_to_arrays` have been reverted.

The added test failed as of https://github.com/spacetelescope/stdatamodels/pull/794/commits/d609ee541e83259b7050934bbf5407ae707edd83 but passes now.

I'll check on the `ResourceWarning` I'm seeing now

emolter:
https://github.com/spacetelescope/RegressionTests/actions/runs/32500064544

src/stdatamodels/fits_support.py
emolter: There is no benefit, I'll revert this change.

I realized I needed to use `fits_hash` on the list of cached HDUs, which isn't strictly an `HDUList` object.  Then I changed the `fits_hash` docstring and variable names to clarify that it can accept a non-`HDUList` list of HDUs (which is kinda a silly statement but true I think).  Then I updated this line so that all usages of `fits_hash` used a plain list.  But as you say it's totally unnecessary, I can just update the `fits_hash` docstring to say an HDUList or plain list of HDU objects are both allowed.

emolter: https://github.com/spacetelescope/stdatamodels/pull/794/commits/50faf9e03f1f7185aae7cb62e6d7c828fda51095

PR #802, opened 2026-08-24
Expand FitsContext to further avoid direct use of hdulist

braingram:
This is an alternative to https://github.com/spacetelescope/stdatamodels/pull/794 based on information learned while reviewing that PR leading to an alternative approach.

Restructure writing in fits_support by:
- expanding FitsContext to act as the "cache" for accessing hdus without using hdulist
- remove some (now) unused private API: _make_hdu, _get_or_make_hdu, _normalize_arrays (combined with datetime conversion tree walk)
- adding a warning when a pre-existing hdulist with ambiguous hdu names is provided for writing. This only impacts asdf_in_fits (no impact on pipeline uses). In the future we should make this warning an error as the resulting file can be unexpected and fail to roundtrip (this is the case on main but with no warning).

Regression tests:
https://github.com/spacetelescope/RegressionTests/actions/runs/32756102529

## Tasks
- [ ] If this change affects user-facing code or public API, add news fragment file(s) to `changes/` (see [the changelog instructions](https://github.com/spacetelescope/stdatamodels/blob/main/changes/README.md)).

- `<PR#>.feature.rst`
- `<PR#>.bugfix.rst`
- `<PR#>.docs.rst`
- `<PR#>.other.rst`
</details>

### DIFF
--- a/src/stdatamodels/fits_support.py
+++ b/src/stdatamodels/fits_support.py
@@ -206,13 +206,10 @@ def _fits_element_writer(fits_context, validator, fits_keyword, instance, schema
 
     hdu_name = schema.get("fits_hdu", "PRIMARY")
 
-    if fits_context.sequence_index is not None:
-        name = (hdu_name, fits_context.sequence_index)
-    else:
-        name = hdu_name
+    key = _name_to_key(hdu_name, fits_context.sequence_index)
 
-    if fits_context.has_name(name):
-        hdu = fits_context.by_name(name)
+    if fits_context.has_key(key):
+        hdu = fits_context.by_key(key)
     else:
         # make a new one
         hdu_type = _get_hdu_type(hdu_name, schema=schema)
@@ -264,14 +261,11 @@ def _fits_array_writer(fits_context, validator, _, instance, schema):
 
     hdu_type = _get_hdu_type(hdu_name, schema=schema, value=instance)
 
-    if fits_context.sequence_index is not None:
-        name = (hdu_name, fits_context.sequence_index)
-    else:
-        name = hdu_name
+    key = _name_to_key(hdu_name, fits_context.sequence_index)
 
     # Do we need to make an HDU?
-    if fits_context.has_name(name):
-        hdu = fits_context.by_name(name)
+    if fits_context.has_key(key):
+        hdu = fits_context.by_key(key)
 
         # is the type correct?
         if isinstance(hdu, hdu_type):
@@ -296,10 +290,10 @@ def _fits_array_writer(fits_context, validator, _, instance, schema):
         fits_context.append(hdu)
 
     if other_index := fits_context.extension_array_links.get(instance_id):
-        if fits_context.get_index(name) != other_index:
+        if fits_context.get_index(key) != other_index:
             raise ValueError("Linking one array to multiple hdus is not supported")
     else:
-        fits_context.extension_array_links[instance_id] = fits_context.get_index(name)
+        fits_context.extension_array_links[instance_id] = fits_context.get_index(key)
 
 
 # This is copied from jsonschema._validators and modified to keep track
@@ -334,6 +328,16 @@ def _fits_type(fits_context, validator, items, instance, schema):
     return asdf_schema.YAML_VALIDATORS["type"](validator, items, instance, schema)
 
 
+def _name_to_key(name, version=None):
+    if version:
+        return name.lower(), version
+    return name.lower()
+
+
+def _hdu_to_key(hdu):
+    return _name_to_key(hdu.name, hdu.header.get("EXTVER"))
+
+
 class FitsContext:
     def __init__(self, hdulist=None):
         self._preindex_hdulist(hdulist)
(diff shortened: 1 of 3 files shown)


src/stdatamodels/fits_support.py
```diff
@@ -320,23 +259,47 @@ def _fits_array_writer(fits_context, validator, _, instance, schema, fits_hdu_ca
...
+        name = hdu_name
+
+    # Do we need to make an HDU?
+    if fits_context.has_name(name):
+        hdu = fits_context.by_name(name)
+
+        # is the type correct?
+        if isinstance(hdu, hdu_type):
+            # TODO can we avoid assigning to data
+            hdu.data = instance
```
braingram: We can almost avoid `hdu.data = something` (which risks data corruption) with this PR. One (follow-up) option for avoiding this is to also delay the hdu creation until the hdulist is ready to be made. At that point we know the hdu type and can provide the data on init. It would be good to:
- check if astropy does what's expected when data is provided at hdu creation and doesn't silently corrupt data like it does for `hdu.data = something`
- weight the extra book keeping required to accomplish this with the benefits (no known uses result in data corruption but this has come up enough times that I suspect there are some remaining)

src/stdatamodels/fits_support.py
```diff
@@ -372,16 +335,98 @@ def _fits_type(fits_context, validator, items, instance, schema):
...
+                )
+                warnings.warn(msg, UserWarning, stacklevel=2)
+                # Only map the first name -> index to match astropy handling of name
+                # confusion where hdulist["SCI"] with 2x SCI returns the first.
+            else:
+                self._name_to_index[name] = index
+            self._index_to_name[index] = name
+
+    def to_hdulist(self, tree):
+        # TODO strip ASDF to appease a unit test, perhaps we remove the test?
```
braingram: This double-strip is needed for:
https://github.com/spacetelescope/stdatamodels/blob/2022cce1117a6eeee9f0e3cee3c09d16827a04aa/tests/test_fits.py#L573
`_load_extra_fits` on main and with this PR don't allow "ASDF" to end up in extra_fits. The test injects information into extra_fits and then checks the injected information doesn't round-trip. Perhaps another way to approach this is to skip any "ASDF" key in extra_fits on save? It seems a little wrong to silently drop data.

src/stdatamodels/fits_support.py
```diff
@@ -501,56 +528,60 @@ def _create_tagged_dict_for_fits_array(hdu, hdu_index):
...
+                ] + cards
+
+                # make a new hdu
+                hdu = hdu_type(data=parts.get("data"), header=fits.Header(merged_cards))
+
+                # replace the old one
+                fits_context.replace(hdu)
+            else:
+                # update the data
+                # TODO can we avoid assigning to data?
```
braingram: Same as the other comment about `hdu.data` assignment.

tests/test_asdf_in_fits.py
```diff
@@ -205,8 +205,25 @@ def test_non_named_hdus(tmp_path):
...
+    ff = fits.HDUList(
+        [fits.PrimaryHDU(), fits.ImageHDU([0], name="SCI"), fits.ImageHDU([1], name="SCI")]
+    )
+    tree = {"hdus": [hdu.data for hdu in ff[1:]]}
+    with pytest.warns(UserWarning, match="Multiple HDUs share"):
+        asdf_in_fits.write(fn, tree, ff)
+
+    with asdf_in_fits.open(fn) as af:
+        for i, hdu in enumerate(af.tree["hdus"]):
+            # TODO this is wrong but matches main
```
braingram: This test passes on main (if the `pytest.warns` is removed above since this PR introduces the warning). The tldr is that providing an hdulist with 2x SCI extensions (no versions) silently causes data mapping errors on main and with this PR those errors trigger a warning. In the future I think we should make these an error since:
- this can only be triggered with asdf_in_fits
- the user can add unambiguous names/versions to the hdus

src/stdatamodels/fits_support.py
```diff
@@ -320,23 +256,44 @@ def _fits_array_writer(fits_context, validator, _, instance, schema, fits_hdu_ca
     if "datatype" in schema:
         yield from validate._validate_datatype(validator, schema["datatype"], instance, schema)
 
-    hdu_name = _get_hdu_name(schema)
+    hdu_name = schema.get("fits_hdu", "PRIMARY")
     _assert_non_primary_hdu(hdu_name)
-    index = fits_context.sequence_index
-    if index is None:
-        index = 0
```
braingram: This creates a lot of EXTVER cards in files that have HDUS with unique names. If I write a file on main with:
```
import stdatamodels.jwst.datamodels as dm, numpy as np
m = dm.ImageModel()
m.data = np.zeros((10, 10))
m.save("foo.fits")
```
The produced file has fitsinfo
```
Filename: foo.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU       7   ()
  1  SCI           1 ImageHDU         9   (10, 10)   float32
  2  ASDF          1 BinTableHDU     11   1R x 1C   [838B]
```
and from fitsheader:
```
EXTNAME = 'SCI     '           / extension name
EXTVER  =                    1 / extension value
```
The code commented here is why this has an EXTVER even though it does not need one.

